PR #2, opened 2018-09-26
Feature base django app

maciejSamerdak:
An updated and renamed 'project-initialization' branch featuring requested changes from the last pull request review.


requirements.txt
```diff
@@ -0,0 +1,4 @@
+Django>2.1, <2.2
```
kbeker: It should be `Django =>, <2.2`

time_monkey/settings/base.py
```diff
@@ -0,0 +1,119 @@
...
+For the full list of settings and their values, see
+https://docs.djangoproject.com/en/1.11/ref/settings/
+"""
+
+import os
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
```
kbeker: You should not delete `PRIV_KEY` at all from here. Add here, but as a comment and add description what is it used for.

time_monkey/settings/base.py
```diff
@@ -0,0 +1,119 @@
...
+# Database
+# https://docs.djangoproject.com/en/1.11/ref/settings/#databases
+
+DATABASES = {
+    'default': {
+        'ENGINE':           'django.db.backends.postgresql',
+        'NAME':             'time_monkey',
+        'ATOMIC_REQUESTS':  True,
+    }
```
kbeker: Please also add here 
```python
# 'USER':     'postgres',
# 'PASSWORD': '',
# 'HOST':     '',
# 'PORT':     '',
```
It should be added as a comment

maciejSamerdak: Do you mean inside the bracket, below the previous fields, but commented out for future use i.e. 'the intern way'

or

above the code, as a template with perhaps additional information regarding it's usage?

requirements.txt
maciejSamerdak: So this way it refers directly to the version specified in the .lock file, correct?

kbeker: Sorry, I see that I have not write it correctly last time. It should be `Django => 2.1, < 2.2`

time_monkey/settings/base.py
kbeker: Just inside the bracket :)

PR #6, opened 2018-11-08
Feature user login

MartynaAnnaGottschling:


users/permissions.py
```diff
@@ -0,0 +1,17 @@
+from rest_framework import permissions
```
TheCM: You should do permissions adding in separate commit, splitting serializers in separate, and modifying/creating new views in separate.

users/urls.py
```diff
@@ -3,7 +3,7 @@
 from users import views
 
 router = DefaultRouter()
-router.register(r'users', views.UserViewSet)
```
TheCM: Remove one view from urls, adding new one? I don't know what's going on in this commit ;)

users/views.py
```diff
@@ -1,8 +1,33 @@
...
+from rest_framework import permissions
+from users.permissions import IsAdminUser, IsOwnerOrAdmin
+
+
+class UsersViewSet(viewsets.ModelViewSet):
+    queryset = CustomUser.objects.all()
+    permission_classes = (IsAdminUser,)
+
+    def get_serializer_class(self):
+        if self.action == 'list':
```
TheCM: Would be fine if there is something like enum for that. `self.action == Action.LIST.value`. This is not critical, but we should do it after demo.

users/views.py
```diff
@@ -4,6 +4,23 @@
 from rest_framework import permissions
 from users.permissions import IsAdminUser, IsOwnerOrAdmin
 
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+from rest_framework.reverse import reverse
+
+
+@api_view(['GET'])
```
TheCM: Same like actions: should be enum here.
But `GET` method is accepted by default so just delete it.

users/views.py
```diff
@@ -4,6 +4,23 @@
...
+def api_root(request, format=None):
```
TheCM: Why api_root?

users/views.py
```diff
@@ -4,6 +4,23 @@
...
+    if not request.user.is_authenticated:
+        return Response({
```
TheCM: Why do you return response with all users when user is not authenticated?

users/views.py
```diff
@@ -4,6 +4,23 @@
...
+                'users': reverse('users-list', request=request, format=format),
+            })
+    else:
+        return Response({
```
TheCM: Why do you return all users and account separately if user is authenticated?

users/urls.py
```diff
@@ -1,10 +1,44 @@
+"""
```
TheCM: Just delete instead of commenting.

users/permissions.py
```diff
@@ -14,4 +14,8 @@ def has_permission(self, request, view):
             return True
         return request.user and request.user.is_authenticated and request.user.user_type == 'Admin'
     def has_object_permission(self, request, view, obj):
-        return obj == request.user or request.user.user_type == 'Admin'
+        if request.user:
+            if request.user.user_type == 'Admin':
```
TheCM: Can be:
`return request.user.user_type == 'Admin' or obj == request.user`

users/serializers.py
```diff
@@ -17,27 +17,52 @@ class UserListSerializer(UserSerializer):
     )
     class Meta:
         model = CustomUser
-        fields = ('url', 'email', 'first_name', 'last_name')
```
TheCM: You can fixup it to earlier commit when you create new serializers.

users/urls.py
```diff
@@ -40,5 +28,5 @@
     url(r'^$', api_root),
     url(r'^users/$', users_list, name='users-list'),
     url(r'^users/(?P<pk>[0-9]+)/$', users_detail, name='users-detail'),
-    url(r'^account/(?P<pk>[0-9]+)/$', user_detail, name='user-detail')
+    url(r'^account/(?P<pk>[0-9]+)/$', user_detail, name='user-detail'),
```
TheCM: you can fixup it

users/urls.py
```diff
@@ -1,18 +1,6 @@
...
-
-urlpatterns = [
-    url(r'^', include(router.urls))
-]
-"""
-
-from users.views import UserViewSet, UsersViewSet, api_root
+from users.views import UserViewSet
+from users.views import UsersViewSet
+from users.views import api_root
```
TheCM: Fixup

users/views.py
```diff
@@ -13,12 +18,25 @@
...
+                    request=request,
+                    format=format,
+                ),
+                'account': reverse(
+                    'user-detail',
+                    args=(request.user.pk,),
+                    request=request,
+                    format=format,
+                ),
             })
```
TheCM: This commit is a huge fixup.

users/permissions.py
```diff
@@ -1,21 +1,21 @@
...
-        return request.user and request.user.is_authenticated and request.user.user_type == 'Admin'
+        return request.user and request.user.is_authenticated and request.user.user_type == CustomUser.UserType.ADMIN.name
     def has_object_permission(self, request, view, obj):
-        if request.user:
-            if request.user.user_type == 'Admin':
+        if request.user.is_authenticated:
+            if request.user.user_type == CustomUser.UserType.ADMIN.name:
                 return True
             return obj == request.user
-        #return obj == request.user or request.user.user_type == 'Admin'
```
TheCM: And fixup

users/tests.py
```diff
@@ -63,14 +64,14 @@ def test_user_should_hold_email_address(self):
             CustomValidationError,
             CustomValidationErrorText.VALIDATION_ERROR_EMAIL_MESSAGE,
         ):
-            CustomUser.objects._create_user(None,"testuserpasswd",False,False)
+            CustomUser.objects._create_user(None,"testuserpasswd",False,False,CustomUser.UserType.EMPLOYEE.name,)
```
TheCM: Add spaces between `,`

users/tests.py
```diff
@@ -63,14 +64,14 @@ def test_user_should_hold_email_address(self):
...
 
     def test_user_should_hold_password(self):
         with self.assertRaisesRegexp(
             CustomValidationError,
             CustomValidationErrorText.VALIDATION_ERROR_PASSWORD_MESSAGE,
         ):
-            CustomUser.objects._create_user("testuser@example.com",None,False,False)
+            CustomUser.objects._create_user("testuser@example.com",None,False,False,CustomUser.UserType.EMPLOYEE.name,)
```
TheCM: Same

users/permissions.py
```diff
@@ -0,0 +1,17 @@
+from rest_framework import permissions
+
+
+class IsAdminUser(permissions.BasePermission):
+    message = 'You are not allowed to enter - only for admin.'
```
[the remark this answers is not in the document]
MartynaAnnaGottschling: Done

users/permissions.py
```diff
@@ -0,0 +1,17 @@
...
+    def has_permission(self, request, view):
+        return request.user and request.user.is_authenticated and request.user.user_type == 'Admin'
+
+
+class IsOwnerOrAdmin(permissions.BasePermission):
+    message = "It's none of your business."
+    def has_permission(self, request, view):
+        if view.get_object() == request.user:
+            return True
+        return request.user and request.user.is_authenticated and request.user.user_type == 'Admin'
```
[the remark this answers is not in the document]
MartynaAnnaGottschling: Done

users/permissions.py
MartynaAnnaGottschling: Done

users/urls.py
MartynaAnnaGottschling: View class has changed name from UsersViewSet to UserViewSet (to split part for detail user and list of users)

users/views.py
MartynaAnnaGottschling: Done

users/views.py
MartynaAnnaGottschling: Done

users/views.py
MartynaAnnaGottschling: Just to include users app (as users list default) to main page, like build-in rest api root does

users/permissions.py
MartynaAnnaGottschling: Done

users/tests.py
MartynaAnnaGottschling: Done

users/tests.py
MartynaAnnaGottschling: Done

users/views.py
MartynaAnnaGottschling: This api_root is a custom version of default api_root supplied by Django REST. I tried to customize 'main page' for different kind of user.

users/views.py
TheCM: But every user will have access to all existing users in database, I'm right?

users/urls.py
```diff
@@ -1,10 +1,32 @@
...
+    'put': 'update',
+    'patch': 'partial_update',
+    'delete': 'destroy'
+})
+
+urlpatterns = format_suffix_patterns([
+    url(r'^$', api_root),
+    url(r'^users/$', users_list, name='users-list'),
+    url(r'^users/(?P<pk>[0-9]+)/$', users_detail, name='users-detail'),
+    url(r'^account/(?P<pk>[0-9]+)/$', user_detail, name='user-detail'),
```
TheCM: `user-detail` and `users-detail` is too close names. Try to use more letters to describe more accurately, what is which view for.

users/views.py
```diff
@@ -1,8 +1,69 @@
...
+from users.common.fields import Action
 from users.models import CustomUser
+from users.permissions import IsAdminUser
+from users.permissions import IsOwnerOrAdmin
+from users.serializers import UserCreateSerializer
+from users.serializers import UserDetailSerializer
+from users.serializers import UserListSerializer
 from users.serializers import UserSerializer
-from rest_framework import viewsets
+from users.serializers import UserUpdateSerializer
```
TheCM: Nice

PR #8, opened 2018-11-09
Chore add reusable project utils

maciejSamerdak:
Adds utils directory for files containing code for general project purposes.

As for now, generic classes for serializers' and models' unit tests are provided.
They contain multi-purposed functions meant to reduce code usage in actual unit test classes and ease test development.

### DIFF
--- a/utils/base_tests.py
+++ b/utils/base_tests.py
@@ -2,6 +2,27 @@ from django.core.exceptions import ValidationError
 from django.test import TestCase
 from django.db.models.base import Model
 from rest_framework.serializers import BaseSerializer
+import math
+
+
+"""
+Generates number with one too many digits than specified, based on e mathematical constant.
+Example: digits=4 will return 135.91
+"""
+def generate_too_many_digits(digits):
+    assert isinstance(digits, int)
+    assert digits > 1
+    return round((math.e / 2 * 10 ** (digits - 2)), 2)
+
+
+"""
+Generates number with one too many decimal places than specified, based on e mathematical constant.
+Example: decimal_places=2 will return 1.359
+"""
+def generate_too_many_decimal_places(decimal_places):
+    assert isinstance(decimal_places, int)
+    assert decimal_places >= 0
+    return round((math.e / 2), decimal_places + 1)
 
 
 class BaseModelTestCase(TestCase):


utils/base_tests.py
```diff
@@ -0,0 +1,164 @@
...
+                    model.full_clean()
+
+    """
+    Test that putting specified value in specified field should result in successful model validation.
+    """
+    def field_should_accept_input(self, field, value):
+        self.__field_input_acceptance_test(
+            field=field,
+            value=value,
+            is_valid=False,
```
TheCM: If validation should be successful, than why is_valid variable is `False`?

utils/base_tests.py
```diff
@@ -0,0 +1,164 @@
...
+    WARNING: It is assumed that the field is of DateTime type.
+    Using any other time-related type can result in fail!
+    """
+    def field_auto_now_test(self, auto_field, update_field, value):
+        model = self.default_model()
+        model.full_clean()
+        model.save()
+        previous_value = getattr(model, auto_field)
+        model_pk = model.pk
+        model = self.model_class.objects.get(pk=model_pk)
```
TheCM: You can use `model.refresh_from_db()`

utils/base_tests.py
```diff
@@ -0,0 +1,164 @@
...
+        setattr(model, update_field, value)
+        model.full_clean()
+        model.save()
+        current_value = getattr(model, auto_field)
+        self.assertNotEqual(previous_value, current_value)
```
TheCM: It would be safer if you use mocking datetimes to do that - it is a possibility that values will not differ at all.
Bu this is improvement to add to Pivotal.

utils/base_tests.py
```diff
@@ -162,3 +163,68 @@ def key_should_not_accept_incorrect_input(self, field, value, error_message=None
...
+    """
+    Create serializer from required_input.
+    """
+    def default_serializer(self):
+        return self.serializer_class(data=self.required_input)
+
+    """
+    Private method containing base code for running serializer validation tests.
+    """
+    def __field_input_acceptance_test(self, field, value, is_valid):
```
TheCM: One underscore is enough:
https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

utils/base_tests.py
```diff
@@ -2,6 +2,23 @@
...
 from django.db.models.base import Model
 from rest_framework.serializers import BaseSerializer
+import math
+
+
+"""
+Generates number with one too many digits then specified, based on e mathematical constant.
+Example: digits=4 will return 135.91
+"""
+def generate_too_many_digits(digits):
```
TheCM: Add python assertion that digit should be greater than 0.

utils/base_tests.py
```diff
@@ -2,6 +2,23 @@
...
+    return round((math.e / 2 * 10 ** (digits - 2)), 2)
+
+
+"""
+Generates number with one too many decimal places then specified, based on e mathematical constant.
+Example: decimal_places=2 will return 1.359
+"""
+def generate_too_many_decimal_places(decimal_places):
```
TheCM: Add python assertion that digit should be greater than 0.

utils/base_tests.py
```diff
@@ -2,6 +2,23 @@
 from django.test import TestCase
 from django.db.models.base import Model
 from rest_framework.serializers import BaseSerializer
+import math
+
+
+"""
+Generates number with one too many digits then specified, based on e mathematical constant.
```
TheCM: than

utils/base_tests.py
```diff
@@ -2,6 +2,23 @@
...
+Example: digits=4 will return 135.91
+"""
+def generate_too_many_digits(digits):
+    return round((math.e / 2 * 10 ** (digits - 2)), 2)
+
+
+"""
+Generates number with one too many decimal places then specified, based on e mathematical constant.
```
TheCM: than

PR #10, opened 2018-11-13
Feature employees app

maciejSamerdak:


employees/tests/test_unit_report_viewset.py
```diff
@@ -0,0 +1,143 @@
...
+        other_report.full_clean()
+        other_report.save()
+
+        request = APIRequestFactory().get(path=reverse('report-list'))
+        request.user = self.user
+        response = ReportViewSet.as_view({'get': 'list'})(request)
+        self.assertEqual(response.status_code, 200)
+        self.assertNotContains(response, other_report.description)
+
+    def test_that_report_list_view_should_add_new_report_on_post(self):
```
maciejSamerdak: The way this test is performed, I cannot verify if anything in the backend is happening.
I.e. I no new object is created in the base here, all I can see is the status code and template contents.

Is this an expected behaviour for views' unit tests? Is what I'm describing above an integration test?

If so, should rename those tests appropriately?

employees/tests/test_unit_report_viewset.py
```diff
@@ -0,0 +1,143 @@
...
+        response = ReportViewSet.as_view({'get': 'retrieve'})(request, pk=self.report.pk)
+        self.assertEqual(response.status_code, 403)
+
+    def test_report_detail_view_should_not_render_non_existing_report(self):
+        request = APIRequestFactory().get(path=reverse('report-detail', args=(999,)))
+        request.user = self.user
+        response = ReportViewSet.as_view({'get': 'retrieve'})(request, pk=999)
+        self.assertEqual(response.status_code, 404)
+
+    def test_report_detail_view_should_update_report_on_put(self):
```
maciejSamerdak: Same as above

employees/tests/test_unit_report_viewset.py
```diff
@@ -0,0 +1,143 @@
...
+                'description': 'Some other description',
+                'project': self.project,
+                'work_hours': Decimal('8.00'),
+            }
+        )
+        request.user = self.user
+        response = ReportViewSet.as_view({'put': 'retrieve'})(request, pk=self.report.pk)
+        self.assertEqual(response.status_code, 200)
+
+    def test_report_detail_view_should_delete_report_on_delete(self):
```
maciejSamerdak: Same as above.

employees/permissions.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework import permissions
+from users.models import CustomUser
+
+
+class IsReportAuthor(permissions.BasePermission):
```
maciejSamerdak: As for now, I am not using any of these anywhere in the app.
Also, I believe permissions should be located in directory such as utils, since everyone might require them in future development.

Should I drop this commit altogether?

employees/tests/test_unit_report_viewset.py
```diff
@@ -0,0 +1,143 @@
+from django.test import TestCase
+from rest_framework.reverse import reverse
+from employees.models import Report
+from users.models import CustomUser
+from managers.models import Project
+import datetime
+from decimal import Decimal
+from employees.views import ReportViewSet
+from rest_framework.test import APIRequestFactory
+from django.contrib.auth.models import AnonymousUser
```
TheCM: Sort these imports in more sensible way. employees adjacent to employees. datetime, decimal at the top, alphabetical order and so one.

employees/tests/test_unit_report_viewset.py
TheCM: Just check if the created object is in database after creation using views.

employees/tests/test_unit_report_viewset.py
TheCM: Also check the database.

employees/tests/test_unit_report_viewset.py
TheCM: Also check the database.

employees/common/constants.py
```diff
@@ -0,0 +1,13 @@
+from django.utils.translation import ugettext_lazy
+from decimal import Decimal
+
+
+class ReportModelConstants:
```
TheCM: I'm wondering why is this a class? Maybe enum would be more appropriate? What do you think?

employees/tests/test_unit_report_model.py
```diff
@@ -0,0 +1,181 @@
+from employees.models import Report
+from managers.models import Project
+from users.models import CustomUser
+import datetime
+from decimal import Decimal
+from employees.common.constants import ReportModelConstants
+from utils.base_tests import BaseModelTestCase
+from utils.base_tests import generate_too_many_decimal_places
+from utils.base_tests import generate_too_many_digits
```
TheCM: Correct imports https://www.python.org/dev/peps/pep-0008/#imports

employees/permissions.py
TheCM: These permissions should be in `users` app since every `Permission` is about the `CustomUser`.

employees/views.py
```diff
@@ -1,3 +1,17 @@
-from django.shortcuts import render
+from employees.models import Report
+from employees.serializers import ReportSerializer
+from rest_framework import permissions
```
TheCM: rest_framework as side package should be at the top.

employees/urls.py
```diff
@@ -0,0 +1,13 @@
+from django.urls import include
+from django.conf.urls import url
+from rest_framework.routers import DefaultRouter
+from employees import views
```
TheCM: WOW, nice imports! :P

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Okay, so during my generic view research I found the expected codes for each operation.
https://www.django-rest-framework.org/api-guide/generic-views/#listmodelmixin

It's nice since those codes describe the exact behaviour I want to test, but now I have a different issue: I can fiddle with multiple HTTP requests here and get same result.
e.g.: If I were to change 'put' to either 'get' or 'post', the status code is always the same.

I just don't get those Rest viewsets :/

employees/common/constants.py
maciejSamerdak: So the only change in here would be defining Enum inheritance? 😛

employees/tests/test_unit_report_viewset.py
maciejSamerdak: And here, after I applied changes in method call, I can also change HTTP request from 'delete' to (if I remember correctly) 'post' and 'get' and I still get the desired result.

Makes little sense to me, but I guess that's just how REST operates?

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Should I put blank line between each group like PEP8 suggests? If so, what is defined as a group? Standard-Third party-Local? Or maybe per module?

```
import datetime
from decimal import Decimal

from django.contrib.auth.models import AnonymousUser
from rest_framework.reverse import reverse
from rest_framework.test import APIRequestFactory

from employees.models import Report
from employees.views import ReportViewSet
```

Like that?

employees/urls.py
maciejSamerdak: No, they are not, they are not alphabetical 😢

employees/permissions.py
maciejSamerdak: Should I drop it then?

employees/common/constants.py
maciejSamerdak: Okay, obviously more than that, because it stopped working 😱

employees/tests/test_unit_report_viewset.py
TheCM: So in REST you have some operations with specified url and method:
- `GET /objects/` one object details. Returns `200`. Database operation: `read`.
- `GET /objects/<id>/` objects list. Returns `200`. Database operation: `read`.
- `POST /objects/` object creation. Returns `201`. Database operation: `write`.
- `DELETE /objects/<id>/` object deletion. Returns `200`. Database operation: `delete`.
- `UPDATE /objects/<id>/` object partial update. Returns `200`. Database operation: `update`.
- `PUT /objects/<id>/` object full update. Returns `200`. Database operation: `update`.

You can distinct particular endpoints checking response and expected result in database. Status code is a confirmation that everything was done correctly.

employees/tests/test_unit_report_viewset.py
TheCM: Yes, see response above.

employees/tests/test_unit_report_viewset.py
TheCM: This is even prettier ;)

employees/common/constants.py
TheCM: https://docs.python.org/3/library/enum.html#creating-an-enum
But you can leave it like it is if you have any argument that this should be a class. Like: I want to inherit from it. Or: I want to create instances.

employees/tests/test_unit_report_viewset.py
maciejSamerdak: I just wonder how am I supposed to interpret action defined as {'put':'create'}. So 'put' is obviously HTTP request and 'create' refers to a viewset element (much like 'retrieve' and 'list' in other cases). But what does it declare here? Why if I were to replace 'put' with 'get' or 'post' it would always return a 201?

TheCM: I think it's DRF bug. Or you can check if method is allowed for specific view or not. This should not happen. You can add chore for that in Pivotal.

employees/permissions.py
TheCM: No, maybe add them as first commit in branch (base commit, at the bottom - it means not at the top, I mean... You know... Trunk of tree... base of the mountain.. anyway...) and in `users` app (beware conflicts with Martyna).

maciejSamerdak: Done

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Done

employees/common/constants.py
maciejSamerdak: Done

employees/tests/test_unit_report_model.py
maciejSamerdak: Done

employees/views.py
maciejSamerdak: Done

employees/urls.py
maciejSamerdak: Done

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Done

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Done

PR #11, opened 2018-11-15
Feature user signup

MartynaAnnaGottschling:


users/serializers.py
```diff
@@ -1,10 +1,17 @@
+from allauth.account import app_settings as allauth_settings
+from allauth.account.adapter import get_adapter
+from allauth.account.utils import setup_user_email
+from allauth.utils import email_address_exists
+
 from django_countries.serializers import CountryFieldMixin
 from django_countries.serializer_fields import CountryField
+from django.utils.translation import ugettext_lazy as _
+
 from rest_framework import serializers
```
TheCM: You can import just needed fields.

users/serializers.py
```diff
@@ -1,10 +1,17 @@
+from allauth.account import app_settings as allauth_settings
+from allauth.account.adapter import get_adapter
```
TheCM: import as allauth_get_adapter

users/serializers.py
MartynaAnnaGottschling: Done

users/serializers.py
MartynaAnnaGottschling: Done

PR #14, opened 2018-12-17
Chore update development-related testing configuration

maciejSamerdak:


lint.sh
```diff
@@ -10,6 +10,6 @@ printf "\n"
 printf "[PYLINT: time-monkey]\n"
 # Find all subdirectories of our python apps and use xargs to pass them as arguments to pylint
 
-find time_monkey/ -maxdepth 1 -mindepth 1 -type d \
+find time_monkey/ employees/ managers/ users/ -maxdepth 1 -mindepth 1 -type d \
```
maciejSamerdak: @MartynaAnnaGottschling suggested I could just write it as `find ./ -maxdepth 1 -mindepth 1 -type d \` and that way it will find tests for all apps, rather than having to remember to type each one of them here manually.

Could someone either accept this or explain why it's wrong?

MartynaAnnaGottschling: In this case `-maxdepth` should change, because 1 would be not enough. At the same time, it looks a little bit unprofessional...

PR #17, opened 2019-01-22
Chore update employees tests

maciejSamerdak:


utils/sample_data_generators.py
```diff
@@ -0,0 +1,25 @@
...
+
+def generate_too_many_digits(digits):
+    assert isinstance(digits, int)
+    assert digits > 1
+    return round((math.e / 2 * 10 ** (digits - 2)), 2)
+
+
+"""
+Generates number with one too many decimal places than specified, based on e mathematical constant.
+Example: decimal_places=2 will return 1.359
```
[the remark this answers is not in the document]
maciejSamerdak: Thought it made sense, since that way you just pass the limit from model constants file for this field. That way, just like function name suggests, it will create value exceeding this limit with marginal difference.

I dunno, it's just me, but personally I find it cleaner than hardcoding these differences like in example above.
I would agree on changing decimal_places parameter from limit value to actual generated value.
Then for testing purposes I would parametrise only this parameter (no difference value to append further), since currently, the way my tests are structured, I use this function only for fail tests.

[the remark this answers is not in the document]
kbeker: In tests we should hardcode some differences to show them that they are differences. I think, that the best way would be just to create function which are creating correct data and in tests change it to malformed for test needs.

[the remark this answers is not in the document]
maciejSamerdak: I hardcoded correct data in my unit tests for models and serializers, rather than deriving them in any way (in my head it's as if user just types it in from some interface). The incorrect values are currently derived from validators' values, so I can get data which exceeds any limit by smallest possible margin. So in my code it's
```
WORK_HOURS_DATA_EXCEEDING_DIGITS_NUMBER = generate_too_many_digits(ReportModelConstants.MAX_DIGITS.value)
    WORK_HOURS_DATA_EXCEEDING_DECIMAL_PLACES = generate_too_many_decimal_places(
                                                    ReportModelConstants.DECIMAL_PLACES.value)
```

I mean, I obviously agree that reusability is a great feature when designing functions. I could even fuse these two functions and expand them to something like
`def generate_decimal(digits=3, decimal_places=2, math_constant=math.e)`, if it's desired. Then I'd change the unit tests to something similar:
```
MARGINALLY_LARGER_DIGITS_NUMBER = ReportModelConstants.MAX_DIGITS.value + 1

WORK_HOURS_DATA_EXCEEDING_DIGITS_NUMBER = generate_decimal(digits = MARGINALLY_LARGER_DIGITS_NUMBER)
```
However, they were created specifically for supplying data for Report related tests. And since that's the only place in the entire app where decimals are used, it will most certainly be their only usage.
Just curious about your general opinion on this idea, really.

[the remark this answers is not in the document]
kbeker: @maciejSamerdak 
>Then I'd change the unit tests to something similar:
> 
> ```
> MARGINALLY_LARGER_DIGITS_NUMBER = ReportModelConstants.MAX_DIGITS.value + 1
> 
> WORK_HOURS_DATA_EXCEEDING_DIGITS_NUMBER = generate_decimal(digits = MARGINALLY_LARGER_DIGITS_NUMBER)
> ```
> However, they were created specifically for supplying data for Report related tests. And since that's the only place in the entire app where decimals are used, it will most certainly be their only usage.
> Just curious about your general opinion on this idea, really.

Instead of giving variables like you mentioned above you should do it exactly it tests e.g.:
```python
    def test_report_serializer_work_hours_field_should_not_accept_value_exceeding_set_digits_number(self):
        self.field_should_not_accept_input(field='work_hours', value=ReportModelConstants.MAX_DIGITS.value + 1)

    def test_report_serializer_work_hours_field_should_not_accept_value_exceeding_set_decimal_places_number(self):
        self.field_should_not_accept_input(field='work_hours', value=ReportModelConstants.DECIMAL_PLACES.value + 1)
```

It will reduce code boilerplate and increase tests readability. Please apply this to our tests.

employees/tests/test_unit_report_model.py
```diff
@@ -156,20 +145,25 @@ def test_report_model_work_hours_field_should_not_be_empty(self):
 
     # PARAM
     def test_report_model_work_hours_field_should_not_accept_value_exceeding_set_digits_number(self):
-        self.field_should_not_accept_input('work_hours', self.WORK_HOURS_DATA_EXCEEDING_DIGITS_NUMBER)
+        input_data = generate_decimal(digits=ReportModelConstants.MAX_DIGITS.value + 1)
+        self.field_should_not_accept_input('work_hours', input_data)
```
maciejSamerdak: @kbeker Is this more to your liking?

kbeker: @maciejSamerdak Try to inline it like this: ```self.field_should_not_accept_input('work_hours', generate_decimal(digits=ReportModelConstants.MAX_DIGITS.value + 1))

If it will be longer than 120 character, leave it this way. :)

utils/sample_data_generators.py
```diff
@@ -0,0 +1,24 @@
+from decimal import Decimal
+import math
+
+
+def generate_decimal(digits=None, decimal_places=None):
+    if not digits:
+        if not decimal_places:
+            digits = 3
+            decimal_places = 2
```
kbeker: Move it to variables declaration instead of `None`. It would be looks better and function will be shorter

kbeker: You can delete all this places where you check if digits or decimal_places exists. Your function should start like this:
```python
def generate_decimal(digits=3, decimal_places=2):
    assert isinstance(digits, int) and digits > 0
    assert isinstance(decimal_places, int) and decimal_places >= 0 # <- if it can be 0 of course :)
.....
```

utils/tests/test_unit_sample_data_generators.py
```diff
@@ -0,0 +1,18 @@
...
+        declared_digits = 4
+        result = generate_too_many_digits(declared_digits)
+        digits = len(str(result).replace('.', ''))
+        self.assertEqual(digits, 5)
+
+    def test_too_many_decimal_spaces_generator_should_return_decimal_with_more_decimal_places_than_declared(self):
+        declared_decimal_places = 2
+        result = generate_too_many_decimal_places(declared_decimal_places)
+        decimal_places = len(str(result).split('.')[1])
+        self.assertEqual(decimal_places, 3)
```
kbeker: You didn't fix tests with new functions.

utils/sample_data_generators.py
maciejSamerdak: Here's the thing: I would like this function to always return 2 decimal places if possible (digits > 2 and decimal_places > 3), since it's more convenient for me. At the same time I want the two parameters to work independently (so it's replicating original usage, where I define only one of them).

Here's the reason I can't just move these values in declaration: let's say I put in 3 decimal places, but the default total digits number is now 3... So I guess I need to append the digits value to make it work.

Ok, but now let's say that I need a number with 2 digits (since I'm already supposed to make this whole function somewhat reusable, I guess). So now the default decimal places are 2 and if we apply the same logic as before, I will have to append the digits number.

So the result now is different than intended.

In order to make it most convenient, I need it to go through those multiple ifs.
Or I drop the 2 digits case, since I personally don't need it anywhere at all.
Or I split it again into two functions, if you want simpler code.

utils/tests/test_unit_sample_data_generators.py
maciejSamerdak: image

utils/sample_data_generators.py
```diff
@@ -0,0 +1,25 @@
+import math
+
+
+"""
+Generates number with one too many digits than specified, based on e mathematical constant.
+Example: digits=4 will return 135.91
+"""
+
+
+def generate_too_many_digits(digits):
```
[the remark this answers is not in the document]
maciejSamerdak: Done

utils/sample_data_generators.py
```diff
@@ -0,0 +1,25 @@
...
+    return round((math.e / 2 * 10 ** (digits - 2)), 2)
+
+
+"""
+Generates number with one too many decimal places than specified, based on e mathematical constant.
+Example: decimal_places=2 will return 1.359
+"""
+
+
+def generate_too_many_decimal_places(decimal_places):
```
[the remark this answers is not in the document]
maciejSamerdak: Done

utils/sample_data_generators.py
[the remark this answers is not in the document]
maciejSamerdak: Done

employees/tests/test_unit_report_model.py
maciejSamerdak: Done

utils/tests/test_unit_sample_data_generators.py
maciejSamerdak: Done

utils/sample_data_generators.py
maciejSamerdak: Done

PR #20, opened 2019-01-28
Feature accurate hour display

### LINKED ISSUE #46: Fix error message in reports form
If you type over 24 huors in reports form, error message shows "(...) 24.00".
Should be "(...) 24:00".

maciejSamerdak:
Resolves #46

Related #106


employees/serializers.py
```diff
@@ -50,3 +50,10 @@ class Meta:
             'description',
             'work_hours',
         )
+
+    def to_representation(self, instance):
+        data = super(ReportSerializer, self).to_representation(instance)
```
kbeker: As far as I know in python3 you don't have to write `super(Class, self)`. `super()` should be enough

employees/serializers.py
```diff
@@ -50,3 +50,10 @@ class Meta:
...
+        # SELF NOTE: For some reason, tests for list get methods got a Queryset as instance in serializer and I dunno why...
+        if type(self.instance) is Report:
```
kbeker: change it to 
```python
if isinstance(self.instance, Report):
```

employees/serializers.py
```diff
@@ -50,3 +50,10 @@ class Meta:
             'description',
             'work_hours',
         )
+
+    def to_representation(self, instance):
+        data = super(ReportSerializer, self).to_representation(instance)
+        # SELF NOTE: For some reason, tests for list get methods got a Queryset as instance in serializer and I dunno why...
```
kbeker: Add it as a issue and in have to be investigate later. After adding issue please delete comment.

maciejSamerdak: image

employees/serializers.py
maciejSamerdak: I've seen it around the web. Thought that maybe it's some sort of good practise, like explicit parametrising in function calls.

maciejSamerdak: Done

employees/serializers.py
maciejSamerdak: Done

employees/serializers.py
maciejSamerdak: Done

Issue #36, opened 2019-02-19
Support Markdown in report description


Szymiks:
Report description field is a common text field, so implementing markdown support helps this field to become more readable.

rwrzesien:
I have discussed this today with @Szymiks and if I understood the case correctly, it is about adding possibility to display HTML tags in description field. From what I have tested, the solution used by @Szymiks only wraps content of the field `<p>` tags, like this:

```
markdown('text')                                                                                                                                           
Out[2]: '<p>text</p>\n'
```

If that solves the issue, then it is OK. But if the case is about adding possibility to edit `description` field with HTML content inside it, then we should take a different approach, which is:
1. Install https://github.com/django-ckeditor/django-ckeditor.
2. Change `description` field into `ckeditor_uploader.fields.RichTextUploadingField`.
3. Add migration.

That should do the case for admin panel. If we need to add WYSIWYG editor outside of the admin panel, this could help https://github.com/django-ckeditor/django-ckeditor#outside-of-django-admin.

@Karrp @Szymiks please take a look on this and share your thoughts.

PR #21, opened 2019-01-28
Feature allow employee to join project

maciejSamerdak:
Resolves #42
Resolves #105

PR #37, opened 2019-02-19
Feature support markdown in report description

Szymiks:
Resolves https://github.com/Code-Poets/sheetstorm/issues/36

Related #106

### DIFF
--- a/employees/models.py
+++ b/employees/models.py
@@ -1,6 +1,7 @@
 from django.core.validators import MaxValueValidator
 from django.core.validators import MinValueValidator
 from django.db import models
+from markdown_deux import markdown
 
 from employees.common.constants import ReportModelConstants
 from employees.common.strings import MAX_HOURS_VALUE_VALIDATOR_MESSAGE
@@ -53,3 +54,7 @@ class Report(models.Model):
     @property
     def work_hours_str(self):
         return self.work_hours.to_eng_string().replace('.', ':')
+
+    @property
+    def markdown_description(self):
+        return markdown(self.description)
--- a/employees/templates/employees/report_list.html
+++ b/employees/templates/employees/report_list.html
@@ -42,8 +42,8 @@
                     {{ report.work_hours_str }}
                 </td>
                 <td class="description-column">
-                    {{ report.description }}
-                </td>
+                    {{ report.markdown_description|safe|linebreaksbr }}
+                <td>
                 <td class="edit-button-column Invisible">
                     {% if report.editable %}
                     <a href="{% url 'custom-report-detail' pk=report.id %}" class="btn btn-info">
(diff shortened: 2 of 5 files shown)


employees/models.py
```diff
@@ -6,6 +6,7 @@
 from employees.common.validators import MaxDecimalValueValidator
 from managers.models import Project
 from users.models import CustomUser
+from markdown_deux import markdown
```
maciejSamerdak: This line should be moved elsewhere.
There's a strict order for package imports, it goes like this:
1. Standard library imports
2. Third-party packages imports
3. Custom packages imports

Each of these groups is also separated by line.
This import should be located in the second group, among Django packages imports.

Don't forget about alphabetical order. 😉

employees/models.py
```diff
@@ -39,3 +40,7 @@ class Report(models.Model):
     editable = models.BooleanField(
         default=True,
     )
+
+    def markdown_description(self):
+        description = self.description
+        return markdown(description)
```
maciejSamerdak: You might as well just put it in one line `return markdown(self.description)`.

employees/models.py
```diff
@@ -39,3 +40,6 @@ class Report(models.Model):
     editable = models.BooleanField(
         default=True,
     )
+
+    def markdown_description(self):
```
rwrzesien: Please make it a property.

employees/templates/employees/report_list.html
```diff
@@ -42,7 +42,7 @@ <h1>{{ UI_text.PAGE_TITLE.value }}</h1>
                     {{ report.work_hours }}
                 </td>
                 <td>
-                    {{ report.description }}
+                    {{ report.markdown_description|safe|linebreaksbr }}
```
rwrzesien: Could you describe the general purpose of this improvement? Is it about displaying `html` tags in code as html? If so, just `safe` filter might do it.

rwrzesien:
@Szymiks Same as in previous pull request, please link to issue description with `Resolves https://github.com/Code-Poets/sheetstorm/pull/21` in first pull request comment. Also it would be good, if the issue has some actual description ;)

employees/templates/employees/report_list.html
Szymiks: Thanks for than question because now I can see better how It works and it answer for @maciejSamerdak  question. Markdown method returns code with 'html' tags so if we write to big text there, we could see an error 'value too long for type character varying(256)'.  Without 'Safe' filter our text field doesn't work good becasue there are 'html' tags and we can see them. Last one filter help our field to find when someone put enter and but now when I tested it again I think that 'linbreaks' works nicer than 'linebreaksbr', but this is only my opinion.

Issue #58, opened 2019-02-27
Login without using domain (@codepoets.it)


Szymiks:
It would be easier because everyone has the same domain.

Karrp:
We will not use this feature.

PR #59, opened 2019-02-27
Feature login without using domain

Szymiks:
Resolves https://github.com/Code-Poets/sheetstorm/issues/58

Our app is using djanngo login system so I think that it works inside django. I have checked many websites and solutions, but the problem is that we don't have username field in our user model. So I've tried to do this as you can see. I am not sure if this is a good way to solve this problem, because urls.py doesn't look readable but I don't have any idea to resolve this problem in another way. Could anyone check correcntess of this code?

### DIFF
--- a/users/forms.py
+++ b/users/forms.py
@@ -1,7 +1,10 @@
 from django.contrib.auth import authenticate
-from django.contrib.auth.forms import UserChangeForm, AuthenticationForm
+from django.contrib.auth.forms import AuthenticationForm
+from django.contrib.auth.forms import UserChangeForm
 from django.contrib.auth.forms import UserCreationForm
 from django_countries.widgets import CountrySelectWidget
+
+from users.common.constants import VALID_EMAIL_DOMAIN_LIST
 from users.models import CustomUser
 
 
@@ -28,15 +31,21 @@ class CustomUserChangeForm(UserChangeForm):
 
 
 class LoginAuthentication(AuthenticationForm):
+
     def clean(self):
         username = self.cleaned_data.get('username')
-        if "@" not in username:
-            username = username + "@codepoets.it"
         password = self.cleaned_data.get('password')
+        domain_list_size = len(VALID_EMAIL_DOMAIN_LIST) - 1
 
         if username is not None and password:
             self.user_cache = authenticate(self.request, username=username, password=password)
+
+            while self.user_cache is None and domain_list_size >= 0:
+                username_temp = username + VALID_EMAIL_DOMAIN_LIST[domain_list_size]
+                self.user_cache = authenticate(self.request, username=username_temp, password=password)
+                domain_list_size -= 1
+
             if self.user_cache is None:
                 raise self.get_invalid_login_error()
-            else:
-                self.confirm_login_allowed(self.user_cache)
+
+        return self.cleaned_data
--- a/users/forms.py
+++ b/users/forms.py
@@ -30,8 +30,8 @@ class CustomUserChangeForm(UserChangeForm):
 class LoginAuthentication(AuthenticationForm):
     def clean(self):
         username = self.cleaned_data.get('username')
-        if "@" not in str(username):
-            username = str(username) + "@codepoets.it"
+        if "@" not in username:
+            username = username + "@codepoets.it"
         password = self.cleaned_data.get('password')
 
         if username is not None and password:
--- a/users/forms.py
+++ b/users/forms.py
@@ -1,4 +1,5 @@
-from django.contrib.auth.forms import UserChangeForm
+from django.contrib.auth import authenticate
+from django.contrib.auth.forms import UserChangeForm, AuthenticationForm
 from django.contrib.auth.forms import UserCreationForm
 from django_countries.widgets import CountrySelectWidget
 from users.models import CustomUser
@@ -24,3 +25,18 @@ class CustomUserChangeForm(UserChangeForm):
         model = CustomUser
         fields = '__all__'
         widgets = {'country': CountrySelectWidget()}
+
+
+class LoginAuthentication(AuthenticationForm):
+    def clean(self):
+        username = self.cleaned_data.get('username')
+        if "@" not in str(username):
+            username = str(username) + "@codepoets.it"
+        password = self.cleaned_data.get('password')
+
+        if username is not None and password:
+            self.user_cache = authenticate(self.request, username=username, password=password)
+            if self.user_cache is None:
+                raise self.get_invalid_login_error()
+            else:
+                self.confirm_login_allowed(self.user_cache)
--- a/users/forms.py
+++ b/users/forms.py
@@ -41,7 +41,7 @@ class LoginAuthentication(AuthenticationForm):
             self.user_cache = authenticate(self.request, username=username, password=password)
 
             while self.user_cache is None and domain_list_size >= 0:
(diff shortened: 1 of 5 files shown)


users/forms.py
```diff
@@ -1,4 +1,5 @@
-from django.contrib.auth.forms import UserChangeForm
+from django.contrib.auth import authenticate
+from django.contrib.auth.forms import UserChangeForm, AuthenticationForm
```
rwrzesien: Please split imports into separate lines. 

This is specific to CodePoets code style, which, if you have not yet a chance to get familiar with, is located here https://github.com/Code-Poets/coding-style/wiki/python-style#zawijanie-wci%C4%99cia-nawiasy. If you don't have access, please let me know in comment. Please read it when you have time and try to follow it. Off course it is almost impossible for any developer regardless experience level to read it only once and follow it without any forgets, so nobody is expecting that from you. But if at some point you are not sure what to do, you should check there first for reference.

users/forms.py
```diff
@@ -24,3 +25,18 @@ class Meta:
         model = CustomUser
         fields = '__all__'
         widgets = {'country': CountrySelectWidget()}
+
+
+class LoginAuthentication(AuthenticationForm):
+    def clean(self):
+        username = self.cleaned_data.get('username')
+        if "@" not in username:
+            username = username + "@codepoets.it"
```
rwrzesien: `"@codepoets.it"` this can be defined as a constant or even as a setting, depending on the outcome of the functional discussion which I want to start about issue related with this pull request. But this discussion is not related with the code itself, so for now please make it a constant.

users/urls.py
```diff
@@ -1,9 +1,8 @@
+from django.urls import include
 from rest_framework.urlpatterns import format_suffix_patterns
 
 from django.conf.urls import url
-from django.conf.urls import include
 from django.urls import path
-
```
rwrzesien: You should keep a line of break between Django specific imports and custom application imports.

users/urls.py
```diff
@@ -32,6 +31,9 @@
     url(r'^api/users/(?P<pk>[0-9]+)/$', users_detail, name='users-detail'),
     url(r'^api/account/(?P<pk>[0-9]+)/$', user_account_detail, name='user-account-detail'),
     url(r'^$', views.index, name='home'),
+    #Override login view
```
rwrzesien: Please add a space between `#` and rest of the comment.

users/urls.py
```diff
@@ -32,6 +31,9 @@
...
+    url('login/', views.CustomUserLoginView.as_view(), name='login'),
+    #Use django login system
```
rwrzesien: Same here.

users/views.py
```diff
@@ -249,3 +251,11 @@ def get(cls, request):
             'serializer': users_serializer,
             'users_list': users_queryset,
         })
+
+
+class CustomUserLoginView(LoginView):
+    """
+       Use Django LoginView with diffrent form_class
```
rwrzesien: Please align the comment. You can also make it one line comment.

users/forms.py
```diff
@@ -24,3 +25,18 @@ class Meta:
...
+        password = self.cleaned_data.get('password')
+
+        if username is not None and password:
+            self.user_cache = authenticate(self.request, username=username, password=password)
+            if self.user_cache is None:
+                raise self.get_invalid_login_error()
+            else:
+                self.confirm_login_allowed(self.user_cache)
```
rwrzesien: Each `clean` method in Django form must return data. I think it is enough to return `self.cleaned_data` here.

users/forms.py
```diff
@@ -24,3 +25,18 @@ class Meta:
...
+
+
+class LoginAuthentication(AuthenticationForm):
+    def clean(self):
+        username = self.cleaned_data.get('username')
+        if "@" not in username:
+            username = username + "@codepoets.it"
+        password = self.cleaned_data.get('password')
+
+        if username is not None and password:
```
rwrzesien: I think there is something funky going on with the data types. If `username` will be `None`, then the line https://github.com/Code-Poets/sheetstorm/pull/59/files#diff-ee6dfbb85387b31ef4a4e46efdb9be53R33 will fail:

```
>>> '@' not in None
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'NoneType' is not iterable
```

I think the quickest solution for this is to move condition `if "@" not in username:` below this condition.

rwrzesien:
Looks good, but some improvements can be applied. 

> Could anyone check correcntess of this code?

Why don't you write a test for it then? :) This should be a part of the pull request, and it will actually allow you to check the code and find bugs before submitting the code. You can find existing tests for user login in https://github.com/Code-Poets/sheetstorm/blob/master/users/tests/test_user_login.py and try to add one related with your case. If you will have problems with that, please let me know.

> I have checked many websites and solutions, but the problem is that we don't have username field in our user model. 

There is actually something like `USERNAME_FIELD = 'email'` on the `CustomUser` model, which tells Django to treat this field as username field.

> So I've tried to do this as you can see. I am not sure if this is a good way to solve this problem, because urls.py doesn't look readable but I don't have any idea to resolve this problem in another way.

I think that your confusion here is understandable, because related issue description says about treating an `email` field as non-email. Implementing this issue actually drastically limits the usage options for the future for anyone without `codepoets` email address. If someone from outside the company would like to use the application, we can't create an account for him without creating email account first. In my personal opinion this limitation is not necessary. I was discussing this with @kbeker today and we want to involve more people into this discussion to figure out if this is really needed.

Taking this into consideration, I think you did a good job for finding a solution to untypical problem. Don't think about this as an unnecessary work, because you definitely can learn something new from it. It would be great if you could at least fix pending comments and add a simple test for this case.

rwrzesien:
One more thing, in first pull request comment please always add something like `Resolves https://github.com/Code-Poets/sheetstorm/issues/58`. This way when the pull request is merged, the related issue is closed automatically.

users/forms.py
MartynaAnnaGottschling: @Szymiks you can use `VALID_EMAIL_DOMAIN_LIST` from `users/common/constants.py`

users/views.py
rwrzesien: Sorry for confusion, when I wrote `one line comment` I have meant something like:
```""" Use Django LoginView with diffrent form_class. """```

Class and method comments should be written this way.

rwrzesien:
Some bugs appeared near usage of `VALID_EMAIL_DOMAIN_LIST`, but I think the reason is that you have over complicated it.

Tests looks very good :+1:

users/forms.py
rwrzesien: Idea from @MartynaAnnaGottschling is very good, but I think you have over complicated the implementation a little bit. You can just use regular `for` loop to iterate over `VALID_EMAIL_DOMAIN_LIST` without need to get its `len()` first, then after line with `self.user_cache = authenticate(...)` check `if self.user_cache is not None`, and if so, `break` the loop. This way you will not have to use extra `domain_list_size` variable.

rwrzesien: There is just one optimization you might add - the check the condition `if self.user_cache is not None:` just below authenticate - this way you won't have to get in next loop step, and it would look more natural. But it is fine the way it is too.

Karrp:
We will not use this feature.

Issue #69, opened 2019-03-01
Sorting in project names separates upper and lowercase


Karrp:
Lowercase and uppercase are sorted independently of each other  (as in uppercase first, lowercase last).

So if we have for example projects named *Sheetstorm*, *Time monkey* and *e_munchkin*, this is going to be the exact output of their sorting.

PR #81, opened 2019-03-06
Bugfix sorting in project names separates upper and lowercase

rwrzesien:
Resolves https://github.com/Code-Poets/sheetstorm/issues/69

### DIFF
--- a/managers/factories.py
+++ b/managers/factories.py
@@ -1,0 +1,13 @@
+import factory
+import factory.fuzzy
+
+from managers.models import Project
+
+
+class ProjectFactory(factory.DjangoModelFactory):
+
+    class Meta:
+        model = Project
+
+    name = factory.Faker('sentence', nb_words=3)
+    start_date = factory.Faker('date')
--- a/managers/views.py
+++ b/managers/views.py
@@ -1,8 +1,9 @@
+from django.db.models.functions import Lower
 from rest_framework import viewsets
 from managers.models import Project
 from managers.serializers import ProjectSerializer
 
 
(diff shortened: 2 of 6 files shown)


managers/factories.py
```diff
@@ -0,0 +1,13 @@
+import factory
+import factory.fuzzy
+
+from managers.models import Project
+
+
+class ProjectFactory(factory.DjangoModelFactory):
```
kbeker: Please add simple test for this factory

managers/views.py
```diff
@@ -1,8 +1,9 @@
+from django.db.models.functions import Lower
```
kbeker: This function just make string which is composed with Upper letter to lower letters?

requirements-development.txt
```diff
@@ -10,6 +10,7 @@ django-allauth==0.37.1
 django-countries==5.3.2
 django-rest-auth==0.9.3
 djangorestframework==3.8.2
+factory-boy==2.11.1
```
kbeker: If it is is normal requirements we don't have to add it also to requirements-develpment I think :)

Karrp:
@rwrzesien Please add this pull requests to Sheet Storm project

managers/views.py
rwrzesien: Yes, but it does it on database side, during SQL query.

managers/factories.py
rwrzesien: As discussed, we don't want to test factories.

requirements-development.txt
rwrzesien: Done :+1:

PR #85, opened 2019-03-08
Feature project list view

### LINKED ISSUE #50: Manager Base Interface
- [ ]  Dashboard with available projects
- [ ]  Interface with people assigned to projects (feature moved to creating raport feature)
- [ ]  Interface with all details of chosen project

Manager Base Interface is directed to the Admin-type and Manager-type user
- for Admin-type user:
- [x] **List** with available projects
- [x] Interface with all **details** of chosen project
- [x] Interface for **updating** details of chosen project
- [ ] Interface for **creating** new project
- [ ] Interface for **deleting** project

- for Manager-type user:
- [x] **List** with manager's projects
- [x] Interface with all **details** of manager's project
- [x] Interface for **updating** details of manager's project
- [ ] Interface for **deleting** manager's project

### LINKED ISSUE #51: Create tests for projects views
Create necessary tests for managers app views

- [ ] Tests for ProjectsList APIviews

- [ ] Tests for ProjectsDetail APIviews

- [ ] Tests for ProjectsUpdateAPIviews

- [ ] Tests for ProjectsCreate APIviews

- [ ] Tests for ProjectsDelete APIviews

MartynaAnnaGottschling:
Resolves #50 
Resolves #51

### DIFF
--- a/managers/templates/managers/projects_list.html
+++ b/managers/templates/managers/projects_list.html
@@ -1,0 +1,208 @@
+{% extends 'base.html' %}
+{% load i18n %}
+{% block content %}
+{% load rest_framework %}
+
+<h1>{% trans "Projects list" %}
+  <button type="button" class="btn btn-default btn-sm" data-toggle="collapse" data-target="#collapseSort" aria-expanded="false" aria-controls="collapseSort">
+    <span class="glyphicon glyphicon-th-list"></span>
+  </button>
+</h1>
+
+<div class="collapse" id="collapseSort">
+  <div class="col-md-4 col-md-offset-4">
+    <nav id="projects-navbar" class="navbar navbar-light" style="background-color: #e6ecff;">
+      <ul class="nav nav-pills">
+        <li class="nav-item">
+          <a class="nav-link" href="#active" style="color:grey">{% trans "Active" %}</a>
+        </li>
+        <li class="nav-item">
+          <a class="nav-link" href="#terminated" style="color:grey">{% trans "Terminated" %}</a>
+        </li>
+        <li class="nav-item">
+          <a class="nav-link" href="#completed" style="color:grey">{% trans "Completed" %}</a>
+        </li>
+        <li class="nav-item">
+          <a class="nav-link" href="#all" style="color:grey">{% trans "All" %}</a>
+        </li>
+        <li class="nav-item dropdown">
+          <a class="nav-link dropdown-toggle" data-toggle="dropdown" href="#" style="color:grey" role="button" aria-haspopup="true" aria-expanded="false"><span class="glyphicon glyphicon-sort-by-attributes"></span> </a>
+          <div class="dropdown-menu">
+            <a class="dropdown-item" href="{% url 'custom-projects-list' %}?sort=name" style="color:grey">{% trans "Name" %}</a><br>
+            <a class="dropdown-item" href="{% url 'custom-projects-list' %}?sort=start_date" style="color:grey">{% trans "Start date" %}</a><br>
+            <a class="dropdown-item" href="{% url 'custom-projects-list' %}?sort=stop_date" style="color:grey">{% trans "Completed date" %}</a><br>
+            <a class="dropdown-item" href="{% url 'custom-projects-list' %}?sort=members_count" style="color:grey">{% trans "Members count" %}</a><br>
+          </div>
+        </li>
+        <li class="nav-item dropdown">
+          <a class="nav-link dropdown-toggle" data-toggle="dropdown" href="#" style="color:grey" role="button" aria-haspopup="true" aria-expanded="false"><span class="glyphicon glyphicon-sort-by-attributes-alt"></span> </a>
+          <div class="dropdown-menu">
+            <a class="dropdown-item" href="{% url 'custom-projects-list' %}?sort=-name" style="color:grey">{% trans "Name" %}</a><br>
+            <a class="dropdown-item" href="{% url 'custom-projects-list' %}?sort=-start_date" style="color:grey">{% trans "Start date" %}</a><br>
+            <a class="dropdown-item" href="{% url 'custom-projects-list' %}?sort=-stop_date" style="color:grey">{% trans "Completed date" %}</a><br>
+            <a class="dropdown-item" href="{% url 'custom-projects-list' %}?sort=-members_count" style="color:grey">{% trans "Members count" %}</a><br>
+          </div>
+        </li>
+      </ul>
+    </nav>
+  </div>
+</div>
+
+<div data-spy="scroll" data-target="#projects-navbar" data-offset="0">
+
+  <div id="active" class="modal-dialog" style="margin-bottom:0">
+    <button class="btn btn-primary btn-block" type="button" data-toggle="collapse" data-target="#collapseActive" aria-expanded="false" aria-controls="collapseActive">
+      {% trans "Active" %}
+    </button>
+  </div>
+  <div class="panel-collapse collapse in" id="collapseActive">
+    <div class="card card-body">
+      <br/>
+      <div class="container">
+        <div class="row">
+          {% for project in object_list.filter_active %}
+            <div class="col-sm-4">
+              <div class="list-group">
+                  <div class="list-group-item list-group-item-action flex-column align-items-start active">
+                    <div class="d-flex w-100 justify-content-between">
+                      <h4 class="mb-1">{{ project.name }}</h4>
+                    </div>
+                    <p class="mb-1">{% trans "started at" %}: {{project.start_date}}</p>
+                    <small>{% trans "managers" %}:
+                      {% for user in project.managers.all %}
+                                {{user.first_name}} {{user.last_name}}<br>
+                      {% endfor %}</small>
+                    <small><span class="glyphicon glyphicon-user"></span> {{project.members.count}}</small>
+                  </div>
+              </div>
+            </div>
+            {% if forloop.counter|divisibleby:3 %}
+              </div>
+              <div class="row">
+            {% endif %}
+          {% endfor %}
+          </div>
+        </div>
+      </div>
+    </div>
+
+  <div id="terminated" class="modal-dialog" style="margin-bottom:0">
+    <button class="btn btn-info btn-block" type="button" data-toggle="collapse" data-target="#collapseTerminated" aria-expanded="false" aria-controls="collapseTerminated">
+      {% trans "Terminated" %}
+    </button>
+  </div>
+  <div class="collapse" id="collapseTerminated">
+    <div class="card card-body">
+      <br/>
+      <div class="container">
+        <div class="row">
+          {% for project in object_list.filter_terminated %}
+            <div class="col-sm-4">
+              <div class="list-group">
+                <div class="list-group-item list-group-item-info flex-column align-items-start">
+                  <div class="d-flex w-100 justify-content-between">
+                    <h4 class="mb-1">{{ project.name }}</h4>
+                  </div>
+                  <p class="mb-1">{% trans "started at" %}: {{project.start_date}}</p>
+                  <small>{% trans "managers" %}:
+                    {% for user in project.managers.all %}
+                              {{user.first_name}} {{user.last_name}}<br>
+                    {% endfor %}</small>
+                  <small><span class="glyphicon glyphicon-user"></span> {{project.members.count}}</small>
+                </div>
+              </div>
+            </div>
+            {% if forloop.counter|divisibleby:3 %}
+              </div>
+              <div class="row">
+            {% endif %}
+          {% endfor %}
+          </div>
+        </div>
+      </div>
+    </div>
+
+  <div id="completed" class="modal-dialog" style="margin-bottom:0">
+    <button class="btn btn-success btn-block" type="button" data-toggle="collapse" data-target="#collapseCompleted" aria-expanded="false" aria-controls="collapseCompleted">
+      {% trans "Completed" %}
+    </button>
+  </div>
+  <div class="collapse" id="collapseCompleted">
+    <div class="card card-body">
+      <br/>
+      <div class="container">
+        <div class="row">
+          {% for project in object_list.filter_completed %}
+          <div class="col-sm-4">
+            <div class="list-group">
+                <div class="list-group-item list-group-item-action list-group-item-success flex-column align-items-start">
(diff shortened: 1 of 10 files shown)


managers/templates/managers/projects_list.html
```diff
@@ -0,0 +1,209 @@
+{% extends 'base.html' %}
+{% load i18n %}
+{% block content %}
+{% load rest_framework %}
+{% load custom_tags %}
+
+<h1>Projects list
```
rwrzesien: If we are planning to use translations, we have add them in templates too. Please add them here and below.

managers/tests/test_unit_managers_apiview.py
```diff
@@ -0,0 +1,125 @@
+import datetime
+
+from rest_framework.reverse import reverse
+from rest_framework.test import APIRequestFactory
+
+from django.test import TestCase
+
+from users.models import CustomUser
+from managers import views
```
rwrzesien: Alpha ;)

managers/templatetags/custom_tags.py
```diff
@@ -0,0 +1,38 @@
+from django import template
```
rwrzesien: Template tags with database queries should only be used when necessary. At best - never :) All data required by the template frontend should be prepared in backend, to keep template frontend logic at minimum. The same effect can be achieved by using Django custom model manager, but it has advantage that it can be used in both template and Python code, so there is no need to duplicate this functionality.

Definitely `filter_terminated`, `filter_active` and `filter_completed` can be moved as custom manager methods because they don't accept any extra parameters. About the rest I can't tell because they are not used in this pull request, but I think it is doable to, so please do them all as custom manager methods.

Please check https://simpleisbetterthancomplex.com/tips/2016/08/16/django-tip-11-custom-manager-with-chainable-querysets.html for an example how to do this. If you need help please contact me.

managers/tests/test_unit_managers_apiview.py
```diff
@@ -0,0 +1,125 @@
...
+            name='Example Project',
+            start_date=datetime.datetime.now().date() - datetime.timedelta(days=30),
+            stop_date=datetime.datetime.now().date(),
+            terminated=False,
+        )
+        self.project.full_clean()
+        self.project.save()
+        self.project.managers.add(self.user)
+        self.project.members.add(self.user)
+        self.url = [
```
rwrzesien: You are creating a list here, but in test you always refer to it as `self.url[0]`. This can be simplified to be just simple string, and then it can accessed directly in test like `self.url`.

managers/tests/test_unit_managers_apiview.py
```diff
@@ -0,0 +1,125 @@
...
+from django.test import TestCase
+
+from users.models import CustomUser
+from managers import views
+from managers.models import Project
+
+
+class ProjectTest(TestCase):
+    def setUp(self):
+        self.user = CustomUser(
```
rwrzesien: Please add `super().setUp()` so the parent class `setUp` method is called too.

managers/tests/test_unit_managers_apiview.py
```diff
@@ -0,0 +1,125 @@
...
+            email="testuser@codepoets.it",
+            first_name='John',
```
rwrzesien: Please try to keep one way of quotations.

managers/tests/test_unit_managers_apiview.py
```diff
@@ -0,0 +1,125 @@
...
+            last_name='Doe',
+            country='PL',
+            user_type=CustomUser.UserType.ADMIN.name,
+        )
+        self.user.password = 'newuserpasswd'
+        self.user.set_password(self.user.password)
```
rwrzesien: If you call `set_password()`, the value `self.user.password` is overwritten with a hashed password, so there is no need to set it directly in the line above.

managers/views.py
```diff
@@ -1,8 +1,36 @@
...
+
+class ProjectsList(APIView):
+    renderer_classes = [renderers.TemplateHTMLRenderer]
+    template_name = 'managers/projects_list.html'
+
+    def get_queryset(self):
+        return Project.objects.all().order_by('id')
+
+    def get(self, request):
+        projects_queryset = self.get_queryset()
```
rwrzesien: The whole logic of this `get` method is actually selecting queryset, so it should be moved to `get_queryset` method.

users/templates/base.html
```diff
@@ -24,9 +24,13 @@
         <tr><br /></tr>
         <tr>
           <td class="padding"><a href="{% url 'home' %}">{% trans 'Home' %}</a></td>
-          <td class="padding"><a href="{% url 'custom-user-update' user.id %}">Account</a></td>
+          <td class="padding"><a href="{% url 'custom-user-update' user.id %}">{% trans 'Account' %}</a></td>
             {% if user.user_type == "ADMIN" %}
           <td class="padding"><a href="{% url 'custom-users-list' %}">{% trans 'Employees' %}</a></td>
+          <td class="padding"><a href="{% url 'custom-projects-list' %}">{% trans 'Projects list' %}</a></td>
+            {% endif %}
+            {% if user.user_type == "MANAGER" %}
```
rwrzesien: This is a good candidate for custom template tag. This way we could use `"MANAGER"` value directly from Python enum, in case if its value would change in the future.

managers/views.py
```diff
@@ -1,8 +1,36 @@
...
 from managers.models import Project
 from managers.serializers import ProjectSerializer
 
 
 class ProjectViewSet(viewsets.ModelViewSet):
     queryset = Project.objects.all()
     serializer_class = ProjectSerializer
+
+
+class ProjectsList(APIView):
```
rwrzesien: If you implement only `GET` method to retrieve a list, inherit from `rest_framework.generics.ListAPIView` (if there will be any problems after switching to this, please drop me a comment).

managers/views.py
```diff
@@ -1,8 +1,36 @@
...
+    template_name = 'managers/projects_list.html'
+
+    def get_queryset(self):
+        return Project.objects.all().order_by('id')
+
+    def get(self, request):
+        projects_queryset = self.get_queryset()
+        if request.user.user_type == CustomUser.UserType.MANAGER.name:
+            projects_queryset = Project.objects.filter(managers__id=request.user.pk)
+        if request.GET.get('sort'):
```
rwrzesien: You can define sorting using pure rest framework functionality https://www.django-rest-framework.org/api-guide/filtering/#orderingfilter . 

As for `members` and `members_count` it will be tricky, because defined items in `ordering_fields` must exists on queryset results. But on the other hand, we don't want to run `annotate(members_count=Count('members')` if it is not needed, as it results in decent extra SQL work. So you need to check `if 'members' in request.GET.get('sort')` in `get_queryset` method, and then add counting only then.

managers/views.py
```diff
@@ -1,8 +1,36 @@
+from django.db.models import Count
+from rest_framework import renderers
 from rest_framework import viewsets
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from users.models import CustomUser
```
rwrzesien: Alpha ;)

sheetstorm/urls.py
```diff
@@ -18,7 +18,7 @@
 urlpatterns = [
     url(r'^admin/', admin.site.urls),
     url(r'^', include('users.urls')),
-    url(r'^managers/', include('managers.urls')),
+    url(r'^', include('managers.urls')),
```
rwrzesien: I think this should stay, as we discussed on project code review, we should keep endpoints separated.

managers/tests/test_unit_templatetags.py
```diff
@@ -0,0 +1,104 @@
...
+        self.assertEqual(list(result), list(valid_result))
+
+    def test_filter_start_date_templatetag_should_return_only_projects_with_given_start_date(self):
+        projects = Project.objects.all()
+        valid_result = Project.objects.filter(start_date=datetime.datetime.now().date())
+        result = custom_tags.filter_start_date(projects, datetime.datetime.now().date())
+        self.assertEqual(result.count(), valid_result.count())
+        self.assertEqual(list(result), list(valid_result))
+
+    def test_filter_start_date_templatetag_should_return_only_projects_with_given_stop_date(self):
```
rwrzesien: Please update test name `_start_date` => `_stop_date`.

managers/tests/test_unit_templatetags.py
```diff
@@ -0,0 +1,104 @@
+import datetime
```
rwrzesien: After moving template tags as custom manager methods please remove those tests too. Also I don't think there should be extra tests for custom manager methods, as those will be just some simple querysets.

managers/views.py
```diff
@@ -1,8 +1,36 @@
...
 
 
 class ProjectViewSet(viewsets.ModelViewSet):
     queryset = Project.objects.all()
     serializer_class = ProjectSerializer
+
+
+class ProjectsList(APIView):
+    renderer_classes = [renderers.TemplateHTMLRenderer]
+    template_name = 'managers/projects_list.html'
```
rwrzesien: I didn't saw it at first, so I have dropped comment in the following pull request, so I will link it here https://github.com/Code-Poets/sheetstorm/pull/86#discussion_r264067210 .

managers/tests/test_unit_managers_apiview.py
MartynaAnnaGottschling: @rwrzesien I did it this way, because after adding next features (detail, update and so on) this will change into array of url (to store all tested views). I thought that this way of prediction would be good ;)

managers/tests/test_unit_model.py
```diff
@@ -1,48 +1,120 @@
+import datetime
```
kbeker: Please change it into `from datetime import datetime`

managers/tests/test_unit_managers_apiview.py
rwrzesien: @MartynaAnnaGottschling I get this, but those should be separated, independent classes. You can always overwrite this attribute in subclasses and use the same `self.url` in all of them, and it will have different value in each subclass. Or do it like @kbeker proposed.

MartynaAnnaGottschling: Done

managers/tests/test_unit_managers_apiview.py
MartynaAnnaGottschling: Done

managers/tests/test_unit_templatetags.py
MartynaAnnaGottschling: Done

managers/tests/test_unit_managers_apiview.py
```diff
@@ -0,0 +1,125 @@
...
+    def test_projects_list_view_should_show_for_managers_only_own_projects(self):
+        self.user.user_type = CustomUser.UserType.MANAGER.name
+        manager_project_list = Project.objects.filter(managers__id=self.user.pk)
+        request = APIRequestFactory().get(path=self.url[0])
+        request.user = self.user
+        response = views.ProjectsList.as_view()(request)
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, self.project.name)
+        projects_list = response.data['projects_queryset']
+        self.assertEqual(list(manager_project_list), list(projects_list))
```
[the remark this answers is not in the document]
MartynaAnnaGottschling: @kbeker Yes, because without that the test returns
`AssertionError: <QuerySet [<Project: Example Project>]> != <QuerySet [<Project: Example Project>]>`
I don't know why exactly.

managers/tests/test_unit_model.py
MartynaAnnaGottschling: Done

managers/tests/test_unit_managers_apiview.py
MartynaAnnaGottschling: @rwrzesien You mean alphabetical order? Without grouping them for exterior ones, django and from each app?

managers/tests/test_unit_managers_apiview.py
MartynaAnnaGottschling: Done

managers/tests/test_unit_model.py
```diff
@@ -1,48 +1,120 @@
...
+
+class TestProjectModelField(BaseModelTestCase):
+    model_class = Project
+    required_input = {
+        'name': 'Example Project',
+        'start_date': datetime.datetime.now().date(),
+    }
+
+    def setUp(self):
+        self.manager = create_user_using_full_clean_and_save(
```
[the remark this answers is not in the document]
MartynaAnnaGottschling: Done

managers/templates/managers/projects_list.html
MartynaAnnaGottschling: Done

managers/tests/test_unit_managers_apiview.py
MartynaAnnaGottschling: Done

managers/tests/test_unit_managers_apiview.py
MartynaAnnaGottschling: Done

managers/views.py
MartynaAnnaGottschling: Done

managers/templatetags/custom_tags.py
MartynaAnnaGottschling: Done

managers/tests/test_unit_templatetags.py
MartynaAnnaGottschling: Done

users/templates/base.html
maciejSamerdak: https://memegenerator.net/img/instances/80356275/consider-it-done.jpg

sheetstorm/urls.py
maciejSamerdak: Done

managers/views.py
maciejSamerdak: Any more sophisticated API view than the basic one requires extra attributes and/or greatly limits view's flexibility.
Since we will be dropping REST anyway, I'll be switching it to Django's generic ListView. It causes far less problems anyway.

maciejSamerdak: Done

managers/views.py
maciejSamerdak: Done

managers/views.py
maciejSamerdak: Done

PR #89, opened 2019-03-08
Feature project delete view

### LINKED ISSUE #50: Manager Base Interface

### LINKED ISSUE #51: Create tests for projects views

MartynaAnnaGottschling:
Resolves #50
Resolves #51


managers/tests/test_unit_managers_function_views.py
```diff
@@ -0,0 +1,74 @@
...
+            name='Example Project',
+            start_date=datetime.datetime.now().date() - datetime.timedelta(days=30),
+            stop_date=datetime.datetime.now().date(),
+            terminated=False,
+        )
+        self.project.full_clean()
+        self.project.save()
+        self.project.managers.add(self.user)
+        self.project.members.add(self.user)
+        self.url = reverse('custom-project-delete', args=(self.project.pk,))
```
kbeker: I think that it would be better to keep this url as a string

MartynaAnnaGottschling: Done

managers/views.py
```diff
@@ -38,10 +40,67 @@ def get_queryset(self):
...
+                'serializer': project_serializer,
+                'project': project,
+                'errors': project_serializer.errors,
+            })
+        project_serializer.save()
+        return redirect('custom-project-detail', pk=pk)
+
+
+def delete_project(request, pk):
+    if request.user.user_type == CustomUser.UserType.ADMIN.name:
```
rwrzesien: @Karrp @MartynaAnnaGottschling Question about how this should exactly work - here implementation shows that only `ADMIN` should be allowed to perform this action. In template, everyone can click on this (which is a little bit miselading, I think it should be shown only for admins), however the real issue is that admins doesn't have access Projects List from the menu.

maciejSamerdak: Hey, might be late to respond but unless you already applied some fixes, I can navigate to project list from menu, while signed in as admin.

And since all delete operations (except for individual reports) should be performed only by admin, you are right that the delete button should be visible only to admin.

MartynaAnnaGottschling: But the delete button is in update template, which is accessable from detail template, only for admin, I think...

rwrzesien: @MartynaAnnaGottschling @maciejSamerdak I think I was able to reach it as manager too. Anyway, the point is that if it should be doable only by admin, only admin should be able to click on it - it might be bad user experience if there will be a button visible for users that when clicked will redirect them to `home` without telling them what actually happened. So we should agree on who can do and see what - for me both ways make sense (admin and manager who created this project or just admin), but whatever we choose lets be consistent.

We will probably solve more of the permissions cases in the future so lets just focus on this one now.

maciejSamerdak: @Karrp says only admin should be able to delete a project.

PR #112, opened 2019-04-03
Chore improve projects programming tools

kbeker:
There is no issue for that. I have added few improvements to our programming tools. They should make our job easier :)

### DIFF
--- a/style.sh
+++ b/style.sh
@@ -1,0 +1,9 @@
+#!/bin/bash -e
+
+printf "[ISORT: sheetstorm]\n"
+./find-files-to-refactor.sh | xargs isort -sl
+printf "\n"
+
+printf "[BLACK: sheetstorm]\n"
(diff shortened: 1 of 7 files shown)


style.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
+
+printf "[ISORT: sheetstorm]\n"
+find-files-to-refactor.sh | xargs isort -sl
+printf "\n"
+
+printf "[BLACK: sheetstorm]\n"
+find-files-to-refactor.sh | xargs black --line-length 120
```
rwrzesien: Why just finding them instead of instantly fixing them? :)

kbeker: @rwrzesien You mean find them all first, and then without `xargs` command fix them all? :)

kbeker: Or just apply `black` to all files in repository?

PR #113, opened 2019-04-03
Add `.idea/` dir to .gitignore

kbeker:
There is no issue for that.


Karrp:
We added '.idea/' to .git/info/exclude. It works like .gitignore, but just on your PC.

Issue #109, opened 2019-04-02
Few problems with Markdown


Szymiks:
Markdown in Sheet storm project doesn't have all options like:
- [x] - checkbox
- [x] - when we use lists there is a problem with linebreaks. It creates strange gaps.

PR #87, opened 2019-03-08
Feature project update view

MartynaAnnaGottschling:
Resolves #50
Resolves #51

PR #117, opened 2019-04-05
Bug fix markdown problems

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/109
Now I think that markdown works better, especially text display looks much better. I've changed framework which is used and I added extensions which allow to use checkboxes, tables and few more options.

### DIFF
--- a/employees/models.py
+++ b/employees/models.py
@@ -1,7 +1,8 @@
 from django.core.validators import MaxValueValidator
 from django.core.validators import MinValueValidator
 from django.db import models
-from markdown_deux import markdown
+from markdown import markdown
+from markdown_checklists.extension import ChecklistsExtension
 
 from employees.common.constants import ReportModelConstants
 from employees.common.strings import MAX_HOURS_VALUE_VALIDATOR_MESSAGE
@@ -34,5 +35,8 @@ class Report(models.Model):
         return self.work_hours.to_eng_string().replace(".", ":")
 
     @property
-    def markdown_description(self) -> markdown:
-        return markdown(self.description)
+    def markdown_description(self) -> str:
+        return markdown(
+            self.description,
+            extensions=["extra", "sane_lists", "wikilinks", "nl2br", "legacy_em", ChecklistsExtension()],
+        )
--- a/employees/templates/employees/report_list.html
+++ b/employees/templates/employees/report_list.html
@@ -3,7 +3,6 @@
 {% block content %}
 {% load rest_framework %}
 {% load static %}
-
 <head>
     <link rel="stylesheet" type="text/css" href="{% static 'employees/style.css' %}">
     <link rel="stylesheet" href="//code.jquery.com/ui/1.12.1/themes/smoothness/jquery-ui.css">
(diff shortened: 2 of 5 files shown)


employees/models.py
```diff
@@ -57,4 +57,5 @@ def work_hours_str(self):
 
     @property
     def markdown_description(self):
-        return markdown(self.description)
+        return markdown(self.description, extensions=['extra', 'sane_lists', 'wikilinks',
```
rwrzesien: Reformat using black ;)

employees/serializers.py
```diff
@@ -49,8 +49,8 @@ class Meta:
             'date',
             'project',
             'author',
-            'description',
             'work_hours',
+            'description',
```
rwrzesien: Why this change?

employees/templates/employees/report_list.html
```diff
@@ -23,7 +22,8 @@ <h1>{{ UI_text.PAGE_TITLE.value }}</h1>
     <form action="{% url 'custom-report-list' %}" method="POST">
         {% csrf_token %}
         {% render_form serializer template_pack='rest_framework/vertical' %}
-        <input type="submit" value="{{ UI_text.CREATE_REPORT_BUTTON.value }}">
+        <a href="https://guides.github.com/features/mastering-markdown/" target="_blank">Markdown is supported</a><br/>
+       <input type="submit" value="{{ UI_text.CREATE_REPORT_BUTTON.value }}">
```

employees/templates/employees/report_list.html
```diff
@@ -62,7 +62,7 @@ <h1>{{ UI_text.PAGE_TITLE.value }}</h1>
                     {{ report.work_hours_str }}
                 </td>
                 <td class="description-column">
-                    {{ report.markdown_description|safe|linebreaksbr }}
+                {{ report.markdown_description|safe }}
```
rwrzesien: Please indent :)

employees/serializers.py
Szymiks: I changed it because  I wanted to show below description field that it is supported by Markdown. I didn't know how to resolve it better, as form is rendering by serializer so I changed it like that. Maybe there is better solution. :)

rwrzesien: You can add `helptext` to model field, check this:
https://github.com/Code-Poets/sheetstorm/pull/87/files#diff-dd359dbd34d8fff0ddd29ecb0703e366R33

and then it gets automatically displayed when rendering whole form. But this is optional, as I guess we are going to change all our templates soon right? Just FYI that it can be achieved without having to put it at the end.

Szymiks: Ok so I will delete this, because as you said templates will be changed so I think it is not very necessary. Thanks!

employees/models.py
Szymiks: Done

employees/templates/employees/report_list.html
Szymiks: Done

employees/templates/employees/report_list.html
Szymiks: Done

PR #119, opened 2019-04-08
Bugfix missing mypy in requirements lock

kbeker:
Requirements installed form requirements-dev was too long and caused problems with mypy and pylint. So I decided to delete them and add all requirements to one file which is `requirements.lock`. Please clean your current virutal-env requirements and install them again from this file.


requirements-development.txt
```diff
@@ -1,41 +0,0 @@
-astroid==2.0.4
```
Karrp: Why did you delete this file?

Karrp: Ok i found explanation in pull request description

PR #130, opened 2019-04-15
Chore add admin report serializer

maciejSamerdak:
Introduces serializer for administrative purposes which will be used by both project-report-detail and admin-report-detail introduced in upcoming PRs.

**Do not delete this branch until #131 and #133 are merged**

PR #133, opened 2019-04-15
Feature project report list

maciejSamerdak:
Introduces list view for all reports from selected project.

Related #130 

**Do not delete this branch until #135 and #134 are merged**

### DIFF
--- a/employees/templates/employees/project_report_list.html
+++ b/employees/templates/employees/project_report_list.html
@@ -6,7 +6,7 @@
 
 <link rel="stylesheet" type="text/css" href="{% static 'employees/style.css' %}">
 
-<h1>{{ project_name }}{{ UI_text.PAGE_TITLE.value }}</h1>
+<h1>{{ object.name }}{{ UI_text.PAGE_TITLE.value }}</h1>
 <div class="container">
     <div class="table-responsive">
         <table class="table">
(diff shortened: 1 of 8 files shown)


employees/views.py
```diff
@@ -220,3 +222,21 @@ def form_valid(self, form: AdminReportForm) -> HttpResponseRedirectBase:
         self.object.editable = True
         self.object.save()
         return super().form_valid(form)
+
+
+@method_decorator(login_required, name="dispatch")
+class ProjectReportList(ListView):
```
rwrzesien: After looking at this finally, I think you were right in the version - it should be `DetailView`, with model Project, and reports should taken in template by some Project model method to apply ordering. This method could look like this:

```
def get_reports_ordered(self):
    return self.report_set.order_by("-date", "project__name")
```

Please only remember to have it already fetched using `.select_related` in `get_queryset()` method.

employees/views.py
```diff
@@ -220,3 +222,21 @@ def form_valid(self, form: AdminReportForm) -> HttpResponseRedirectBase:
...
+    template_name = "employees/project_report_list.html"
+    model = Report
+
+    def get_queryset(self) -> QuerySet:
+        return Report.objects.filter(project__id=self.kwargs["pk"]).order_by("-date", "project__name")
+
+    def get_context_data(self, **kwargs: Any) -> dict:
+        context = super().get_context_data(**kwargs)
+        context["UI_text"] = ProjectReportListStrings
+        context["project_name"] = get_object_or_404(Project, pk=self.kwargs["pk"])
```
rwrzesien: When https://github.com/Code-Poets/sheetstorm/pull/133/files#r283077801 is applied you don't need this, because you have project as `object`.

employees/views.py
```diff
@@ -220,3 +222,21 @@ def form_valid(self, form: AdminReportForm) -> HttpResponseRedirectBase:
...
+        return context
+
+    def get_success_url(self) -> str:
+        return reverse("project-report-list", kwargs={"pk": self.object.project.id})
```
rwrzesien: Please remember to update this too.

employees/templates/employees/project_report_list.html
```diff
@@ -0,0 +1,59 @@
...
+                <th class="project-reports-author-header">{{ UI_text.AUTHOR_COLUMN_HEADER.value }}</th>
+                <th class="project-reports-work-hours-header">{{ UI_text.WORK_HOURS_COLUMN_HEADER.value }}</th>
+                <th class="project-reports-description-header">{{ UI_text.DESCRIPTION_COLUMN_HEADER.value }}</th>
+                <th class="project-reports-creation-date-header">{{ UI_text.CREATION_DATE_COLUMN_HEADER.value }}</th>
+                <th class="project-reports-last-update-header">{{ UI_text.LAST_UPDATE_COLUMN_HEADER.value }}</th>
+                <th class="project-reports-editable-header">{{ UI_text.EDITED_COLUMN_HEADER.value }}</th>
+                <td class="project-reports-edit-button-header Invisible"></td>
+            </tr>
+            <tbody>
+                {% regroup object_list by date as reports_by_date %}
```
rwrzesien: Here you would use the `get_reports_ordered` method.

rwrzesien:
@kbeker Looks good but please change to DetailView with Project as model, it looks more natural in this case.

employees/views.py
kbeker: Done

employees/views.py
kbeker: Done

employees/views.py
kbeker: Done

employees/templates/employees/project_report_list.html
kbeker: Done

PR #134, opened 2019-04-15
Feature project report detail

maciejSamerdak:
Introduces report editing view linked to **project-report-list**.

Related #133

### DIFF
--- a/employees/templates/employees/project_report_detail.html
+++ b/employees/templates/employees/project_report_detail.html
@@ -15,8 +15,12 @@
     <form action="{% url 'project-report-detail' pk=report.pk %}" method="POST">
         {% csrf_token %}
         {{ form|crispy }}
-        <input type="submit" value="{{ UI_text.UPDATE_REPORT_BUTTON.value }}">
-        <input type="submit" name="discard" value="{{ UI_text.DISCARD_CHANGES_BUTTON.value }}">
+        <button type="submit" class="btn btn-success">
+            <span class="glyphicon glyphicon-ok"></span> {{ UI_text.UPDATE_REPORT_BUTTON.value }}
+        </button>
+        <a href="{% url 'project-report-list' pk=report.project.pk %}" class="btn btn btn-default">
+            <span class="glyphicon glyphicon-repeat"></span> {{ UI_text.DISCARD_CHANGES_BUTTON.value }}
+        </a>
     </form>
 </div>
 {% endblock %}
--- a/employees/templates/employees/project_report_detail.html
+++ b/employees/templates/employees/project_report_detail.html
@@ -1,0 +1,26 @@
+{% extends 'base.html' %}
+
+{% load i18n %}
+{% load static %}
+{% load crispy_forms_tags %}
+
+{% block extra_head %}
+<link rel="stylesheet" type="text/css" href="{% static 'employees/style.css' %}">
+{% endblock %}
+
+{% block content %}
+<h1>{{ UI_text.PAGE_TITLE.value }}{{ report.project }} ({{ report.date }})</h1>
+</br>
+<div class="modal-dialog">
+    <form action="{% url 'project-report-detail' pk=report.pk %}" method="POST">
+        {% csrf_token %}
+        {{ form|crispy }}
+        <input type="submit" value="{{ UI_text.UPDATE_REPORT_BUTTON.value }}">
+        <input type="submit" name="discard" value="{{ UI_text.DISCARD_CHANGES_BUTTON.value }}">
+    </form>
+</div>
+{% endblock %}
+
+{% block extra_script %}
+{{ form.media }}
+{% endblock %}
--- a/employees/tests/test_unit_report_viewset.py
+++ b/employees/tests/test_unit_report_viewset.py
@@ -11,7 +11,6 @@ from employees.common.strings import ProjectReportListStrings
 from employees.common.strings import ReportListStrings
 from employees.models import Report
 from employees.models import TaskActivityType
-from employees.views import ProjectReportDetail
 from employees.views import ProjectReportList
 from employees.views import ReportDetail
 from employees.views import ReportList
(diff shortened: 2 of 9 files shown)


maciejSamerdak:
Due to lack of admin's serializer, I changed this view to a Django-based generic view, in resemblance to AdminReportView.

employees/tests/test_unit_report_viewset.py
```diff
@@ -642,3 +643,100 @@ def test_project_report_list_get_queryset_method_should_return_queryset_containi
...
+
+    def test_project_report_detail_view_should_display_report_details(self):
+        request = APIRequestFactory().get(path=reverse("project-report-detail", args=(self.report.pk,)))
+        request.user = self.user
+        response = ProjectReportDetail.as_view()(request, pk=self.report.pk)
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, self.report.description)
+        self.assertEqual(response.data["serializer"].instance, self.report)
+
+    def test_project_report_detail_view_should_not_be_accessible_for_unauthenticated_users(self):
```
rwrzesien: I think this test is not needed, `login_decorator` is working well :) Lets just test our custom logic.

employees/tests/test_unit_report_viewset.py
```diff
@@ -642,3 +643,100 @@ def test_project_report_list_get_queryset_method_should_return_queryset_containi
...
+        request = APIRequestFactory().get(path=reverse("project-report-detail", args=(self.report.pk,)))
+        request.user = AnonymousUser()
+        response = ProjectReportDetail.as_view()(request, pk=self.report.pk)
+        self.assertEqual(response.status_code, 403)
+
+    def test_project_report_detail_view_should_not_render_non_existing_report(self):
```
rwrzesien: Same here, routing should work OK by default too, I think this test is not necessary.

employees/tests/test_unit_report_viewset.py
```diff
@@ -642,3 +643,100 @@ def test_project_report_list_get_queryset_method_should_return_queryset_containi
         self.assertFalse(other_project_report in queryset)
         self.assertEqual(queryset[0], other_report_1)
         self.assertEqual(queryset[1], self.report)
+
+
+class ProjectReportDetailTests(TestCase):
+    def setUp(self):
+        self.user = CustomUser(
```
rwrzesien: Please use `UserFactory`.

employees/tests/test_unit_report_viewset.py
```diff
@@ -642,3 +643,100 @@ def test_project_report_list_get_queryset_method_should_return_queryset_containi
...
+            email="testuser@codepoets.it", password="newuserpasswd", first_name="John", last_name="Doe", country="PL"
+        )
+        self.user.full_clean()
+        self.user.save()
+
+        self.project = Project(name="Test Project", start_date=datetime.datetime.now())
+        self.project.full_clean()
+        self.project.save()
+
+        self.report = Report(
```
rwrzesien: Please user `ReportFactory`. Then you don't have to create `self.project` too.

employees/templates/employees/project_report_detail.html
```diff
@@ -0,0 +1,26 @@
...
+
+{% block content %}
+<h1>{{ UI_text.PAGE_TITLE.value }}{{ report.project }} ({{ report.date }})</h1>
+</br>
+<div class="modal-dialog">
+    <form action="{% url 'project-report-detail' pk=report.pk %}" method="POST">
+        {% csrf_token %}
+        {{ form|crispy }}
+        <input type="submit" value="{{ UI_text.UPDATE_REPORT_BUTTON.value }}">
+        <input type="submit" name="discard" value="{{ UI_text.DISCARD_CHANGES_BUTTON.value }}">
```
rwrzesien: Is this being handled by Django by default? 

Generally I don't like making a POST request just to **not** make any changes. I think from UX perspective it is simpler to "go back" or "cancel" or something like that. I would suggest replacing it with a link back.

employees/tests/test_unit_report_viewset.py
```diff
@@ -642,3 +643,100 @@ def test_project_report_list_get_queryset_method_should_return_queryset_containi
...
+            description="Some description",
+            author=self.user,
+            project=self.project,
+            work_hours=Decimal("8.00"),
+        )
+        self.report.full_clean()
+        self.report.save()
+
+    def test_project_report_detail_view_should_display_report_details(self):
+        request = APIRequestFactory().get(path=reverse("project-report-detail", args=(self.report.pk,)))
```
rwrzesien: Please use `self.client` that is delivered with Django `TestCase`. You can simply make `self.client.force_login(self.user)` to make it authenticated (ie. in `setUp()`). You can take a look how it is used in other refactored tests for Django views.

employees/tests/test_unit_report_viewset.py
```diff
@@ -642,3 +643,100 @@ def test_project_report_list_get_queryset_method_should_return_queryset_containi
...
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(current_description, new_description)
+        self.assertTrue(self.report.editable)
+
+    def test_project_report_detail_view_should_not_update_report_on_discard(self):
+        new_description = "Some other description"
+        request = APIRequestFactory().post(
+            path=reverse("project-report-detail", args=(self.report.pk,)),
+            data={
+                "date": datetime.datetime.now().date(),
```
rwrzesien: If its reused (or part of it is), add it to `setUp()` like `self.data`.

rwrzesien:
@maciejSamerdak Looks nice, please just adapt the test and please explain me how the "discard" works :)

employees/tests/test_unit_report_viewset.py
maciejSamerdak: You think so? This way at least we know that the view isn't protected anymore if anything goes wrong. Or are there other, better ways to test that?

employees/templates/employees/project_report_detail.html
maciejSamerdak: I'm happy to inform you that everything is in absolute order and this solution in fact does not work at all. 😄 

Not only does the test fail, I also cannot see any other button except for **Update** in rendered view. I guess I must have skipped the unit tests on this branch, during the hassle of multiple rebases.

maciejSamerdak: Well, I'm happy to inform you that it does not work at all. 😁 

The _discard_ button just works as another submit, changes apply. I'll fix it up in a jiff.

maciejSamerdak:
@rwrzesien I added requested changes. I also refactored unit tests a little and moved them to `test_views`. However, I'd like to highlight one issue: after refactoring the tests, they performance dropped significantly. It's like a double-action revolver vs a Tommy gun.

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Done

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Done

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Done

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Done

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Done

employees/templates/employees/project_report_detail.html
maciejSamerdak: Done

employees/tests/test_unit_report_viewset.py
rwrzesien: Right, lets keep it then.

rwrzesien:
@maciejSamerdak 
> It's like a double-action revolver vs a Tommy gun.

Nice comparison :) Well that is because previously we was running view directly as a function. Now it is sending request through all other stuff like middleware, routing etc. - basically in the same way as you would send a request to Django server running. It takes slower **but** also behaves more like a regular HTTP call, allowing us to test/discover more stuff (for example if any other layer has impact on the view).

rwrzesien:
@maciejSamerdak Looks good, tests are much more compact now, this is exactly what I have meant. If you want add back the test `test_project_report_detail_view_should_not_be_accessible_for_unauthenticated_users` but if you don't want its fine for me too.

maciejSamerdak:
I recovered the login-required test, just in case. When integration tests are introduced, covering this mechanism in greater detail, it could be removed.

kbeker:
Accidentally closed. Merged

PR #135, opened 2019-04-15
Feature project report list username grouping

maciejSamerdak:
Introduces reports per user grouping in **project-report-list** view.

Related #133 

**Do not delete this branch until #136 is merged**

### DIFF
--- a/employees/views.py
+++ b/employees/views.py
@@ -21,6 +21,7 @@ from rest_framework.views import APIView
 
 from employees.common.strings import AdminReportDetailStrings
 from employees.common.strings import AuthorReportListStrings
+from employees.common.strings import ProjectReportListStrings
 from employees.common.strings import ReportDetailStrings
 from employees.common.strings import ReportListStrings
 from employees.forms import AdminReportForm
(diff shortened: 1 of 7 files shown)


employees/views.py
```diff
@@ -174,13 +174,21 @@ class ProjectReportList(APIView):
     permission_classes = (permissions.IsAuthenticated,)
 
     @classmethod
-    def get_queryset(cls, pk: int) -> QuerySet:
-        return Report.objects.filter(project=pk).order_by("-date")
+    def get_queryset(cls, project_pk: int, author_pk: int) -> QuerySet:
```
rwrzesien: This method should not accept any extra parameters because it might be used in other places in this view and they will not be aware of that. Instead of that do in `get()` `self.project = get_object_or_404(Project, pk=pk)` and here use `self.project.pk` and ...

...`user.pk`, that comes from a different query. But why? It can be done in single query. Reports should be accessible directly from CustomUser like `user.report_set.all` or `user.report_set.all()` depending if you are using it in template or python code. So it is enough to get `project.members.all()`. But you can also do this directly in template. So it is only enough to pass `Project` object to html template, and get everything what is needed from it then.

But probably you paid attention and remember that I have said that data should be prepared in backend and queries should not be done in template. That applies here too. That is why you can use `select_related` (https://docs.djangoproject.com/pl/2.1/ref/models/querysets/#select-related) to fetch for all required related data in the backend.

Please treat it as exercise on how you can limit amount of database queries from N to 1 and complexity of the view from building complex mapping structure to passing single object to template.

maciejSamerdak: In case of this particular view class (**APIView**, extending Django's **View**), which is the lowest level, it has no _get_queryset_ method of its own, so this is in fact it's definition and thus I have full control over it's usage in the view. This is the reason why I allowed myself to deviate from convention and modify it a little to suit my needs.

To be honest, I completely forgot about the `select_related` mechanism during development, despite it being quite obvious and me using it frequently in previous project. I haven't thought of it for some reason, so thanks for the feedback along with providing documentation highlighting it's performance benefits. 😄 

So if I understand correctly, it's perfectly acceptable to pass single object (Project) to template and **_THEN_** call it's related objects with aforementioned method _**IN**_ the template itself?

rwrzesien: > So if I understand correctly, it's perfectly acceptable to pass single object (Project) to template and THEN call it's related objects with aforementioned method IN the template itself?

It is already evaluated, because it was retrieved using `.get()`.

maciejSamerdak:
Works like a charm, at the expense of so little code. I still can't get over it! 😂

rwrzesien:
> Works like a charm, at the expense of so little code. I still can't get over it! 

The less code doing more, the better :)

rwrzesien:
@kbeker We could merge it first to https://github.com/Code-Poets/sheetstorm/pull/133/ and then apply change to DetailView in that branch.

kbeker:
Accidentally closed. Merged

PR #137, opened 2019-04-15
Feature hours per day counter

maciejSamerdak:
This adds sum total of work hours from all reports for each day. It is displayed in **custom-report-list** and **project-report-list** at the bottom of each day section.

Related #136 

**Do not delete this branch until #138 is merged**


rwrzesien:
@maciejSamerdak Could you write in few words what this pull request adds? I don't know if I understand it correctly but I think it should sum amount of hours from given reports list for each day - is that correct?

maciejSamerdak:
Correct. Is current description sufficient?

employees/templates/employees/report_list.html
```diff
@@ -52,8 +52,34 @@ <h1>{{ UI_text.PAGE_TITLE.value }}</h1>
...
+                            {% if report.editable %}
+                            <a href="{% url 'custom-report-detail' pk=report.id %}" class="btn btn-info">
+                                <span class="glyphicon glyphicon-pencil"></span>
+                            </a>
+                            {% endif %}
+                        </td>
+                    </tr>
+                    {% else %}
+                <tr>
+                    <td style="width:230px"><strong>{{ UI_text.HOURS_PER_DAY_LABEL.value }}</strong></td>
```
rwrzesien: I think setting `width` this way won't work, it will get aligned to other cells - or other cells will be aligned to this one.

rwrzesien:
@maciejSamerdak 
I have a plan on how to handle this one:
1. Remove `query_as_dict` function, it is bad to the bone :)

2. Where `query_as_dict` was used, for now just use `self.get_queryset()`, rename `reports_dict` to `reports`.

3. In `reports_list.html` template use `regroup by date` like in https://github.com/Code-Poets/sheetstorm/blob/master/employees/templates/employees/author_report_list.html#L33.

4. Get work hours sum per day using SQL query. But this where the tricky part starts - I don't think it can be done in a single query with just Django querysets. So I suggest to do this in two. The first stays the same, the seconds looks like this:
`Report.objects.all().extra({'date_created' : "date(date)"}).values('date_created').annotate(created_count=Sum('work_hours'))`

You can return a dict directly from it like this:
```
dict(Report.objects.all().extra({'date_created' : "date(date)"}).values('date_created').annotate(created_count=Sum('work_hours')).values_list('date_created','created_count'))                                                                                             
Out[43]: 
{datetime.date(2019, 1, 13): Decimal('31.89'),
 datetime.date(2019, 1, 14): Decimal('37.01')}
```

So you can pass it into template and get sum by date. This should be ReportManager method.

5. If you want to format the date in template, use https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#date.

If you can't do this or don't have time please let me know, I can do this but wanted to leave it to you to see how it can be simplified.

maciejSamerdak:
I've entirely rewritten the entire feature according to suggested changes. Report list view, along with it's template, went through some refactoring as well.

employees/templatetags/data_structure_element_selectors.py
```diff
@@ -0,0 +1,14 @@
...
+from typing import Dict
+
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def get_key_value(dictionary: Dict[Any, Any], index: Any) -> Any:
+    try:
```
rwrzesien: Why not `return dictionary.get(index, "")` ? And I think its faster by some micro time ;)

employees/tests/test_unit_templatetags.py
```diff
@@ -0,0 +1,16 @@
...
+
+class GetKeyValueTests(TestCase):
+    def setUp(self):
+        self.dictionary = {"one": 1, "two": 2, "three": 3}
+
+    def test_get_key_value_function_should_return_an_element_of_given_dict_at_given_key(self):
+        result = get_key_value(self.dictionary, "two")
+        self.assertEqual(result, 2)
+
+    def test_get_key_value_function_should_handle_key_error(self):
```
rwrzesien: If you apply https://github.com/Code-Poets/sheetstorm/pull/137/files#r284477996 pelase remember to change test name.

employees/views.py
```diff
@@ -45,25 +45,26 @@ def perform_create(self, serializer: ReportSerializer) -> None:
...
 
 
-def query_as_dict(query_set: QuerySet) -> Dict[str, Any]:
-    dictionary = {}  # type: Dict[str, Any]
-    for record in query_set:
-        key = record.date
-        dictionary.setdefault(key, [])
-        dictionary[key].append(record)
-    return dictionary
+def daily_hours_sum(queryset: QuerySet) -> dict:
```
rwrzesien: This should be `ReportManager` method. We don't have one yet so please create one for Reports. Here is more info on it https://docs.djangoproject.com/en/2.2/topics/db/managers/#custom-managers. It will be a nice exercise too.

rwrzesien:
@maciejSamerdak We are almost there, most important thing left is making this queryset as `ReportManager` method.

employees/views.py
maciejSamerdak: I have a couple of questions, I'd like to get out of the way first:
1. As far as I understand, a manager should return a QuerySet object. This means I shouldn't create a function returning dictionary. Shall I then cast each call in the view with `dict()`? Wouldn't this be less clean  and convenient?
2. QuerySets used in this calculation may vary from view to view. Should I define each set in the manager as methods? Could I use parameters in manager's methods?

This solution seems somewhat restrictive, should I follow conventions. I get why it's worth to move them to a manager, but at the moment I'm not sure how to properly handle this one.

maciejSamerdak:
I've moved the hours counting function to a **QuerySet** in `models.py` in similar fashion to #207.

I've also managed to break BLACK and ISORT somehow, because they both claim to reformat `employees/tests/test_unit_templatetags.py` every single time **full-check** is executed, but don't apply any changes whatsoever.

employees/templates/employees/report_list.html
```diff
@@ -38,45 +41,46 @@ <h1>{{ UI_text.PAGE_TITLE.value }}</h1>
...
-                <td class="edit-button-column Invisible hidden-print">
-                    {% if report.editable %}
-                    <a href="{% url 'custom-report-detail' pk=report.id %}" class="btn btn-info">
-                        <span class="glyphicon glyphicon-pencil"></span>
-                    </a>
-                    {% endif %}
-                </td>
-            </tr>
+                    <td class="project-column">{{ report.project }}</td>
+    ﻿               <td class="task-activities-column">{{ report.task_activities }}</td>
```

employees/views.py
rwrzesien: > As far as I understand, a manager should return a QuerySet object. This means I shouldn't create a function returning dictionary. Shall I then cast each call in the view with dict()? Wouldn't this be less clean and convenient?

You don't have to return `QuerySet` from `Manager`, it can be anything (just as you did). The wider concept of that is that there is `models.QuerySet` and `models.Manager`. Their responsibilities are well described here https://docs.djangoproject.com/en/2.2/topics/db/managers/. But also as you can find in this doc, it says

> In lieu of the above approach which requires duplicating methods on both the QuerySet and the Manager, QuerySet.as_manager() can be used to create an instance of Manager with a copy of a custom QuerySet’s methods:

This is what we do here for simplicity. I have also find this to be common practice on most of the projects I have seen (I mean I have never seen in production code methods being split between `models.QuerySet` and `models.Manager`.

> QuerySets used in this calculation may vary from view to view. Should I define each set in the manager as methods? Could I use parameters in manager's methods?

Yes, you can pass parameters. I think the answer for first part of the question is "it depends on a case". Sometimes its better to write separate method, sometimes its better to parametrize. You need to follow your common sense, experience and project specifics.

employees/templatetags/data_display_filters.py
```diff
@@ -0,0 +1,8 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def decimal_to_hours(data):
+    return str(data).replace(".", ":")
```
rwrzesien: @Karrp @kbeker @maciejSamerdak 
By looking at this I have just realized we have a quite serious bug in the code.
We are holding working hours as decimal. It mean that when you create a report and set "0:30" (which I would understand as "half hour") in form, it stores value 0.30. When you set a value "0.5" (which I would as "half hour" too), you store value 0.50, but display 0:50.

So when I create a report and add "1:30" to my working hours, and another "1:30" in another report, how much to I get in total? Decimal 2.60.

This would be a serious issue when calculating sums etc. Now I am starting to understand why @maciejSamerdak was creating those complex transitions in him pull requests.

So we should ASAP change the field type of `work_hours` to `DurationField` https://docs.djangoproject.com/en/2.2/ref/models/fields/#durationfield. I have created an issue for that and I think it should have top priority.

employees/views.py
maciejSamerdak: Thanks for detailed response. Previously I might've written few counter-intuitive functions, BECAUSE I took to much liberty. So I wanted to be sure my implementation wouldn't go against common practices.

But then I saw your code which you sent me and my doubts evaporated. 😃

PR #139, opened 2019-04-15
Feature month segregation for report list

maciejSamerdak:
**custom-report-list** view now accepts a month and year parameters to separate reports by each month. A navigation bar is introduced, containing next and previous button to navigate between adjacent months, as well as form allowing to select desired month and year to jump to.

Related #138


rwrzesien:
@maciejSamerdak Could you please add description on what is this pull request doing?

maciejSamerdak:
Description updated. Feature is correspondent to #138. I provided full description here again to not necessitate looking into another PR.

employees/views.py
```diff
@@ -162,69 +179,73 @@ def _create_project_join_form(self) -> ProjectJoinForm:
...
+        self.reports = self.get_queryset()
+        self.context_data = self.get_context_data(**kwargs)
+
+    def get(self, _request: HttpRequest, year: int, month: int) -> Response:  # pylint: disable=unused-argument
+        return Response(self.context_data)
+
+    def post(  # type: ignore
+        self, request: HttpRequest, year: int, month: int
+    ) -> Union[Response, HttpResponseRedirectBase]:
+        year = int(year)
```
dybi: this seems awkward - `year` and `month` are already declared as  `int`s in the function's signature. Why should we be casting them?

maciejSamerdak: I just checked this and the parameters provided are indeed in string format. This causes an error as soon as `_default_date()` is called. I do wonder myself why there are no warnings raised regarding integer typing, if in reality a string is returned. 
Zrzut ekranu 2019-05-29 o 16 45 26

users/templates/base.html
```diff
@@ -76,6 +78,7 @@
             {% if user.user_type == manager or user.user_type == admin %}
             <li>
                 {% if request.path == custom_projects_list_url %}
+                <!--highlight won't work now because of date parameters-->
```
maciejSamerdak: I don't know what to do about this. In order to have a tab highlighted, a full URL is checked. But the exact parameters are available in sub-template, so we cannot access them from here. Perhaps there'd be some way to reliably check for a part of URL? Or maybe something like a bool value that would be passed from the rendered view?

Anyway, should I solve this bug with this PR or just leave it for now and add an issue for it, so we can deal with it later?

employees/tests/test_unit_report_viewset.py
```diff
@@ -431,11 +431,64 @@ def test_serializer_project_field_queryset_should_contain_only_projects_user_is_
...
         request = APIRequestFactory().post(path=self.url, data={"join": "join"})
         request.user = self.user
-        response = ReportList.as_view()(request)
+        response = ReportList.as_view()(request, **self.view_kwargs)
         project_field_choices = response.data["serializer"].fields["project"].choices.values()
         self.assertTrue(self.project.name in project_field_choices)
         self.assertFalse(new_project.name in project_field_choices)
 
+    def test_custom_report_list_view_should_redirect_to_another_month_if_month_switch_was_called_on_post(self):
+        request = APIRequestFactory().post(path=self.url, data={"date": "09-2020", "month-switch": "month-switch"})
```
rwrzesien: If https://github.com/Code-Poets/sheetstorm/pull/274 gets merged first, please remember to update tests, otherwise @Szymiks please remember to refactor those too.

employees/views.py
rwrzesien: For url resolving Python regex lib is used, so the matched object is string.

employees/views.py
```diff
@@ -158,34 +166,48 @@ def redirect_to_current_month(self) -> HttpResponseRedirectBase:
     ),
     name="dispatch",
 )
-class ReportList(UserIsManagerOfCurrentReportProjectMixin, UserIsAuthorOfCurrentReportMixin, APIView):
+class ReportList(
```
rwrzesien: Same here about getting along with https://github.com/Code-Poets/sheetstorm/pull/274

Now when I see how this view grew, I think we should separate creating from displaying list. But this is just my opinion, which is driven by the fears it will be harder to maintain the current version. But it is rather a business decision, so please just take my opinion into consideration.

users/templates/base.html
rwrzesien: Aren't the available in `request` object which should be accessible in this template too?

employees/views.py
maciejSamerdak: I also think this might be a good idea. It was convenient at first, but now it's getting harder to maintain and develop. This way, we could also connect together all the list-related views with single parent class, as @dybi suggested. This way, I could also just move all the functionality of **MonthSegregationMixin** to such view (since we still need to overload `get()` and `post()`  in each view to get it to work 100%).

users/templates/base.html
kbeker: There were more same problems as it in SheetStorm. I fixed them added template_tag.

employees/views.py
kbeker: I personally think that firstly we should merge everything what is necessary and in next step we start to change code without changing functionalities. There are 3 PRs left to merge to start refactoring code as @rwrzesien suggested at start. So now I will merge it and after we merge all Maciek's pull requests we should all have little talk how we want to do this. @rwrzesien @dybi @maciejSamerdak @MartynaAnnaGottschling @Szymiks @Karrp

employees/tests/test_unit_report_viewset.py
kbeker: @Szymiks please be aware of this that I'm merging it now.

kbeker:
This is merged, but before I merged I forgot to change base branch so I am forced to close this PR because this is already in master.

employees/views.py
rwrzesien: > I personally think that firstly we should merge everything what is necessary and in next step we start to change code without changing functionalities.

Off course, this is what I have meant too.

> we should all have little talk how we want to do this

Please add me to a meeting, if I will be able I would like to participate.

PR #140, opened 2019-04-15
Feature hours per month counter

maciejSamerdak:
Introduces total hours per month counter for all report list views.

Related #138

### DIFF
--- a/employees/views.py
+++ b/employees/views.py
@@ -57,6 +57,7 @@ class ReportList(APIView):
     permission_classes = (permissions.IsAuthenticated,)
     hide_join = False
     daily_hours_sum = None  # type: QuerySet
+    monthly_hours_sum = None  # type: QuerySet
 
     def get_queryset(self) -> QuerySet:
         logger.info(f"User with id: {self.request.user.pk} get reports queryset")
@@ -92,6 +93,7 @@ class ReportList(APIView):
         super().initial(request, *args, **kwargs)
         self.reports = self.get_queryset()
         self.daily_hours_sum = self.reports.order_by().get_work_hours_sum_for_all_dates()
+        self.monthly_hours_sum = self.reports.order_by().get_work_hours_sum_for_all_authors()
 
     def get(self, _request: HttpRequest) -> Response:
         logger.info(f"User with id: {self.request.user.pk} get to the ReportList view")
@@ -100,6 +102,7 @@ class ReportList(APIView):
                 "serializer": self._create_serializer(),
                 "object_list": self.reports,
                 "daily_hours_sum": self.daily_hours_sum,
+                "monthly_hours_sum": self.monthly_hours_sum,
                 "UI_text": ReportListStrings,
                 "project_form": self._create_project_join_form(),
                 "hide_join": self.hide_join,
@@ -122,6 +125,7 @@ class ReportList(APIView):
                     "serializer": reports_serializer,
                     "object_list": self.reports,
                     "daily_hours_sum": self.daily_hours_sum,
+                    "monthly_hours_sum": self.monthly_hours_sum,
                     "UI_text": ReportListStrings,
                     "project_form": self._create_project_join_form(),
                     "hide_join": self.hide_join,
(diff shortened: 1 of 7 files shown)


employees/views.py
```diff
@@ -62,10 +62,22 @@ def hours_per_day_counter(reports: list) -> Decimal:
     return hours_sum
 
 
+def hours_per_month_counter(dictionary: dict) -> Decimal:
```
rwrzesien: I think this could be fully achieved in queryset.

employees/views.py
```diff
@@ -219,7 +235,7 @@ def include_users_in_reports_dict(self, project: Project, year: int, month: int)
             queryset = self.get_queryset(project_pk=project.pk, author_pk=user.pk, year=year, month=month)
             user_reports_dict = query_as_dict(queryset)
             key = user.email
-            reports_dict[key] = user_reports_dict
+            reports_dict[key] = [user_reports_dict, decimal_to_hours_string(hours_per_month_counter(user_reports_dict))]
```
rwrzesien: This is fundamentally wrong because there should be no need to iterate through members and calculate hours for each separately, it should be also done in pull request, but lets not over complicate it now unless you want to refactor the whole method?

rwrzesien:
@maciejSamerdak 
Please try to apply the same approach as in https://github.com/Code-Poets/sheetstorm/pull/137#issuecomment-491627064

If you can't do this or don't have time please let me know, I can do this but wanted to leave it to you to see how it can be simplified.

employees/views.py
```diff
@@ -55,6 +55,14 @@ def daily_hours_sum(queryset: QuerySet) -> dict:
     )
 
 
+def monthly_hours_sum(queryset: QuerySet) -> dict:
```
rwrzesien: Please also make this a `ReportManager` method, but I would like to understand it correctly - in this queryset you want to sum of all work hours for each author for each month? Because what it currently does is getting sum of all work hours for each author but from the whole queryset - if queryset contains data from more then a month, it is calculated for that whole time period, without dividing into months, only by author.

rwrzesien:
@maciejSamerdak Same here, please make it a `ReportManager` method and lets clarify what the queryset should do exactly.

maciejSamerdak:
I've moved hours counting function to **ReportManager** in same fashion as in #137.

There's still one issue. Currently this feature is also introduced to the regular **report list**...
But it won't work correctly, until #139 is merged. Currently, this sums up work hours from all reports displayed on page, so it will work properly as soon as month segregation is introduced.

Should I keep it in case #139 is merged first? Should I delete it otherwise?
If it should stay this way, should I also introduce this feature to **author report list**? This one doesn't even have segregation feature in development yet.

employees/views.py
rwrzesien: @maciejSamerdak What about that ^ ?

rwrzesien:
> There's still one issue. Currently this feature is also introduced to the regular report list...
But it won't work correctly, until #139 is merged. Currently, this sums up work hours from all reports displayed on page, so it will work properly as soon as month segregation is introduced.

Lets wait until everything else what is needed is merged.

maciejSamerdak:
@Karrp wanted this to be introduced with 0.3.0. I've changed the base to master already.

rwrzesien:
@maciejSamerdak Please add test proving that `monthly_hours_sum` method will sum hours grouped by months.

employees/views.py
maciejSamerdak: Correct, this issue would be resolved with next PRs, which add date filtering in list views showing only one month on page. Having that in mind I simplified this function, since the queryset provided in the view would already be filtered. Again, this feature is **fast-forwarded**, but @Karrp wanted it introduced ASAP, knowing this issue.

maciejSamerdak:
This feature does indeed count hours for entirety of queryset rather than just one month. This is because it was originally meant to be introduced after #138 where views pass querysets that are already filtered. Knowing this issue @Karrp wants it introduced with 0.3.0 due to popular demand.

On @kbeker request, I described it in an issue #239 After each list view is split per month (currently 2 in #138 and #139), we can consider it resolved.

kbeker:
After talk with @rwrzesien I'm dismissing his `request changes` from this PR. Everything explained above

PR #145, opened 2019-04-18
Chore refactor singup view

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/144

Template doesn't look good but I just created it to test, I have already done in PR my singup interface, after review I just need to add few things in new interface.

### DIFF
--- a/users/common/exceptions.py
+++ b/users/common/exceptions.py
@@ -11,7 +11,3 @@ class SheetStormBaseException(Exception):
         self.error_code = error_code
         self.error_message = "" if error_message is None else error_message
         super().__init__(error_message)
-
-
-class CustomValidationError(SheetStormBaseException):
-    pass
--- a/users/common/utils.py
+++ b/users/common/utils.py
@@ -2,12 +2,11 @@ import logging
 import random
 import string
 
-from rest_framework import serializers
+from django.core.exceptions import ValidationError
 
 from users.common import constants
 from users.common.constants import ErrorCode
-from users.common.exceptions import CustomValidationError
-from users.common.strings import CustomValidationErrorText
+from users.common.strings import ValidationErrorText
 
 logger = logging.getLogger(__name__)
 
@@ -20,49 +19,26 @@ def generate_random_phone_number(length):
     return "".join(random.choices(string.digits, k=length))
 
 
-def raise_serializer_exception(error_message):
-    logger.warning(f"Serializer raised exception with message: {error_message}")
-    raise serializers.ValidationError(error_message)
-
-
-def raise_custom_validation_exception(error_message, error_code):
-    logger.warning(f"Custom valid exception with message: {error_message} and error code: {error_code}")
-    raise CustomValidationError(error_message, error_code)
-
-
-def prepare_proper_exception(call_instance, error_message, error_code):
-    if "serializers" in str(type(call_instance)):
-        logger.debug(
-            f"Serializer: {call_instance} caught exception with message: {error_message} and error_code: {error_code}"
-        )
-        raise_serializer_exception(error_message)
-    else:
-        raise_custom_validation_exception(error_message, error_code)
-
-
-def custom_validate_email_function(call_instance, email):
-    logger.info(f"Custom validate email function with email: {email} sent from: {type(call_instance)}")
+def custom_validate_email_function(email):
+    logger.info(f"Custom validate email function with email: {email}")
     if email.count("@") == 1:
         if email.split("@")[0] in ["", " "]:
-            logger.debug(f"Empty first part in email")
-            prepare_proper_exception(
-                call_instance,
-                CustomValidationErrorText.VALIDATION_ERROR_EMAIL_MALFORMED_FIRST_PART,
-                ErrorCode.CREATE_USER_EMAIL_MISSING,
+            logger.warning(f"Empty first part in email")
+            raise ValidationError(
+                message=ValidationErrorText.VALIDATION_ERROR_EMAIL_MALFORMED_FIRST_PART,
+                code=ErrorCode.CREATE_USER_EMAIL_MISSING,
             )
 
         else:
             if email.split("@")[1] not in constants.VALID_EMAIL_DOMAIN_LIST:
-                logger.debug(f"Invalid email domain: {email}")
-                prepare_proper_exception(
-                    call_instance,
-                    CustomValidationErrorText.VALIDATION_ERROR_EMAIL_MESSAGE_DOMAIN,
-                    ErrorCode.CREATE_USER_EMAIL_DOMAIN,
+                logger.warning(f"Invalid email domain: {email}")
+                raise ValidationError(
+                    message=ValidationErrorText.VALIDATION_ERROR_EMAIL_MESSAGE_DOMAIN,
+                    code=ErrorCode.CREATE_USER_EMAIL_DOMAIN,
                 )
     else:
-        logger.debug(f"Invalid email: {email}. It should consist of name.surname@codepoets.it")
-        prepare_proper_exception(
-            call_instance,
-            CustomValidationErrorText.VALIDATION_ERROR_EMAIL_AT_SIGN_MESSAGE,
-            ErrorCode.CREATE_USER_EMAIL_SIGN_MISSING,
+        logger.warning(f"Invalid email: {email}. It should consist of name.surname@codepoets.it")
+        raise ValidationError(
+            message=ValidationErrorText.VALIDATION_ERROR_EMAIL_AT_SIGN_MESSAGE,
+            code=ErrorCode.CREATE_USER_EMAIL_SIGN_MISSING,
         )
--- a/users/forms.py
+++ b/users/forms.py
@@ -1,7 +1,9 @@
+from django import forms
 from django.contrib.auth.forms import UserChangeForm
 from django.contrib.auth.forms import UserCreationForm
 from django_countries.widgets import CountrySelectWidget
 
+from users.common import constants
 from users.models import CustomUser
 
 
@@ -27,3 +29,11 @@ class CustomUserChangeForm(UserChangeForm):
         model = CustomUser
         fields = "__all__"
         widgets = {"country": CountrySelectWidget()}
+
+
+class CustomUserSignUpForm(UserCreationForm):
+    email = forms.EmailField(max_length=constants.EMAIL_MAX_LENGTH, required=True, label="id_email")
+
+    class Meta:
+        model = CustomUser
+        fields = ("email", "password1", "password2")
--- a/users/templates/accounts/success-registration.html
+++ b/users/templates/accounts/success-registration.html
@@ -1,0 +1,12 @@
+{% extends 'base.html' %}
+{% load i18n %}
+{% load static %}
+{% block content %}
+<head>
+    <script type="text/javascript" src="{% static 'users/scripts/redirect_to_home_page.js' %}"></script>
+</head>
+<h1>{{ MESSAGES.CONGRATULATIONS.value }}</h1>
+<p>{{ MESSAGES.ACCOUNT_CREATED.value }}</p>
+<p>{{ MESSAGES.REDIRECTION_INFO.value }}</p>
+<a href="{% url 'home' %}" class="btn btn-default">{{ MESSAGES.OKAY_BUTTON.value }}</a>
+{% endblock %}
--- a/users/templates/signup.html
+++ b/users/templates/signup.html
@@ -1,24 +1,30 @@
 {% extends 'base.html' %}
 
 {% load i18n %}
-{% load rest_framework %}
 
 {% block content %}
 <h2>{% trans 'Sign up' %}</h2>
 
 <div class="modal-dialog" style="margin-bottom:0">
-  <form action="{% url 'signup' %}" method="post">
-      {% csrf_token %}
-      {% render_form serializer template_pack='rest_framework/vertical' %}
-
-      {% if non_field_errors %}
-      <div class="form-group has-error">
-          {% for error in non_field_errors %}
-            <span class="help-block">{{ error }}</span><br>
-          {% endfor %}
+    <form action="{% url 'signup' %}" method="post">
+        {% csrf_token %}
+        <div class="fieldWrapper">
+            <label for="{{ form.email.label }}">{% trans 'Email' %}</label>
+                {{ form.email }}
+                {% include 'partials/errors_in_field.html' with errors=form.email.errors %}
+                {% include 'partials/errors_in_field.html' with errors=form.non_field_errors %}
         </div>
-      {% endif %}
-      <button type="submit" class="btn btn-default">Sign in</button>
-  </form>
+        <div class="fieldWrapper">
+            <label for="{{ form.password1.label }}">{% trans 'Password' %}</label>
+                {{ form.password1 }}
+        </div>
+        <div class="fieldWrapper">
+            <label for="{{ form.password2.label }}">{% trans 'Password confirmation' %}</label>
+                {{ form.password2 }}
+                {% include 'partials/errors_in_field.html' with errors=form.password2.errors %}
+
+        </div>
+        <button type="submit" class="btn btn-default">{% trans 'Sign in' %}</button>
+    </form>
 </div>
 {% endblock %}
--- a/users/tests/test_unit_users_apiview.py
+++ b/users/tests/test_unit_users_apiview.py
@@ -157,10 +157,10 @@ class SignUpTests(TestCase):
                 "email": self.user.email,
                 "first_name": self.user.first_name,
                 "last_name": self.user.last_name,
-                "password": self.user.password,
-                "password_confirmation": self.user.password,
+                "password1": self.user.password,
+                "password2": self.user.password,
             },
         )
         self.assertEqual(response.status_code, 302)
-        self.assertEqual(response.url, "/accounts/login/")
+        self.assertEqual(response.url, reverse("success-signup"))
         self.assertEqual(CustomUser.objects.all().count(), 1)
--- a/users/tests/test_utils_functions.py
+++ b/users/tests/test_utils_functions.py
@@ -4,7 +4,7 @@ from parameterized import parameterized
 
 import users.common.utils
 import users.common.validators
-from users.common.strings import CustomValidationErrorText
+from users.common.strings import ValidationErrorText
 
 
 class TestCustomModelEmailValidationFunction(TestCase):
--- a/users/tests/test_utils_functions.py
+++ b/users/tests/test_utils_functions.py
@@ -1,60 +1,21 @@
+from django.core.exceptions import ValidationError
 from django.test import TestCase
-from rest_framework import serializers
+from parameterized import parameterized
 
 import users.common.utils
 import users.common.validators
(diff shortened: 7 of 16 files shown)


users/views.py
```diff
@@ -92,30 +92,15 @@ def index(request: HttpRequest) -> HttpResponse:
...
-                {
-                    "serializer": serializer,
-                    "errors": serializer.errors,
-                    "non_field_errors": serializer.errors.get("non_field_errors"),
-                }
-            )
-        else:
-            serializer.save(request)
-            return redirect("login")
+def signup(request: HttpRequest) -> HttpResponse:
```
rwrzesien: Use subclass of Django `LoginView`.

users/common/exceptions.py
```diff
@@ -1,17 +1,15 @@
 from typing import Optional
 from typing import Union
 
+from django.core.exceptions import ValidationError
+
 from users.common.constants import ErrorCode
 
 
-class SheetStormBaseException(Exception):
+class CustomValidationError(ValidationError):
```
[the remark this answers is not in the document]
Szymiks: So in our validator I should rasie just `Validation Error`?

users/forms.py
```diff
@@ -27,3 +31,28 @@ class Meta:
         model = CustomUser
         fields = "__all__"
         widgets = {"country": CountrySelectWidget()}
+
+
+class CustomUserSignUpForm(UserCreationForm):
+    email = forms.EmailField(max_length=constants.EMAIL_MAX_LENGTH, required=True, label="id_email")
+    password1 = forms.CharField(help_text="", label="id_password1", widget=forms.PasswordInput())
```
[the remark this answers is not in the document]
Szymiks: I am not sure if I understood. You mean to add some translatable text for error message? Or to show translatable help text?

users/forms.py
```diff
@@ -27,3 +31,28 @@ class Meta:
...
+        user.email = self.cleaned_data["email"]
+
+        if commit:
+            user.save()
+
+    def clean_email(self) -> str:
+        email = self.cleaned_data["email"]
+        try:
+            custom_validate_email_function(self, email)
+        except CustomValidationError as error:
```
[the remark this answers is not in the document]
Szymiks: Done

users/templates/signup.html
```diff
@@ -6,19 +6,37 @@
...
-          {% for error in non_field_errors %}
-            <span class="help-block">{{ error }}</span><br>
-          {% endfor %}
-        </div>
-      {% endif %}
-      <button type="submit" class="btn btn-default">Sign in</button>
-  </form>
+    <form action="{% url 'signup' %}" method="post">
+        {% csrf_token %}
+<div class="fieldWrapper">
```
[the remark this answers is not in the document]
Szymiks: Done

users/views.py
Szymiks: I didn't found LoginView subclass of Django to SignUp but I found FormView and wrote it in generic way as I think you wanted.

users/common/exceptions.py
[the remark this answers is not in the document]
rwrzesien: Standard `ValidationError` should do in this case, as we are using it in context of standard Django mechanisms.

users/forms.py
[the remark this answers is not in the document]
rwrzesien: For help text describing the password rules. And it should be wrapped it `gettext()` to make it translatable. 

But the password rules are defined in Django settings, so maybe there is some smart way to generate this help text directly from those rules? Then it won't have to be updated each time when rules change.

users/views.py
rwrzesien: It is here https://github.com/django/django/blob/master/django/contrib/auth/views.py#L38 . You should reuse as much of it as possible, basically I think that it should be enough to:
- override this view class.
- set custom template in overridden class.
- override `get_context_data` if any extra context data is needed.
- set custom form inheriting from `AuthenticationForm` if any modification is needed.

Sorry for late description on this but I wasn't sure how much we want to achieve with the custom sign up view, and now I have checked related issue that basically we want to get rid of `APIView`. So I think we should use `LoginView` as much as possible and override only what is needed.

users/forms.py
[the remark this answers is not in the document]
Szymiks: We are using default Django validators so all rules are stored in django mechanism, what is more they are returned as `ngettext()` so I think they could be translatable. If we would like to change something in those validators, for instance change minimum of password length we should do something like this 
```
'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
        'OPTIONS': {
            'min_length': 9,
        }
```
However we could also create our own validator with helptext included and add it to settings in the same way as other.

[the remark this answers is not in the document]
rwrzesien: So is it possible to put the returned value here into help text too?

users/views.py
```diff
@@ -92,30 +93,13 @@ def index(request: HttpRequest) -> HttpResponse:
...
-                    "errors": serializer.errors,
-                    "non_field_errors": serializer.errors.get("non_field_errors"),
-                }
-            )
-        else:
-            serializer.save(request)
-            return redirect("login")
+    def form_valid(self, form: CustomUserSignUpForm) -> Union[Response, HttpResponseRedirectBase]:
+        form.save()
+        return redirect("home")
```
rwrzesien: Maybe we should add some other view here saying that account was created and you can login now? @Karrp @kbeker  what do you think?

users/forms.py
```diff
@@ -27,3 +31,28 @@ class Meta:
...
+        user.email = self.cleaned_data["email"]
+
+        if commit:
+            user.save()
+
+    def clean_email(self) -> str:
+        email = self.cleaned_data["email"]
+        try:
+            custom_validate_email_function(self, email)
+        except ValidationError as error:
```
rwrzesien: This doesn't make sense, you are catching exception just to reraise it, let it be raised directly in `custom_validate_email_function` call.

users/forms.py
```diff
@@ -27,3 +31,28 @@ class Meta:
...
+    password1 = forms.CharField(help_text="", label="id_password1", widget=forms.PasswordInput())
+    password2 = forms.CharField(help_text="", label="id_password2", widget=forms.PasswordInput())
+
+    class Meta:
+        model = CustomUser
+        fields = ("email", "password1", "password2")
+
+    def save(self, commit: bool = True) -> None:
+        user = super(CustomUserSignUpForm, self).save(commit=False)
+        user.email = self.cleaned_data["email"]
```
rwrzesien: This is not needed, email should be already set on user.

users/tests/test_unit_users_apiview.py
```diff
@@ -182,10 +182,10 @@ def test_signup_view_should_add_new_user_on_post(self):
...
                 "last_name": self.user.last_name,
-                "password": self.user.password,
-                "password_confirmation": self.user.password,
+                "password1": self.user.password,
+                "password2": self.user.password,
             },
         )
         self.assertEqual(response.status_code, 302)
-        self.assertEqual(response.url, "/accounts/login/")
+        self.assertEqual(response.url, "/")
```
rwrzesien: User `reverse('home')` here.

users/forms.py
Szymiks: Done

users/forms.py
Szymiks: Done

users/tests/test_unit_users_apiview.py
Szymiks: Done

users/views.py
Szymiks: So I am waiting for answer from them.

Karrp: I think there should be some success info.
Other view should be ok.
Or maybe popup would be better? And redirect after few seconds or pressing "OK" button.

users/forms.py
rwrzesien: Why is it still there then? :)

rwrzesien:
@Szymiks Looks good now, please just resolve redirect after sign up case.

users/views.py
kbeker: I think that popup with this info would be nice :)

users/templates/accounts/success-registration.html
```diff
@@ -0,0 +1,13 @@
+{% extends 'base.html' %}
+{% load i18n %}
+{% load rest_framework %}
```
kbeker: Why you loading here `rest_framework`? I think that it unnecessary here.

users/templates/accounts/success-registration.html
```diff
@@ -0,0 +1,13 @@
...
+{% load static %}
+{% block content %}
+<head>
+    <script type="text/javascript" src="{% static 'users/scripts/redirect_to_home_page.js' %}"></script>
+</head>
+<h1>{% trans "Congratulations!" %}</h1>
+<p>{% trans "Your account has been successfully created! Now you can sign in!" %}</p>
+<p>{% trans "You will be redirected in few seconds to the login site or press a button to make it faster." %}</p>
+<a href="{% url 'home' %}" class="btn btn-default">Okay!</a>
+{% endblock %}
```
kbeker: You missed blank line here

users/templates/signup.html
```diff
@@ -7,18 +7,38 @@
...
+        <div class="fieldWrapper">
+            <label for="{{ form.email.label }}">Email:</label>
+                {{ form.email }}
+                {% if form.non_field_errors or form.email.errors %}
+                        <div class="alert alert-danger">
+                            {% for error in form.email.errors %}
+                                {{ error }}
+                            {% endfor %}
+                            {% for err in form.non_field_errors %}
+                                {{ err }}
```
kbeker: Please don't make name shortcuts.

kbeker:
@Szymiks Also pylint gave output:
```
************* Module users.serializers
users/serializers.py:88:4: R0201: Method could be a function (no-self-use)
```

users/templates/accounts/success-registration.html
Szymiks: Removed

users/templates/signup.html
Szymiks: Done

users/templates/accounts/success-registration.html
Szymiks: done

users/common/exceptions.py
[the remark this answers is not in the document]
Szymiks: done

Szymiks:
> @Szymiks Also pylint gave output:
> 
> ```
> ************* Module users.serializers
> users/serializers.py:88:4: R0201: Method could be a function (no-self-use)
> ```

Done but I make it static because when I make it functional one test doesn't pass.

users/views.py
```diff
@@ -97,33 +99,18 @@ def index(request: HttpRequest) -> HttpResponse:
...
-                    "serializer": serializer,
-                    "errors": serializer.errors,
-                    "non_field_errors": serializer.errors.get("non_field_errors"),
-                }
-            )
-        else:
-            serializer.save(request)
-            return redirect("login")
+
+class SuccessView(TemplateView):
```
rwrzesien: Please make this view name more descriptive, like `UserSignUpSuccess` or something similar.

users/tests/test_utils_functions.py
```diff
@@ -1,60 +1,56 @@
...
 
     def test_user_should_hold_email_with_valid_domain(self):
         email = "testuser@invaliddomain.com"
-        with self.assertRaises(serializers.ValidationError) as serializer_exception:
-            users.common.utils.custom_validate_email_function(CustomRegisterSerializer(), email)
+        with self.assertRaises(ValidationError) as exception:
+            users.common.utils.custom_validate_email_function(email)
         self.assertTrue(
-            CustomValidationErrorText.VALIDATION_ERROR_EMAIL_MESSAGE_DOMAIN_SHORT in str(serializer_exception.exception)
+            CustomValidationErrorText.VALIDATION_ERROR_EMAIL_MESSAGE_DOMAIN_SHORT in str(exception.exception)
```
rwrzesien: It is not `CustomValidationError` text anymore.

users/tests/test_utils_functions.py
```diff
@@ -1,60 +1,56 @@
...
 
 import users.common.utils
 import users.common.validators
-from users.common.exceptions import CustomValidationError
 from users.common.strings import CustomValidationErrorText
-from users.serializers import CustomRegisterSerializer
 
 
 class TestCustomModelEmailValidationFunction(TestCase):
     def test_user_email_should_hold_at_least_one_at_sign(self):
```
rwrzesien: You should use `parametrized` for test methods in this class as they are not different from tests in `TestCustomSerializerEmailValidationFunction` class. Check here https://github.com/Code-Poets/sheetstorm/blob/master/managers/tests/test_views.py#L117.

users/templates/signup.html
```diff
@@ -1,24 +1,43 @@
...
-          {% for error in non_field_errors %}
-            <span class="help-block">{{ error }}</span><br>
-          {% endfor %}
+    <form action="{% url 'signup' %}" method="post">
+        {% csrf_token %}
+        <div class="fieldWrapper">
+            <label for="{{ form.email.label }}">Email:</label>
+                {{ form.email }}
+                {% if form.non_field_errors or form.email.errors %}
+                        <div class="alert alert-danger">
```
rwrzesien: Here you are wrapping whole errors list in div, but here https://github.com/Code-Poets/sheetstorm/pull/145/files#diff-0d040cecf685f46b2f769b9db182b639R34 you are wrapping each error separately. Please unify the way you do it and extract this to separate partial template and include it https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#include.

users/common/utils.py
```diff
@@ -20,49 +19,29 @@ def generate_random_phone_number(length):
     return "".join(random.choices(string.digits, k=length))
 
 
-def raise_serializer_exception(error_message):
-    logger.warning(f"Serializer raised exception with message: {error_message}")
-    raise serializers.ValidationError(error_message)
-
-
 def raise_custom_validation_exception(error_message, error_code):
```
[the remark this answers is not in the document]
Szymiks: Done

users/forms.py
```diff
@@ -27,3 +29,18 @@ class Meta:
...
+
+
+class CustomUserSignUpForm(UserCreationForm):
+    email = forms.EmailField(max_length=constants.EMAIL_MAX_LENGTH, required=True, label="id_email")
+
+    class Meta:
+        model = CustomUser
+        fields = ("email", "password1", "password2")
+
+    def save(self, commit: bool = True) -> CustomUser:
```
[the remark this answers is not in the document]
Szymiks: Done

users/templates/accounts/success-registration.html
```diff
@@ -0,0 +1,13 @@
+{% extends 'base.html' %}
+{% load i18n %}
+{% load rest_framework %}
+{% load static %}
+{% block content %}
+<head>
+    <script type="text/javascript" src="{% static 'users/scripts/redirect_to_home_page.js' %}"></script>
+</head>
+<h1>{% trans "Congratulations!" %}</h1>
```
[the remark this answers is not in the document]
Szymiks: Done

users/views.py
Szymiks: Done

users/tests/test_utils_functions.py
Szymiks: done

users/templates/signup.html
Szymiks: done

users/tests/test_utils_functions.py
Szymiks: This change makes changes in many files, I think that I should change commit name after rebase, is that right @rwrzesien ?

users/tests/test_utils_functions.py
Szymiks: I am not sure if I understood, because I have never used the parameterized test before, but I tried to implement it and it is ready to look at it, now :).

users/templates/signup.html
```diff
@@ -1,24 +1,29 @@
...
 <h2>{% trans 'Sign up' %}</h2>
 
 <div class="modal-dialog" style="margin-bottom:0">
-  <form action="{% url 'signup' %}" method="post">
-      {% csrf_token %}
-      {% render_form serializer template_pack='rest_framework/vertical' %}
+    <form action="{% url 'signup' %}" method="post">
+        {% csrf_token %}
+        <div class="fieldWrapper">
+            <label for="{{ form.email.label }}">Email:</label>
```
rwrzesien: Please add translation to strings in this file or move it to enum too.

users/templates/accounts/success-registration.html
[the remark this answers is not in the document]
rwrzesien: OK but please check if you have done it everywhere.

users/tests/test_utils_functions.py
rwrzesien: If you think this change is not really related to this commit, go ahead and create new one for it.

users/tests/test_utils_functions.py
rwrzesien: Oh wow, I was rather thinking about combining this class and `TestCustomSerializerEmailValidationFunction` but you take a different way and its nice too. But please take a loon on tests in `TestCustomSerializerEmailValidationFunction` - doesn't they do exactly the same now?

users/templates/signup.html
rwrzesien: Not exactly what I have meant, please create one universal mini-template for this (with generic name) and use it for each field in a form. Those included templates can take parameters, so you can do something like `{% include 'partials/errors_in_field.html'  with erorrs=form.password2.errors %}`

rwrzesien:
Almost there, the only things left in my opinion are:
https://github.com/Code-Poets/sheetstorm/pull/145#discussion_r285800157
https://github.com/Code-Poets/sheetstorm/pull/145#discussion_r285805728
https://github.com/Code-Poets/sheetstorm/pull/145#discussion_r285797564  + https://github.com/Code-Poets/sheetstorm/pull/145#discussion_r285797858

users/tests/test_utils_functions.py
Szymiks: Yes they are exactly the same, so should I delete those test ? We will not use serializers any more and what is more tests looks the same. For now I will remove this class `TestCustomSerializerEmailValidationFunction`

users/templates/signup.html
Szymiks: Done

users/templates/accounts/success-registration.html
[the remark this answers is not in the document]
Szymiks: `Okay` string  added, too :)

users/templates/signup.html
Szymiks: Done

users/tests/test_utils_functions.py
rwrzesien: Do not remove yet, we will remove it along with serializer, for now lets keep it.

users/templates/signup.html
rwrzesien: Almost good, but why you are iterating twice over the same list, one inside another? :) I think once should be enough, or am I missing something?

Szymiks: done

PR #157, opened 2019-04-26
Chore add logs to manager app

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/154

I have add few methods, because I couldn't add any logs when something would happen

### DIFF
--- a/managers/models.py
+++ b/managers/models.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Any
 from typing import Set
 
@@ -11,6 +12,8 @@ from django.utils.translation import gettext_lazy as _
 from managers.commons.constants import MAX_NAME_LENGTH
 from managers.commons.constants import MESSAGE_FOR_CORRECT_DATE_FORMAT
 from users.models import CustomUser
+
+logger = logging.getLogger(__name__)
 
 
 class ProjectQuerySet(models.QuerySet):
@@ -47,6 +50,7 @@ class Project(models.Model):
 def update_user_type(sender: Project, action: str, pk_set: Set, **kwargs: Any) -> None:
     assert sender == Project.managers.through
     project = kwargs["instance"]
+    logger.debug(f"Updates on project with id: {sender.pk} for users with id: {pk_set}")
     if action in ["pre_remove", "post_remove"]:
         change_user_type_to_employee(pk_set)
     elif action in ["pre_add", "post_add"]:
@@ -57,10 +61,15 @@ def update_user_type(sender: Project, action: str, pk_set: Set, **kwargs: Any) -
 
 def change_user_type_to_manager(project: Project) -> None:
     for manager in project.managers.all():
+        logger.debug(f"User with user type: {manager.user_type} with id: {manager.pk} in managers")
         if manager.user_type == CustomUser.UserType.EMPLOYEE.name:
+            logger.debug(
+                f"User with user type: {manager.user_type} and id: {manager.pk} is in managers for project with id: {project.pk}"
+            )
             manager.user_type = CustomUser.UserType.MANAGER.name
             manager.full_clean()
             manager.save()
+            logger.debug(f"User with id: {manager.pk} have changed user type to manager")
         else:
             continue
 
@@ -69,8 +78,11 @@ def change_user_type_to_employee(pk_set: Set) -> None:
     for user_id in pk_set:
         user = CustomUser.objects.get(pk=user_id)
         if not user.manager_projects.exists() and user.user_type != CustomUser.UserType.ADMIN.name:
+            logger.info(f"User: {user} user type change to employee")
             user.user_type = CustomUser.UserType.EMPLOYEE.name
             user.full_clean()
             user.save()
+            logger.info(f"User with id: {user.pk} have changed user type to employee")
         else:
+            logger.debug(f"User with id: {user.pk} has not changed to employee")
             continue
--- a/managers/views.py
+++ b/managers/views.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Any
 from typing import Type
 from typing import Union
@@ -23,6 +24,13 @@ from managers.models import Project
 from users.models import CustomUser
 from utils.decorators import check_permissions
 
+logger = logging.getLogger(__name__)
+
+
+class ProjectViewSet(viewsets.ModelViewSet):
+    queryset = Project.objects.all().order_by(Lower("name"))
+    serializer_class = ProjectSerializer
+
 
 @method_decorator(login_required, name="dispatch")
 @method_decorator(
@@ -44,6 +52,7 @@ class ProjectsListView(ListView):
     queryset = Project.objects.all()
 
     def get_queryset(self) -> QuerySet:
+        logger.debug(f"Get project query set for user with id: {self.request.user.pk}")
         if self.request.user.user_type == CustomUser.UserType.ADMIN.name:
             projects_queryset = self.queryset
         elif self.request.user.user_type == CustomUser.UserType.MANAGER.name:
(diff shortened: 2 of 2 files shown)


managers/models.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
rwrzesien: Please remove all from `models.py`.

managers/views.py
```diff
@@ -34,17 +38,20 @@ class ProjectsList(ListView):
     template_name = "managers/projects_list.html"
 
     def get_queryset(self) -> QuerySet:
+        logger.info(f"Get project query set for user: {self.request.user}")
```
rwrzesien: Not needed or debug.

managers/views.py
```diff
@@ -34,17 +38,20 @@ class ProjectsList(ListView):
...
         if self.request.user.user_type == CustomUser.UserType.ADMIN.name:
             projects_queryset = Project.objects.all().order_by("id")
         elif self.request.user.user_type == CustomUser.UserType.MANAGER.name:
             projects_queryset = Project.objects.filter(managers__id=self.request.user.pk)
         else:
             projects_queryset = Project.objects.none()
         if self.request.GET.get("sort"):
+            logger.info("Sort projects")
```
rwrzesien: Not needed.

managers/views.py
```diff
@@ -34,17 +38,20 @@ class ProjectsList(ListView):
...
             if "members" in self.request.GET.get("sort"):
                 projects_queryset = Project.objects.annotate(members_count=Count("members")).order_by(
                     self.request.GET.get("sort")
                 )
+                logger.debug(f"Sorted projects by members: {projects_queryset}")
```
rwrzesien: Not needed.

managers/views.py
```diff
@@ -57,6 +64,7 @@ class ProjectDetail(APIView):
 
     @staticmethod
     def get(_request: HttpRequest, pk: int) -> Response:
+        logger.info(f"User: {_request.user} get in to project detail with id: {pk}")
```
rwrzesien: Add `user.pk`.

managers/views.py
```diff
@@ -68,13 +76,19 @@ class ProjectCreateView(CreateView):
...
+        logger.info(f"User: {self.request.user} is in project create view")
         context_data = super().get_context_data(**kwargs)
         context_data["back_url"] = self.get_success_url()
         return context_data
 
     def get_success_url(self) -> str:  # pylint: disable=no-self-use
         return reverse("custom-projects-list")
 
+    def form_valid(self, form: ProjectForm) -> HttpRequest:
+        project = form.save()
```
rwrzesien: Isn't this being done `super().form_valid(form)` or somewhere else already?

managers/views.py
```diff
@@ -89,17 +103,31 @@ def get_context_data(self, **kwargs: Any) -> dict:
         return context_data
 
     def get_success_url(self) -> str:
+        logger.info(f"User: {self.request.user} is in ProjectUpdate view for project with id: {self.kwargs['pk']}")
         return reverse("custom-project-detail", kwargs={"pk": self.kwargs["pk"]})
 
+    def form_valid(self, form: ProjectForm) -> HttpRequest:
+        project = form.save()
```
rwrzesien: Same here.

managers/views.py
```diff
@@ -89,17 +103,31 @@ def get_context_data(self, **kwargs: Any) -> dict:
...
+            logger.debug(f"User: {request.user} want to delete project")
             return redirect(reverse("home"))
 
         return super().dispatch(request, *args, **kwargs)
 
     def get_success_url(self) -> str:
         return reverse("custom-projects-list")
+
+    def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        project = self.get_object()
```
rwrzesien: User `super()` here too instead of rewriting whole method.

managers/models.py
Szymiks: So how we should logs signals?

managers/views.py
Szymiks: That's true, but how then I could log information about created project and his id ?

managers/views.py
Szymiks: I made it debug, but in final decision I could delete it too :)

managers/views.py
Szymiks: Done

managers/views.py
Szymiks: Done

managers/views.py
Szymiks: Done

Szymiks: done

managers/views.py
Szymiks: Done

managers/views.py
Szymiks: I am not sure if I did it in good way,
at the first moment i just wrote it like this
```python
    def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
        super().delete(request, args, kwargs)
        success_url = self.get_success_url()
        logger.info(f"Project with id: {request.project.pk} has been deleted by user with id: {self.request.user.pk}")
        return HttpResponseRedirect(success_url)
```
But I didn't choose this way because I wasn't sure that it is proper way.

managers/models.py
Szymiks: As we talk I use it only to handlers.

managers/models.py
```diff
@@ -43,6 +46,7 @@ def __str__(self) -> str:
 def update_user_type(sender: Project, action: str, pk_set: Set, **kwargs: Any) -> None:
     assert sender == Project.managers.through
     project = kwargs["instance"]
+    logger.debug(f"{action}")
```
rwrzesien: This says nothing actually, please add `sender.pk` and `pk_set` too and make it more descriptive.

managers/models.py
```diff
@@ -65,8 +71,10 @@ def change_user_type_to_employee(pk_set: Set) -> None:
     for user_id in pk_set:
         user = CustomUser.objects.get(pk=user_id)
         if not user.manager_projects.exists():
+            logger.info(f"User: {user} user type change to employee")
```
rwrzesien: `user.pk`

managers/models.py
```diff
@@ -65,8 +71,10 @@ def change_user_type_to_employee(pk_set: Set) -> None:
...
             user.user_type = CustomUser.UserType.EMPLOYEE.name
             user.full_clean()
             user.save()
         else:
+            logger.debug(f"User: {user} has not changed to employee")
```
rwrzesien: `user.pk`

managers/views.py
```diff
@@ -89,17 +101,31 @@ def get_context_data(self, **kwargs: Any) -> dict:
...
+            logger.debug(f"User with id: {request.user.pk} want to delete project")
             return redirect(reverse("home"))
 
         return super().dispatch(request, *args, **kwargs)
 
     def get_success_url(self) -> str:
         return reverse("custom-projects-list")
+
+    def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        project = super().get_object()
```
rwrzesien: Use `super().delete()` here too instead of rewriting logic.

managers/models.py
Szymiks: Done

managers/models.py
Szymiks: Done

managers/models.py
Szymiks: Done

managers/views.py
Szymiks: done

Issue #159, opened 2019-04-26
If admin is assigned to project as manager, his user_type is changed to manager


maciejSamerdak:
When admin creates a new project and assigns himself as a manager, his admin status is revoked and he turns into a regular manager.

Same thing happens if he's assigned by other user, both on creation and update.

This is definitively related to signals, admin's use case was omitted during implementation.

PR #165, opened 2019-04-30
Bugfix repair handlers for admin in handlers

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/159

Admin user type shouldn't change to `Manager` after assignee him as a project manager.

### DIFF
--- a/managers/tests/test_unit_project_signals.py
+++ b/managers/tests/test_unit_project_signals.py
@@ -17,6 +17,11 @@ class TestProjectSignals(TestCase):
         )
         self.manager.full_clean()
         self.manager.save()
+        self.admin = CustomUser(
+            email="admin@codepoets.it", password="adminpasswd", user_type=CustomUser.UserType.ADMIN.name
+        )
+        self.admin.full_clean()
+        self.admin.save()
         self.project = Project(name="Test Project", start_date=timezone.now())
         self.project.full_clean()
         self.project.save()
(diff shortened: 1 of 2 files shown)


rwrzesien:
Please add test for this case and its good to go.

Szymiks:
> Please add test for this case and its good to go.
I added test but also some changes in other method when Admin is no longer manager of any project, because it wasn't work too.

managers/tests/test_unit_project_signals.py
```diff
@@ -33,3 +38,15 @@ def test_user_should_be_updated_to_employee_when_he_is_no_longer_manager_of_any_
         self.client.post(path=f"/admin/managers/project/{self.project.pk}/change/", args=self.project)
         self.manager.refresh_from_db()
         self.assertEqual(self.manager.user_type, CustomUser.UserType.EMPLOYEE.name)
+
+    def test_admin_should_not_be_updated_to_manager_when_he_is_assigne_as_a_project_manager(self):
+        self.project.managers.add(self.admin)
+        self.client.post(path="/admin/managers/project/add/", args=self.project)
+        self.admin.refresh_from_db()
+        self.assertNotEqual(self.admin.user_type, CustomUser.UserType.MANAGER.name)
```
kbeker: @Szymiks For future information. Always better to check if it is equal to something than not equal :) I'll fix it, because I want to merge it now :)

managers/tests/test_unit_project_signals.py
```diff
@@ -33,3 +38,15 @@ def test_user_should_be_updated_to_employee_when_he_is_no_longer_manager_of_any_
...
+
+    def test_admin_should_not_be_updated_to_employee_when_he_is_no_longer_manager_of_any_project(self):
+        self.project.managers.remove(self.admin)
+        self.client.post(path=f"/admin/managers/project/{self.project.pk}/change/", args=self.project)
+        self.admin.refresh_from_db()
+        self.assertNotEqual(self.admin.user_type, CustomUser.UserType.EMPLOYEE.name)
```
kbeker: Same here

Issue #163, opened 2019-04-29
Add date format as a hint in help_text in creating new project view.


Karrp:
Related #155 because this is a quick workaround that will remain.

As @rwrzesien said [there](https://github.com/Code-Poets/sheetstorm/issues/155#issuecomment-487327343):
> Also, it would be good to add date format as a hint in `help_text`, this is a good task for @Szymiks . If possible please make it dynamic, so the same value is used in `ProjectForm.Meta.widgets` and in `help_text` (it might be defined in constants for example).

PR #166, opened 2019-04-30
Chore add date format to new project

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/163

I'm not sure if I understood this problem, but I give this PR to your assessment.


managers/models.py
```diff
@@ -25,7 +26,7 @@ def filter_completed(self) -> QuerySet:
 
 class Project(models.Model):
     name = models.CharField(max_length=MAX_NAME_LENGTH)
-    start_date = models.DateField()
+    start_date = models.DateField(help_text=MESSAGE_FOR_CORRECT_DATE_FORMAT)
```
rwrzesien: Doesn't it require migration?

managers/commons/constants.py
```diff
@@ -1 +1,3 @@
 MAX_NAME_LENGTH = 64
+CORRECT_DATE_FORMAT = "%Y-%m-%d"
+MESSAGE_FOR_CORRECT_DATE_FORMAT = "Please enter date in this format: YYYY-MM-DD"
```
rwrzesien: The case is that now `MESSAGE_FOR_CORRECT_DATE_FORMAT` will not get updated when `CORRECT_DATE_FORMAT` changes, which was a little bit my point here. But message "Please enter date in this format: %Y-%m-%d" might look weird. So we either do it one way or another. What do you guys think? @Karrp @kbeker

managers/models.py
kbeker: Sure it does. Please add migration and add in front of this commit name `[M]`

managers/commons/constants.py
kbeker: @rwrzesien As far as I understand, this PR should add some hint for user how should he put correct date, right? I see that you are talking here about possible out-dating code. Maybe we should add here simply `assert` to constrain remember about making changes with this constants, or system-check, but about this I think it would be over-engineering?

managers/models.py
Szymiks: Ok, so I will wait for final decision and then I will add migrations, too.

managers/commons/constants.py
Szymiks: In my opinion this format `%Y-%m-%d` looks very weird so I decided to make it in this way. Yes it will not be updated, however we have to add migrations after each update, it could be a small inconvenience, isn't it? I found some solution, here -> https://github.com/tutorcruncher/django-bootstrap3-datetimepicker.
We could write our date format like that `YYYY-MM-DD` and then `BasePickerInput` convert it to this `%Y-%m-%d`
But if we have help_text in models then we still need migrations after each update.

managers/commons/constants.py
```diff
@@ -1 +1,3 @@
 MAX_NAME_LENGTH = 64
+CORRECT_DATE_FORMAT = "YYYY-MM-DD"
+MESSAGE_FOR_CORRECT_DATE_FORMAT = "Please enter date in this format: " + CORRECT_DATE_FORMAT
```
rwrzesien: Use formatted string `f"Please enter date in this format: {CORRECT_DATE_FORMAT}"`.

managers/commons/constants.py
rwrzesien: @kbeker 
Yes, definitely system-check would be too much, but about `assert` how would you do it without over-complicating it? My point is that the `CORRECT_DATE_FORMAT` might change but `MESSAGE_FOR_CORRECT_DATE_FORMAT` forgot to be updated. The whole pull request is about reusing same date format in several places.

@Szymiks 
I like the solution, if it works correctly with bootstrap datepicker then it is OK for me. Just change to formatted string in one place and I think we are good to go.

rwrzesien:
@kbeker I think that solution proposed by @Szymiks is nice, after fixing formatted string it should be good to go if its OK with you. But if you have an idea about that assert without making it too complicated, lets add it. If not, I think its good to go (please remember about migrations!).

PR #167, opened 2019-04-30
Fixes to base and js script

Szymiks:
No issue for this.

The problem was with class name and position of `float menu.js`, because it uses jquery. Before uploaded jquery we was wanted to use float menu.js was running.

### DIFF
--- a/users/static/users/scripts/float_menu.js
+++ b/users/static/users/scripts/float_menu.js
@@ -4,6 +4,6 @@ $(document).ready(function () {
     });
 });
 
-$('.btn').on("click", function () {
-    $('.glyphicon').toggleClass('glyphicon-chevron-left glyphicon-chevron-right');
+$('#sidebarCollapse').on("click", function () {
+    $('.toggle').toggleClass('glyphicon-chevron-right');
 });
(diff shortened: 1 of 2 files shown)


users/templates/base.html
```diff
@@ -10,7 +10,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
-    <link rel="stylesheet" type="text/css" href="{% static 'users/center_style.css' %}"/>
+<!--    <link rel="stylesheet" type="text/css" href="{% static 'users/center_style.css' %}"/>-->
```
rwrzesien: Either uncomment or remove :)

Szymiks: removed:)

users/static/users/scripts/float_menu.js
```diff
@@ -4,6 +4,6 @@ $(document).ready(function () {
     });
 });
 
-$('.btn').on("click", function () {
-    $('.glyphicon').toggleClass('glyphicon-chevron-left glyphicon-chevron-right');
-});
+$('#sidebarCollapse').on("click", function () {
+    $('.toggle').toggleClass('glyphicon-chevron-right');
+});
```
kbeker: @Szymiks Missing blank line :)

PR #168, opened 2019-04-30
Bugfix no project selected on join crash

### LINKED ISSUE #141: If no project is selected on project join popup during report creation, the application crashes.
If there are no more projects that user can join to, the list of projects is empty in the project join popup. If we then proceed to 'join' with no project selected, a crash occurs **MultiValueDictKeyError at /reports/**.

maciejSamerdak:
Resolves #141 

**Done:**

- If there are no projects available for joining, the form with project selection visible in popup is replaced with appropriate message, preventing user from submitting an empty form resulting in app crash

### DIFF
--- a/employees/views.py
+++ b/employees/views.py
@@ -57,17 +57,16 @@ class ReportList(APIView):
     renderer_classes = [renderers.TemplateHTMLRenderer]
     template_name = "employees/report_list.html"
     reports_dict = {}  # type: Dict[str, Any]
-    project_form = ""
     permission_classes = (permissions.IsAuthenticated,)
+    hide_join = False
 
     def get_queryset(self) -> QuerySet:
         return Report.objects.filter(author=self.request.user).order_by("-date", "project__name")
 
-    def _add_project(self, serializer: ReportSerializer, project: Project) -> None:
+    def _add_project(self, project: Project) -> None:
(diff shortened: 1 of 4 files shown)


employees/views.py
```diff
@@ -78,20 +77,24 @@ def _create_serializer(self) -> ReportSerializer:
         reports_serializer.fields["date"].initial = str(datetime.datetime.now().date())
         return reports_serializer
 
+    def _create_project_join_form(self) -> ProjectJoinForm:
+        project_form_queryset = Project.objects.exclude(members__id=self.request.user.id).order_by("name")
+        if not project_form_queryset:
```
rwrzesien: `if not project_form_queryset.exists()`

rwrzesien:
This solution seems OK, what I would just do is separate the message display logic from form creation - form still could be created with empty queryset (instead of returning `None`), but not displayed depending on different variable passed to template. This way we avoid tightly coupling those two elements.

I would also add some validation to the form itself, in case if someone tries to still manually call POST on this view, that checks if queryset is empty and then raise ValidationError. This way we will secure it in both frontend (message) and backend (validation).

maciejSamerdak:
I introduced fixes that change the current implementation to one that's closer to what @rwrzesien suggested.

Issue #171, opened 2019-04-30
As an admin, when I press 'discard' while editing other user's data, I get redirected to my own data


maciejSamerdak:
When I'm on user details editing view as an administrator and I press 'Discard' button while editing user other than me, I get redirected to same view, but the id parameter is switched and I can see my own data.

I'd guess somewhere in the template, logged user's id is passed to POST instead of edited user's id.

PR #179, opened 2019-05-08
Bugfix changed sended pk while editing user in admin update template

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/171


maciejSamerdak:
Just one thing, there's a pattern to the PRs we're applying currently. I'm not sure if there's a defined template already, but under *resolves* you should add *Done* section and list the key changes you applied in this PR beneath.

Issue #92, opened 2019-03-10
Add Django manage command to create initial data instead of using fixtures


rwrzesien:
Add Django manage command to create initial data instead of using fixtures. This script should create initial data using factories. It might have possibility to remove current data (up to discuss).

PR #196, opened 2019-05-13
Chore add django manage command to create initial data instead of using fixtures

rwrzesien:
Resolves https://github.com/Code-Poets/sheetstorm/issues/92


rwrzesien:
@kbeker 
> @rwrzesien Everything is OK, but I wondering if it should not be check if user with that specific name exists? I have now case where I have admin and I got exception, because it still wants to create this user. Can you add check mechanism to check if user/project exists?

Sure, but lets discuss how to do this in best way. So my assumption was that either the database is empty - and you run the script and its fine, or the dabatase is already filled - and you run the script and it fails. Whole function is wrapped in transaction so there is no partially added data. But you are right, you should be able to run this command at any time without exceptions.
So what should happen if user/project exists? Should its creation then be skipped, and existing object user later (for example existing user for adding report)? 

I am also thinking than maybe instead of hardcoded dates we should generate them? Like finished project is always finished yesterday, reports are added in last month etc. - what do you think?

kbeker:
> So what should happen if user/project exists? Should its creation then be skipped, and existing object user later (for example existing user for adding report)?

I think that if for example `Admin` exists then its creation should be skipped, but rest users should be created. In another example if user exists - get instance of him from database and add it to project which should also be check if exists. I think that its similar to what you wrote above :-) @rwrzesien ?

> I am also thinking than maybe instead of hardcoded dates we should generate them? Like finished project is always finished yesterday, reports are added in last month etc. - what do you think?

I totally agree that dates should be generated instead of hardcoded ones. :)

rwrzesien:
@kbeker Updated on discussed things.

Issue #192, opened 2019-05-09
Admin can change report's author


maciejSamerdak:
Despite admin being an admin, I believe author field is the one field he shouldn't be able to change in report. I cannot think of a good reason why this would be necessary, how can a user upload a report that's not his? Why would admin apply someone's report to someone else? If such situation would ever occur, he can always use Django's admin site.

rwrzesien:
@maciejSamerdak 
> Why would admin apply someone's report to someone else?

Reports send to clients :)

PR #201, opened 2019-05-14
Disable possibility to change report's author by Admin user

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/192

### DIFF
--- a/employees/forms.py
+++ b/employees/forms.py
@@ -20,5 +20,5 @@ class ProjectJoinForm(forms.Form):
 class AdminReportForm(forms.ModelForm):
     class Meta:
         model = Report
-        fields = ("date", "description", "task_activities", "author", "project", "work_hours")
+        fields = ("date", "description", "task_activities", "project", "work_hours")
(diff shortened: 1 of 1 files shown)


rwrzesien:
@dybi This form is used in two views, `AdminReportView` and `ProjectReportDetail` so please make sure both of them are meant to only be used by admin (ask someone from the team, I don't think its documented yet). But on the other hand, I think that "lower" privilege user shouldn't be able to change this too? @Karrp @kbeker @maciejSamerdak what do you guys think?

Currently we don't have any checks for that too, so if it will turn out that one of this views is supposed to be used only by admin, maybe you could add this too (I think decorator like `user_is_type_admin` would be best for this) as it would be really valuable improvement which we can reuse in other views.

Another question would be - can actually any user select `author` when creating report, like it is done currently? Should I be able to create a report for someone else? @Karrp @kbeker @maciejSamerdak what do you guys think? It is a separate issue of course but lets at least discuss here.

PR #205, opened 2019-05-15
Change UserList view from Rest APIView to Django DetailView

kbeker:
I have change view to django generic view, but I have one business question: Only Admin should be able get employees list? @Karrp

### DIFF
--- a/users/tests/test_unit_users_apiview.py
+++ b/users/tests/test_unit_users_apiview.py
@@ -9,31 +9,6 @@ from users.common.utils import generate_random_phone_number
 from users.models import CustomUser
 
 
-class UserListTests(TestCase):
-    def setUp(self):
-        self.user = CustomUser(
-            email="testuser@codepoets.it", password="newuserpasswd", first_name="John", last_name="Doe", country="PL"
-        )
-        self.user.full_clean()
-        self.user.save()
-        self.url = reverse("custom-users-list")
-
-    def test_user_list_view_should_display_users_list_on_get(self):
-        request = APIRequestFactory().get(path=self.url)
-        request.user = self.user
-        response = views.UserList.as_view()(request)
-        self.assertEqual(response.status_code, 200)
-        self.assertContains(response, self.user.user_type)
-        users_list = response.data["users_list"]
-        self.assertTrue(self.user in users_list)
-
-    def test_user_list_view_should_not_be_accessible_for_unauthenticated_user(self):
-        request = APIRequestFactory().get(path=self.url)
-        request.user = AnonymousUser()
-        response = views.UserList.as_view()(request)
-        self.assertEqual(response.status_code, 403)
-
-
 class UserUpdateByAdminTests(TestCase):
     def setUp(self):
         self.user = CustomUser(
--- a/users/views.py
+++ b/users/views.py
@@ -16,7 +16,8 @@ from django.shortcuts import get_object_or_404
 from django.shortcuts import redirect
 from django.shortcuts import render
 from django.urls import reverse_lazy
-from rest_framework import permissions
+from django.utils.decorators import method_decorator
+from django.views.generic import ListView
 from rest_framework import renderers
 from rest_framework import viewsets
 from rest_framework.response import Response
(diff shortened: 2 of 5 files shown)


users/tests/test_unit_users_apiview.py
```diff
@@ -12,26 +12,29 @@
 class UserListTests(TestCase):
     def setUp(self):
         self.user = CustomUser(
-            email="testuser@codepoets.it", password="newuserpasswd", first_name="John", last_name="Doe", country="PL"
+            email="testuser@codepoets.it",
```
rwrzesien: I see it was there before, but I think you can now use safely values generated by factory here, so you can remove `email`, `first_name` and `last_name`

rwrzesien: If this tests Django views please move this tests to `test_views.py`, we will then clearly see how much is left to change :)

users/views.py
```diff
@@ -199,23 +204,19 @@ def post(cls, request: HttpRequest, pk: int) -> Union[Response, HttpResponseRedi
...
-        users_queryset = cls.get_queryset()
-        users_serializer = UserListSerializer(context={"request": request})
-        return Response({"serializer": users_serializer, "users_list": users_queryset})
+    model = CustomUser
+    queryset = CustomUser.objects.prefetch_related("projects")
+
+    def get_context_data(
+        self, *, object_list: Optional[List] = None, **kwargs: Any  # pylint: disable=unused-argument
+    ) -> dict:
+        if self.request.user.is_admin:
```
rwrzesien: Nice try ;) But lets add a decorator for it that we can reuse all over the place. This decorator should take a list of user types that are allowed for a given view, this way we have only one decorator for everything!

users/tests/test_unit_users_apiview.py
```diff
@@ -12,26 +12,29 @@
...
+        response = self.client.get(self.url)
         self.assertEqual(response.status_code, 200)
         self.assertContains(response, self.user.user_type)
-        users_list = response.data["users_list"]
-        self.assertTrue(self.user in users_list)
+        self.assertContains(response, self.user.email)
+        self.assertContains(response, self.user.first_name)
+        self.assertContains(response, self.user.last_name)
 
     def test_user_list_view_should_not_be_accessible_for_unauthenticated_user(self):
```
rwrzesien: How about adding a test that non-admin users can't access the view? There is already similar one called `test_delete_project_function_view_should_not_delete_project_on_non_admin_request`.

utils/decorators.py
```diff
@@ -0,0 +1,20 @@
+from functools import wraps
+from typing import Callable
+from typing import Tuple
+
+from django.contrib.auth.views import redirect_to_login
+from django.core.handlers.wsgi import WSGIRequest
+
+
+def check_permissions(permissions: Tuple[str], redirect_path: str = "/accounts/login/") -> Callable:
```
rwrzesien: "/accounts/login/" => Is it possible to use `reverse` here?

utils/decorators.py
```diff
@@ -0,0 +1,20 @@
...
+    def decorator(view_func: Callable) -> Callable:
+        @wraps(view_func)
+        def _wrapped_view(request: WSGIRequest, *args: dict, **kwargs: dict) -> Callable:
+            for permission in permissions:
+                if not getattr(request.user, permission):
```
rwrzesien: I would base it on database field `user_type`, those properties like `is_admin` might change in the future.

utils/decorators.py
kbeker: I tried at first, but It does not working in this case. I had to left it as it is.

utils/decorators.py
kbeker: Changed. Now we checking `user_type`

utils/decorators.py
rwrzesien: Then please try to set it to `None` and then in decorator code `if None then reverse()`. Should work I think.

utils/decorators.py
rwrzesien: Looks good, only thing I would like to ask you to add is an assert that passed `allowed_user_types` are correct.

kbeker: OK, In progress

Karrp:
Yes, only the administrator should be able to obtain a list of employees.

PR #206, opened 2019-05-15
Chore fix security vulnerabilities

### LINKED ISSUE #32: Security vulnerabilities in pyyaml, requests and django versions the project depends on
Github reports three  dependencies with security vulnerabilities. They need to be updated to latest versions.

kbeker:
Resolves #32


cameel:
This looks really messy. There's no context to tell why some things were added or removed which makes this pull request very hard to review and say whether the code is correct or not.

For example, why does this pull request remove `pylint` and `ipython`? Why were they there to begin with? Did the project, when deployed in production, really require them? And if it did not, why was `mypy` not removed as well? Why was `pipdeptree` in the requirements? Or maybe all those removals are just a bug and not intentional?

Please make the commits more granular so that someone looking at the changes does not have to guess whether they are bugs or intentional changes. For example if they were removed intentionally, there should be a separate commit like "Remove development dependencies mistakenly added to requirements.txt".

Issue #161, opened 2019-04-29
Export reports to Excel compatible file.


Karrp:
In the end of May we need to send reports of employees to customer. It's good to have plan B if something in our application would go wrong. We need an export mechanism to files such as CSV or XLSX compatible with Excel.

PR #212, opened 2019-05-17
Feature add export reports to xlsx

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/161

This feature should give us possibility to export reports to excel compatible format. In this case it should be `xlsx` format.

**Done**
 - Reports are covnerted to `xlsx` format
 - Reports for single user should have `Date , Projects, Task activities, hours, description` columns
 - Last row should have functional total hours counter.
 - Reports for project should have `Date, Task activities, Hours, Description' columns and sheet per employee

**Unresolved yet**
 - Tests for this feature
 - [x] I have a problem with markdown, if someone has description with markdowns symbols then we could see it in excel like: `** Important message **, #Header, - something` *
 - View returned all reports without date filtered. I think that it would be better to implement it when those features will be merged : 
https://github.com/Code-Poets/sheetstorm/pull/138
https://github.com/Code-Poets/sheetstorm/pull/139
But I am not sure.

***We have to delete markdown from description**

### DIFF
--- a/employees/common/exports.py
+++ b/employees/common/exports.py
@@ -1,0 +1,137 @@
+from openpyxl import Workbook
+from openpyxl.cell import Cell
+from openpyxl.styles import Alignment
+from openpyxl.styles import Font
+from openpyxl.utils import get_column_letter
+from openpyxl.worksheet.worksheet import Worksheet
+
+from employees.common.constants import ExcelGeneratorSettingsConstants
+from users.models import CustomUser
+
+
+def set_format_styles_for_main_cells(cell: Cell):
+    cell.font = Font(name=ExcelGeneratorSettingsConstants.FONT.value, bold=True)
+    cell.alignment = Alignment(horizontal=ExcelGeneratorSettingsConstants.CENTER_ALINGMENT.value)
+
+
+def set_columns_width(worksheet: Worksheet, col_num: int, width_settings: list):
+    column_letter = get_column_letter(col_num)
+    column_dimensions = worksheet.column_dimensions[column_letter]
+    column_dimensions.width = ExcelGeneratorSettingsConstants.COLUMNS_WIDTH_FOR_SINGLE_USER.value[col_num - 1]
+
+
+def fill_headers(worksheet: Worksheet, headers: list, width_settings: list):
+    for col_num, column_title in enumerate(headers, 1):
+        cell = worksheet.cell(row=ExcelGeneratorSettingsConstants.HEADERS_ROW.value, column=col_num)
+        cell.value = column_title
+        set_format_styles_for_main_cells(cell)
+        set_columns_width(worksheet, col_num, width_settings)
+
+
+def set_and_fill_description_cell(cell: Cell, cell_value: str):
+    wrapped_alignment = Alignment(vertical=ExcelGeneratorSettingsConstants.VERCTICAL_TOP.value, wrap_text=True)
+    cell.alignment = wrapped_alignment
+    cell.value = cell_value
+
+
+def set_and_fill_hours_cell(cell: Cell, cell_value: str):
+    cell.value = ExcelGeneratorSettingsConstants.TIMEVALUE_FORMULA.value.format(cell_value)
+    cell.number_format = ExcelGeneratorSettingsConstants.HOURS_FORMAT.value
+
+
+def fill_current_report_data(storage_data: dict):
+    for col_num, cell_value in enumerate(storage_data["data"], 1):
+        cell = storage_data["worksheet"].cell(row=storage_data["current_row"], column=col_num)
+        if col_num == storage_data["description_column"]:
+            set_and_fill_description_cell(cell, cell_value)
+        elif col_num == storage_data["hours_column"]:
+            set_and_fill_hours_cell(cell, cell_value)
+        else:
+            cell.value = cell_value
+
+
+def summarizing_reports(worksheet: Worksheet, last_row: int, hours_column: int, formula: str):
+    total_cell = worksheet.cell(row=last_row, column=ExcelGeneratorSettingsConstants.TOTAL_COLUMN.value)
+    total_cell.value = ExcelGeneratorSettingsConstants.TOTAL.value
+    set_format_styles_for_main_cells(total_cell)
+    total_hours_cell = worksheet.cell(row=last_row, column=hours_column)
+    total_hours_cell.value = formula.format(last_row - 1)
+    total_hours_cell.number_format = ExcelGeneratorSettingsConstants.TOTAL_HOURS_FORMAT.value
+    set_format_styles_for_main_cells(total_hours_cell)
+
+
+def set_active_worksheet_name(workbook: Workbook, author: CustomUser) -> Worksheet:
+    worksheet = workbook.active
+    worksheet.title = f"{author.first_name} {author.last_name[0]}"
+    return worksheet
+
+
+def generate_xlsx_for_single_user(author: CustomUser) -> Workbook:
+    reports = author.get_reports_created()
+    workbook = Workbook()
+    worksheet = set_active_worksheet_name(workbook, author)
+    fill_headers(
+        worksheet,
+        ExcelGeneratorSettingsConstants.HEADERS_FOR_SINGLE_USER.value,
+        ExcelGeneratorSettingsConstants.COLUMNS_WIDTH_FOR_SINGLE_USER.value,
+    )
+    current_row = ExcelGeneratorSettingsConstants.FIRST_ROW_FOR_DATA.value
+
+    for report in reports:
+        report_data = {
+            "data": [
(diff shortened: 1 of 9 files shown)


employees/templates/employees/project_report_list.html
```diff
@@ -10,6 +10,9 @@
 
 {% block content %}
 <h1>{{ object.name }}{{ UI_text.PAGE_TITLE.value }}</h1>
+<a href="{% url 'export-project-data-xlsx' pk=object.pk %}" class="btn btn-info">
+        Export Reports
+    </a>
```
rwrzesien: Please align.

employees/views.py
```diff
@@ -255,3 +259,29 @@ def form_valid(self, form: AdminReportForm) -> HttpResponseRedirectBase:
...
+
+
+@method_decorator(login_required, name="dispatch")
+class ExportUserReportView(LoginRequiredMixin, DetailView):
+    model = CustomUser
+
+    def render_to_response(self, context: dict, **response_kwargs: Any) -> HttpResponse:
+        author = super().get_object()
+        response = HttpResponse(content_type="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")
+        response["Content-Disposition"] = f'attachment; filename="{author.email}-reports.xlsx"'
```
rwrzesien: If you want to keep everything in constants, why not those two values? Event more that they are reused below.

employees/common/exports.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+    total_hours_cell.number_format = ExcelGeneratorSettingsConstants.TOTAL_HOURS_FORMAT.value
+    set_format_styles_for_main_cells(total_hours_cell)
+
+
+def generate_xlsx_for_single_user(author: CustomUser):
+    reports = author.get_reports_created()
+    workbook = Workbook()
+    worksheet = workbook.active
+    worksheet.title = f"{author.first_name} {author.last_name}"
```
rwrzesien: How about adding datetime of the report generation too?

employees/common/exports.py
```diff
@@ -0,0 +1,140 @@
...
+    summarizing_reports(worksheet, last_row)
+    return workbook
+
+
+def generate_xlsx_for_project(project: Project):
+    authors = project.members.all()
+    workbook = Workbook()
+    for author in authors:
+        worksheet = workbook.active
+        worksheet.title = f"{author.first_name} {author.last_name}"
```
rwrzesien: How about adding datetime of the report generation too?

employees/common/exports.py
```diff
@@ -0,0 +1,140 @@
...
+    last_row = fill_data_in_reports_for_single_user(worksheet, reports)
+    summarizing_reports(worksheet, last_row)
+    return workbook
+
+
+def generate_xlsx_for_project(project: Project):
+    authors = project.members.all()
+    workbook = Workbook()
+    for author in authors:
+        worksheet = workbook.active
```
rwrzesien: Those 6 lines above are quite important and reused in two methods (here and above). How about creating a separate method out of them?

employees/common/exports.py
```diff
@@ -0,0 +1,140 @@
...
+    cell.alignment = Alignment(horizontal=ExcelGeneratorSettingsConstants.CENTER_ALINGMENT.value)
+
+
+def set_columns_width(worksheet, col_num, width_settings):
+    column_letter = get_column_letter(col_num)
+    column_dimensions = worksheet.column_dimensions[column_letter]
+    column_dimensions.width = width_settings[col_num - 1]
+
+
+def fill_headers_for_single_user(worksheet):
```
rwrzesien: There are some methods that are duplicated for user and project:
fill_headers_for_single_user / fill_headers_for_reports_in_project
fill_data_in_reports_for_single_user / fill_data_in_reports_for_project
summarizing_reports / summarizing_reports_for_project
generate_xlsx_for_single_user / generate_xlsx_for_project

Could you please review them for a common parts that can be extracted to separate functions, or maybe the whole part can be a single method with just few params?

employees/common/exports.py
```diff
@@ -0,0 +1,140 @@
...
+        set_columns_width(worksheet, col_num, width_settings)
+
+
+def fill_data_in_reports_for_single_user(worksheet, reports):
+    row_number = ExcelGeneratorSettingsConstants.FIRST_ROW_FOR_DATA.value
+    for report in reports:
+        data_in_row = [
+            report.date,
+            report.project.name,
+            str(report.task_activities),
```
rwrzesien: Why not `report.task_activities.name`?

employees/common/exports.py
```diff
@@ -0,0 +1,140 @@
...
+from openpyxl.styles import Alignment
+from openpyxl.styles import Font
+from openpyxl.utils import get_column_letter
+
+from employees.common.constants import ExcelGeneratorSettingsConstants
+from managers.models import Project
+from users.models import CustomUser
+
+
+def set_format_styles_for_main_cells(cell):
```
rwrzesien: You are also missing typings in this pull request.

employees/common/exports.py
Szymiks: Adding datetime to each sheet? Or just to filename for instance: 
 - Reports single user - `Jan Kowalski_Datetime-reports.xlsx`
 - Project reports - `Sheet storm_datetime-reports.xlsx`

employees/common/exports.py
Szymiks: Adding datetime to each sheet? Or just to filename for instance:

    Reports single user - Jan Kowalski_Datetime-reports.xlsx
    Project reports - Sheet storm_datetime-reports.xlsx
    I think addding datetime to each sheet could be very ugly but this is just my opinion.

employees/templates/employees/project_report_list.html
Szymiks: Done

employees/views.py
Szymiks: Done

employees/common/exports.py
Szymiks: During implementation I was trying to do it but I didn't find better option than I have written. Because our file with single user reports is different than for whole project. Project reports doesn't have for example Project column, because all reports are from project which generated our data. I could make it in one big method but there will be many if syntax and I think it could be more hard to read but I know that now it is not easy to read, too.

employees/common/exports.py
Szymiks: I created separate method to set title of active worksheet, I think I couldn't give there more stuff because generate xlsx for project and single user are different and filters will be different but maybe I made something wrong.

employees/common/exports.py
Szymiks: Done

employees/common/exports.py
Szymiks: I have tried to extract to separate functions. I don't have any better idea

employees/common/exports.py
Szymiks: I think done but I am not sure about typings to `worksheet`, `workbook` `cell`.

employees/views.py
rwrzesien: Where? I can't see it :)

employees/common/exports.py
rwrzesien: I think adding to file name should be enough.

employees/common/exports.py
rwrzesien: I think adding to file name should be enough.

employees/common/exports.py
rwrzesien: Check if mypy check passes without warnings :)

employees/common/exports.py
rwrzesien: Looks very good, please try to make `fill_data_in_reports_for_single_user` and `fill_data_in_reports_for_project` single function by passing `storage_data` as parameter.
Please also make sure you have cleaned unused values from constants enums.

employees/views.py
Szymiks: I am sorry I added it only to one view

employees/common/exports.py
Szymiks: There are no warning but I gave them object type:P

employees/common/exports.py
Szymiks: I have added some solution

employees/common/exports.py
rwrzesien: Nice try :) The correct types should be in `openpyxl` library which you have installed in venv (btw. you can set venv for project in Pycharm, ten ctrl+click on something will take you to the source code), so you can check them there.

employees/common/exports.py
rwrzesien: Looks very good!

employees/common/exports.py
Szymiks: Done:)

PR #214, opened 2019-05-19
Chore change project list to django view

rwrzesien:
Resolves https://github.com/Code-Poets/sheetstorm/issues/210


managers/views.py
```diff
@@ -17,40 +16,45 @@
...
         else:
-            projects_queryset = Project.objects.none()
-        if self.request.GET.get("sort"):
-            if "members" in self.request.GET.get("sort"):
-                projects_queryset = Project.objects.annotate(members_count=Count("members")).order_by(
-                    self.request.GET.get("sort")
-                )
-            else:
-                projects_queryset = projects_queryset.order_by(self.request.GET.get("sort"))
+            assert False
```
kbeker: @rwrzesien Maybe instead of `assert False` we could raise here CustomException like `UnexpectedUserTypeException`?

rwrzesien: As discussed, lets keep assert as this should not happen and we don't want to handle this situation.

kbeker: As w discussed - @cameel school :D

PR #228, opened 2019-05-22
Bugfix wrong projects displayed after report creation fail

### LINKED ISSUE #222: On report creation, if report is invalid, the project field is filled with all projects from database on response
Whenever validation for serializer fails, the project filed on response is filled with all projects from database. So user can now fill report for any project he doesn't belong to.
This is most likely caused by wrongly filtered Project query passed to serializer in that particular case.

Should be done:
 ------------
 - On invalid form, project field should be filled only with projects the user belongs to

maciejSamerdak:
Resolves #222 

All serializer creation is now done via _create_serializer method. Now each time serializer is instantiated, all required filtrations on it's fields' querysets are performed.

Done:
- All instances of serializer creation have been moved to the `_create_serializer` method

### DIFF
--- a/employees/tests/test_unit_report_viewset.py
+++ b/employees/tests/test_unit_report_viewset.py
@@ -327,6 +327,25 @@ class ReportCustomListTests(TestCase):
             serializer = view._create_serializer()
             self.assertEqual(serializer.fields["date"].initial, "2010-01-21")
 
+    def test_custom_report_list_create_serializer_method_should_return_serializer_filled_with_data_that_was_provided(
+        self
+    ):
+        request = APIRequestFactory().post(
+            path=self.url,
+            data={
+                "date": datetime.datetime.now().date(),
+                "description": "Some description",
+                "project": self.project,
+                "work_hours": Decimal("8.00"),
+                "task_activities": TaskActivityType.objects.get(name="Other"),
+            },
+        )
+        request.user = self.user
+        view = ReportList()
+        view.request = request
+        serializer = view._create_serializer(data=request.POST)
+        self.assertTrue(serializer.is_valid())
+
     def test_custom_report_list_view_should_add_user_to_project_selected_in_project_join_form_on_join(self):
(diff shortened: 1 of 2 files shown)


employees/tests/test_unit_report_viewset.py
```diff
@@ -327,6 +327,27 @@ def test_custom_report_list_create_serializer_method_should_return_serializer_wi
             serializer = view._create_serializer()
             self.assertEqual(serializer.fields["date"].initial, "2010-01-21")
 
+    def test_custom_report_list_create_serializer_method_should_return_serializer_filled_with_data_that_was_provided(
+        self
+    ):
+        from django.test import RequestFactory
```
rwrzesien: Please use `self.client` like we do in view tests, it should work here too and when we switch to Django views there will be less to change.

employees/tests/test_unit_report_viewset.py
```diff
@@ -327,6 +327,27 @@ def test_custom_report_list_create_serializer_method_should_return_serializer_wi
...
+                "project": self.project,
+                "work_hours": Decimal("8.00"),
+                "task_activities": TaskActivityType.objects.get(name="Other"),
+            },
+        )
+        request.user = self.user
+        view = ReportList()
+        view.request = request
+        serializer = view._create_serializer(data=request.POST)
+        self.assertTrue(serializer.is_valid())
```
rwrzesien: Please add test that will prove that for invalid POST request the `project` field value will stay the same.

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Previously used solution works just as well, so import has been removed.

employees/tests/test_unit_report_viewset.py
maciejSamerdak: Added this and few more.

maciejSamerdak:
@dybi Rebased on master and added to release notes.

Issue #225, opened 2019-05-22
User's birth date has no limits


maciejSamerdak:
We can currently employ children, embryos and time travelers. It's nice that we don't discriminate, but this might be a little overboard and illegal.

Should be done:
 ------------
 - Birth date should be earlier than 18 years ago
 - Birth date shouldn't be earlier than at least the beginning of 20th century
 - 
 -

dybi:
closing, as merge failed to close automatically

PR #231, opened 2019-05-24
Bugfix user birth day has no limit

dybi:
resolves:
https://github.com/Code-Poets/sheetstorm/issues/225


kbeker:
@dybi Also please remember when you are creating PR, please add milestone, projects, label and assign yourself to this PR. Same thing please do with issue which is related to this PR. :)

employees/tests/test_unit_report_serializer.py
```diff
@@ -22,16 +22,9 @@
...
-        "date": datetime.datetime.now().date(),
-        "description": "Some description",
-        "author": None,
-        "project": None,
-        "work_hours": Decimal("8.00"),
-    }
 
     def setUp(self):
-
+        super().setUp()
```
rwrzesien: :+1:

users/common/strings.py
```diff
@@ -1,46 +1,44 @@
 # pylint: disable=line-too-long
-from django.utils.translation import ugettext_lazy
+from django.utils.translation import ugettext_lazy as _
```
rwrzesien: :+1: :+1: :+1:

users/serializers.py
```diff
@@ -43,6 +45,13 @@ def validate_email(self, email: str) -> str:
                 raise serializers.ValidationError(CustomValidationErrorText.VALIDATION_ERROR_SIGNUP_EMAIL_MESSAGE)
         return email
 
+    def validate_date_of_birth(self, date_of_birth: datetime.date) -> datetime.date:  # pylint: disable=no-self-use
```
rwrzesien: This should be done directly on `CustomUser.date_of_birth` field as `validators=[MaxValidator(), MinValidator]` (https://docs.djangoproject.com/en/2.2/ref/validators/#maxvaluevalidator). This way we control it on lowest possible level in the app and don't have to worry about it anywhere else. If for some reason it won't work there, then `clean()` method of the model. The case is that we want to remove all remaining of `django-rest-framework` so this serializer will be removed soon, and we shouldn't loose this logic.

rwrzesien:
@dybi Looks good, please just move the check to model instead of serializer, as it will be removed soon.

users/serializers.py
dybi: @rwrzesien  Ok, I will move it, but ther's a problem with `MinValueValidator / MaxValueValidator`. Namely, the min and max value are not set in stone, they depend on the current date (for example, if yesterday user was not 18 years old validation should rasie exception, but if today user turned 18, it should pass).

I will create a custom validator and use it in model.

Issue #223, opened 2019-05-22
On report creation, post request is re-sent on page refresh, resulting in report duplication


maciejSamerdak:
If we just created a new report and we hit refresh, the POST will be resent and duplicate will be created each time.

Should be done:
 ------------
 - Report shouldn't be recreated on refresh
 - 
 - 
 -

PR #233, opened 2019-05-24
Fix creating report in `ReportList` view and adapt test

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/223


dybi:
@rwrzesien Ok, I can do it. What would be the preferred way? One issue per view? Or all at once?

rwrzesien:
@dybi One per view I think is better because its easier to review, test and finish ;)

PR #236, opened 2019-05-24
Bugfix 24 hours validation does not work on report creation

### LINKED ISSUE #221: Validation for sum of hours per day being less than 24 does not work on report creation.
This is most likely caused because this validation wasn't implemented for REST-based serializer, which is currently in use in `CustomReportList` and `CustomReportDetail`

Should be done:
 ------------
 - serializer should include the 24 hours validation

maciejSamerdak:
Resolves #221

### DIFF
--- a/employees/serializers.py
+++ b/employees/serializers.py
@@ -50,7 +50,7 @@ class ReportSerializer(serializers.HyperlinkedModelSerializer):
         model = Report
         fields = ("url", "date", "project", "author", "task_activities", "description", "work_hours")
 
-    def to_representation(self, instance: Report) -> Report:
+    def to_representation(self, instance: Report) -> OrderedDict:
         data = super().to_representation(instance)
         if isinstance(self.instance, Report):
             data["work_hours"] = self.instance.work_hours_str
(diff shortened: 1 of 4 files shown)


kbeker:
@maciejSamerdak mea culpa for deleting branch which was based of this PR :P

maciejSamerdak:
Ok, now it's back to it's last state

maciejSamerdak:
@kbeker I shalt not judge thy sins, save it for thy Creator!

employees/serializers.py
```diff
@@ -48,8 +50,29 @@ class Meta:
...
-    def to_representation(self, instance: Report) -> Report:
+    def to_representation(self, instance: Report) -> OrderedDict:
         data = super().to_representation(instance)
         if isinstance(self.instance, Report):
             data["work_hours"] = self.instance.work_hours_str
         return data
+
+    def validate(self, data: OrderedDict) -> OrderedDict:
+        data = super().validate(data)
+        if "date" in data:
```
dybi: Ok, since it seems that the serializer is never created with `partial` flag (https://www.django-rest-framework.org/api-guide/serializers/#partial-updates), it's safe to assume, that `"date"` is always present in `data`. So, please remove the `if`.

maciejSamerdak: Based on that, I additionally changed `data.get("work_hours", 0)` to just `data["work_hours"]`

Issue #224, opened 2019-05-22
Not all projects are presented under *all* label in project list


maciejSamerdak:
The cause is yet unknown. Here are some details regarding the issue:
![Zrzut ekranu 2019-05-22 o 12 57 20](https://user-images.githubusercontent.com/17569113/58178831-33aeff00-7ca7-11e9-876b-8effa43e9a0d.png)
![Zrzut ekranu 2019-05-22 o 12 59 02](https://user-images.githubusercontent.com/17569113/58178837-37428600-7ca7-11e9-9ce5-d089b002904c.png)
![Zrzut ekranu 2019-05-22 o 12 58 47](https://user-images.githubusercontent.com/17569113/58178841-390c4980-7ca7-11e9-8add-8dfc5f196f66.png)

Should be done:
 ------------
 - 
 - 
 - 
 -

PR #237, opened 2019-05-24
Bugfix-show-all-project-list

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/224

I am not sure if this is right solution, but I have checked that our view using this queryset:
`Project.objects.all()` and our for loop is using `object_list` so still the same query is there because view doesn't refresh it. When I've added `object_list.all` in template it is work fine.

### DIFF
--- a/managers/views.py
+++ b/managers/views.py
@@ -45,14 +45,14 @@ class ProjectsListView(ListView):
         "members_count",
         "-members_count",
     ]
-    queryset = Project.objects.all()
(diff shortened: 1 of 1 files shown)


Szymiks:
> The problem is elsewhere. Please replace `queryset = Project.objects.all()` to `model = Project` and `self.queryset` to `super().get_queryset()` in ProjectsListView.

Done

Issue #246, opened 2019-05-27
Change users.views.UserCreate do django view


dybi:
Should be done:
 ------------
 - change view
 - adapt tests
 - 
 -

PR #252, opened 2019-05-28
Change `UserCreate` from `APIView` to Django `CreateView` 

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/246

### DIFF
--- a/users/templates/user_create.html
+++ b/users/templates/user_create.html
@@ -1,26 +1,26 @@
 {% extends 'base.html' %}
 
 {% load i18n %}
-{% load rest_framework %}
+{% load crispy_forms_tags %}
 
 {% block content %}
-<h2>
-  <small>
-    <a href="{% url 'custom-users-list' %}">
-      <span class="glyphicon glyphicon-chevron-left"></span>
-    </a>
-  </small>
-  &nbsp;Create new employee&nbsp;
-</h2>
+    <h2>
+        <small>
+            <a href="{% url 'custom-users-list' %}">
+                <span class="glyphicon glyphicon-chevron-left"></span>
+            </a>
+        </small>
+        {% trans "Create new employee&" %}
+    </h2>
 
-<div class="modal-dialog">
-  <form action="{% url 'custom-user-create' %}" method="post">
-      {% csrf_token %}
-      {% render_form serializer template_pack='rest_framework/vertical' %}
-      <br/>
-     <button type="submit" class="btn btn-primary">
-       <span class="glyphicon glyphicon-plus"></span> Create
-    </button>
-  </form>
-</div>
+    <div class="modal-dialog">
+        <form action="{% url 'custom-user-create' %}" method="post">
+            {% csrf_token %}
+            {{ form|crispy }}
+            <br/>
+            <button type="submit" class="btn btn-primary">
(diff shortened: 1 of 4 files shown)


users/templates/user_create.html
```diff
@@ -1,26 +1,26 @@
...
-  </small>
-  &nbsp;Create new employee&nbsp;
-</h2>
+    <h2>
+        <small>
+            <a href="{% url 'custom-users-list' %}">
+                <span class="glyphicon glyphicon-chevron-left"></span>
+            </a>
+        </small>
+        &nbsp;Create new employee&nbsp;
```
rwrzesien: Please wrap in {% trans "Create new employee&" %}.

users/templates/user_create.html
```diff
@@ -1,26 +1,26 @@
...
-    </button>
-  </form>
-</div>
+    <div class="modal-dialog">
+        <form action="{% url 'custom-user-create' %}" method="post">
+            {% csrf_token %}
+            {{ form|crispy }}
+            <br/>
+            <button type="submit" class="btn btn-primary">
+                <span class="glyphicon glyphicon-plus"></span> Create
```
rwrzesien: Same here.

users/tests/test_unit_users_apiview.py
```diff
@@ -122,7 +122,12 @@ def test_user_create_view_should_display_create_user_form_on_get(self):
 
     def test_user_create_view_should_add_new_user_on_post(self):
         request = APIRequestFactory().post(
-            path=reverse("custom-user-create"), data={"email": "anothertestuser@codepoets.it"}
+            path=reverse("custom-user-create"),
```
rwrzesien: We agreed that we will move refactored tests to `*/tests/test_views.py` to keep track which ones are not refactored yet, also name is fitting better :)

rwrzesien: Also it would be great if you could refactor them to use `self.client`, as `APIRequestFactory` is rather for rest framework views testing. You can find all examples in already refactored tests.

users/views.py
```diff
@@ -93,31 +97,17 @@ class UserSignUpSuccess(TemplateView):
...
-            return Response({"serializer": serializer, "errors": serializer.errors})
-        email = serializer.validated_data.get("email")
-        serializer.save()
-        user = CustomUser.objects.get(email=email)
-        user.set_password("passwduser")
-        user.full_clean()
-        user.save()
-        logger.info(f"New user with id {user.pk} has been created")
-        return redirect("custom-users-list")
+    def from_valid(self, form: CustomUserCreationForm) -> HttpResponseRedirect:
```
rwrzesien: I think this is not needed, as the super class is doing exactly that, but please check.

dybi: It is - normally `form-valid` redirects to `DetailView` of newly created user. If we want to preserve the old behaviour (of redirecting to users list), this little method must stay as is ;)

rwrzesien: But the upper class is doing the same, please check:
https://github.com/django/django/blob/master/django/views/generic/edit.py#L55

And `save()` is done in different class which is also in the inheritance tree:
https://github.com/django/django/blob/master/django/views/generic/edit.py#L123

PR #258, opened 2019-05-30
Bugfix default creation task activity type during migrations

kbeker:
Bug occurred because default task_activity_type was created in migration. Migrations from employee app had to be modified.


kbeker:
> This looks so convoluted it has to work, right?

I hope so :D

Issue #247, opened 2019-05-27
Change users.views.UserUpdate do django view


dybi:
Should be done:
 ------------
 - change view
 - adapt tests
 - 
 -

dybi:
Closing as merge did not close this (for whatever reason)

PR #262, opened 2019-05-30
Chore change `UserUpdate` to be Django class based view

dybi:
resloves: #247


users/forms.py
```diff
@@ -18,6 +20,28 @@ class Meta:
         fields = ("email",)
 
 
+class SimpleUserChangeForm(ModelForm):
+    """
+    A form for updating users by EMPLOYEE users..
```
kbeker: One unnecessary dot at the end ;)

users/templates/user_update.html
```diff
@@ -26,7 +27,8 @@ <h1>Account details</h1>
   </div>
   <form class="form-horizonal" action="{% url 'custom-user-update' pk=user.pk %}" method="POST">
     {% csrf_token %}
-    {% render_form serializer template_pack='rest_framework/vertical' %}
+{#    {% render_form serializer template_pack='rest_framework/vertical' %}#}
```
kbeker: Style of parenthesis

users/forms.py
rwrzesien: About the comment itself, I think it is not possible to define how this form will be used in the future, so I would drop `by EMPLOYEE users` part. As we all know those comments get outdated very quickly ;)

users/forms.py
```diff
@@ -18,6 +20,28 @@ class Meta:
...
+    """
+
+    class Meta:
+        model = CustomUser
+        fields = ("first_name", "last_name", "date_of_birth", "phone_number", "country")
+        widgets = {"country": CountrySelectWidget(), "date_of_birth": DatePickerInput(options={"format": "YYYY-MM-DD"})}
```
rwrzesien: Please use constant for "YYYY-MM-DD" format, I think its already defined somewhere (or in settings?).

users/forms.py
```diff
@@ -18,6 +20,28 @@ class Meta:
...
+
+class AdminUserChangeForm(ModelForm):
+    """
+    A form for updating users by ADMIN users. Includes more fields than `SimpleUserChangeForm`.
+    """
+
+    class Meta:
+        model = CustomUser
+        fields = ("email", "first_name", "last_name", "date_of_birth", "phone_number", "country", "user_type")
+        widgets = {"country": CountrySelectWidget(), "date_of_birth": DatePickerInput(options={"format": "YYYY-MM-DD"})}
```
rwrzesien: Same here about the format.

users/forms.py
```diff
@@ -28,7 +52,7 @@ class CustomUserChangeForm(UserChangeForm):
     class Meta:
         model = CustomUser
         fields = "__all__"
-        widgets = {"country": CountrySelectWidget()}
+        widgets = {"country": CountrySelectWidget(), "date_of_birth": DatePickerInput(options={"format": "YYYY-MM-DD"})}
```
rwrzesien: And here.

users/templates/user_update.html
```diff
@@ -1,7 +1,8 @@
 {% extends 'base.html' %}
 
 {% load i18n %}
-{% load rest_framework %}
+{#{% load rest_framework %}#}
```
rwrzesien: If its not needed just remove it.

users/templates/user_update.html
rwrzesien: Same here.

users/views.py
```diff
@@ -110,47 +113,26 @@ def from_valid(self, form: CustomUserCreationForm) -> HttpResponseRedirect:
...
-
-    @classmethod
-    def return_suitable_serializer_for_post_method(
-        cls, request: HttpRequest, user: CustomUser
-    ) -> Union[UserUpdateByAdminSerializer, UserUpdateSerializer]:
-        if request.user.user_type == CustomUser.UserType.ADMIN.name:
-            return UserUpdateByAdminSerializer(user, data=request.data, context={"request": request})
-        else:
-            return UserUpdateSerializer(user, data=request.data, context={"request": request})
+    def get_context_data(self, **kwargs: dict) -> dict:
```
rwrzesien: There is a method `get_form_class()` where you should put this logic, because it might be also used in other places (or ie. some 3rd party libs would call it).

PR #263, opened 2019-05-31
Bugfix admin report form

kbeker:
Problem occurs when Admin or Manager edit employee report.


rwrzesien:
@kbeker Could you please explain the problem a little more? Because from the pull request it looks like you are implementing handling for duration field in form, which as far as I remember works exactly like that out of the box.

kbeker:
> @kbeker Could you please explain the problem a little more? Because from the pull request it looks like you are implementing handling for duration field in form, which as far as I remember works exactly like that out of the box.

@rwrzesien 
Sure, I should do it in PR description earlier :P. When as a Manager or Admin we get to ProjectReportDetail or AdminReportView of any user without this bugfix we get work_hours in format `HH:MM:SS` and if manager or admin edit this report and send work_hours as a `08:00` it will convert it into `00:08:00` so instead of 8 hours it will be just 8 minutes. This PR fixes it and in this views manager or admin always get work_hours in format `HH:MM` and if he edit report and send work_hours as a `08:00` it will be converted correctly :)

PR #266, opened 2019-06-03
Bugfix sorting exported reports for project and access permissions

kbeker:
The point is that when manager exported project with all reports they were not sorted at all. This PR fixes it.

### DIFF
--- a/employees/tests/test_unit_export_data.py
+++ b/employees/tests/test_unit_export_data.py
@@ -6,8 +6,11 @@ from employees.common.constants import ExcelGeneratorSettingsConstants
 from employees.common.exports import generate_xlsx_for_project
 from employees.common.exports import generate_xlsx_for_single_user
 from employees.factories import ReportFactory
+from employees.models import Report
 from managers.factories import ProjectFactory
 from users.factories import AdminUserFactory
+from users.factories import ManagerUserFactory
+from users.factories import UserFactory
 
 
 class DataSetUpToTests(TestCase):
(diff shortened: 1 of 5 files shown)


employees/tests/test_unit_export_data.py
```diff
@@ -97,3 +100,28 @@ def test_hours_should_be_the_same_in_excel(self):
 
     def test_description_should_be_the_same_in_excel(self):
         self.assertEqual(self.report.description, self.workbook_for_user.active.cell(row=2, column=5).value)
+
+
+class DataSetUpForSortingTests(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.employee = UserFactory()
+        self.manager = UserFactory(user_type=CustomUser.UserType.MANAGER.name)
```
rwrzesien: Please add a subclass of UserFactory called `ManagerUserFactory` that only sets `user_type` to manager, I have added `AdminUserFactory` in my previous pull request and it would be nice to have full set of those ;)

kbeker:
@rwrzesien I added now one more thing to this PR. Please look at it also because it is related with your PR with permissions.

employees/views.py
```diff
@@ -317,8 +317,11 @@ def render_to_response(self, context: dict, **response_kwargs: Any) -> HttpRespo
 
 
 @method_decorator(login_required, name="dispatch")
-@method_decorator(check_permissions(allowed_user_types=[CustomUser.UserType.ADMIN.name]), name="dispatch")
-class ExportReportsInProjectView(DetailView):
+@method_decorator(
```
rwrzesien: Looks good! :)

employees/views.py
```diff
@@ -317,8 +317,11 @@ def render_to_response(self, context: dict, **response_kwargs: Any) -> HttpRespo
...
+    check_permissions(allowed_user_types=[CustomUser.UserType.ADMIN.name, CustomUser.UserType.MANAGER.name]),
+    name="dispatch",
+)
+class ExportReportsInProjectView(DetailView, UserIsManagerOfCurrentReportProjectMixin):
```
rwrzesien: Just reverse classes here.

kbeker: @rwrzesien Exactly what I wanted you to see! When they are reversed they don't work :D And I am wondering why :P

employees/tests/test_unit_export_data.py
```diff
@@ -97,3 +100,28 @@ def test_hours_should_be_the_same_in_excel(self):
 
     def test_description_should_be_the_same_in_excel(self):
         self.assertEqual(self.report.description, self.workbook_for_user.active.cell(row=2, column=5).value)
+
+
+class DataSetUpForSortingTests(TestCase):
```
rwrzesien: I think that as long as this is used only in single subclass, there is no need to keep it in separate class.

employees/views.py
rwrzesien: I have done this by myself, but please remember to update `AccessPermissionsTestCase` if you modify anything regarding access permission.

rwrzesien:
@kbeker Thanks to this problem which you have found I have discovered some problems in access permission tests, for which fix, as discussed, I will add to this pull request, so it should resolve both problems. I have also resolved some other small problems with access permissions which I have found along the way.

employees/tests/test_unit_export_data.py
kbeker: Done

kbeker:
> @kbeker Thanks to this problem which you have found I have discovered some problems in access permission tests, for which fix, as discussed, I will add to this pull request, so it should resolve both problems. I have also resolved some other small problems with access permissions which I have found along the way.

I tested one more time everything and I think all works :). Merging

PR #269, opened 2019-06-03
Feature add playbook that deploy new sheetstorm code to the server

bartoszbetka:


kbeker:
It is already merged too but I forgot to change base branch before merging it into master.

Issue #244, opened 2019-05-27
Change employees.views.ReportList to django view


dybi:
Should be done:
 ------------
 - Change view
 - adapt tests

PR #274, opened 2019-06-04
Chore refactor report list view

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/244

**Should be done**
 - add new report
- display report list as before
- could join to project
- display only user projects
- display all available projects for current user in project join form


employees/forms.py
```diff
@@ -46,6 +50,8 @@ class Meta:
         fields = ("date", "description", "task_activities", "project", "work_hours")
         widgets = {"date": DatePickerInput(format="%Y-%m-%d")}
 
-    def __init__(self, *args: Any, **kwargs: Any):
-        super(ReportForm, self).__init__(*args, **kwargs)
-        self.fields["project"].queryset = kwargs["instance"].author.projects.all()
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        user = kwargs.pop("user", None)
```
rwrzesien: In this form you are passing `user` through `kwargs`, in previous `queryset` as extra parameter. Please unify this. I suggest that if element is required for the form it can be extra parameter, if it is optional it could be in `kwargs`. Also, with current code if you don't pass `user` variable will be set to None and line https://github.com/Code-Poets/sheetstorm/pull/274/files#diff-8e5e768a1bdc111c08d6403f77441ee3R56 will raise exception.

Btw. is this necessary to set both `queryset` and `choices` on choice field? I think it depends on if it is `ChoiceField` or `ModelChoiceField`. Please use only one if possible.

employees/tests/test_views.py
```diff
@@ -140,7 +141,7 @@ def setUp(self):
         self.project = ProjectFactory()
         self.project.members.add(self.user)
         self.client.force_login(self.user)
-        self.report = ReportFactory(author=self.user, project=self.project)
+        self.report = ReportFactory(author=self.user, project=self.project, task_activities=self.task_type)
```
rwrzesien: `task_activities` should be now auto created in `ReportFactory`, please remove it as well as creating it at the beginning of `setUp` if it is not needed for particular reason.

employees/tests/test_views.py
```diff
@@ -217,3 +215,168 @@ def test_delete_report_view_should_delete_report_on_post(self):
         response = self.client.post(path=self.url)
         self.assertEqual(response.status_code, 302)
         self.assertFalse(Report.objects.filter(pk=self.report.pk).exists())
+
+
+class ReportCustomListTests(TestCase):
+    def setUp(self):
+        task_type = TaskActivityType(pk=1, name="Other")
```
rwrzesien: Same here about `task_activity` in `ReportFactory`.

employees/views.py
```diff
@@ -67,108 +63,60 @@ def perform_create(self, serializer: ReportSerializer) -> None:
...
+        context_data = super().get_context_data(**kwargs)
+        context_data["object_list"] = (
+            super().get_queryset().filter(author=self.request.user).order_by("-date", "project__name", "-creation_date")
+        )
+        context_data["daily_hours_sum"] = context_data["object_list"].order_by().get_work_hours_sum_for_all_dates()
+        context_data["monthly_hours_sum"] = context_data["object_list"].order_by().get_work_hours_sum_for_all_authors()
+        context_data["project_form"] = ProjectJoinForm(
+            queryset=Project.objects.exclude(members__id=self.request.user.id).order_by("name")
         )
+        context_data["hide_join"] = not Project.objects.exclude(members__id=self.request.user.id).exists()
```
rwrzesien: This looks a little bit over complicated, it should be enough to do
```
Project.objects.filter(members=self.request.user).exists()
```

employees/views.py
```diff
@@ -67,108 +63,60 @@ def perform_create(self, serializer: ReportSerializer) -> None:
...
-        report = reports_serializer.save(author=self.request.user)
-        logger.info(f"User with id: {self.request.user.pk} created new report with id: {report.pk}")
-        return redirect("custom-report-list")
+    def form_invalid_for_join_project(self, form: ProjectJoinForm) -> HttpResponse:
+        return self.render_to_response(context={"form": form})
+
+    def form_valid(self, form: ReportForm) -> ReportForm:
+        self.object = form.save(commit=False)  # pylint: disable=attribute-defined-outside-init
+        self.object.author = self.request.user
+        return super(ReportCreateView, self).form_valid(form)
```
rwrzesien: `super(ReportCreateView, self)` => `super()`.

employees/forms.py
Szymiks: Okay so I will add it as a additional parameter. But how we can set `user` variable to None? I am taking user from request or report author which is required during creating report.

I was wondering about it, and during adapt test I found that if I set queryset only, then my response doesn't have this information about what is in choicefield so I can't test it but when I runserver it is working fine and it displays all user projects. In other way when I set only choices then I have got this information in reponse, however there is possibility to add report by `POST` when you are not in some project. It looks like it works without  validation or something.

employees/views.py
Szymiks: I am not sure, because we want to hide join button where there is no project to join by current user. This will return true even when user will be in one project and it will hide join button and show message `No project to join`.

employees/tests/test_views.py
Szymiks: Done

employees/tests/test_views.py
Szymiks: Done

employees/views.py
Szymiks: Done

employees/tests/test_views.py
```diff
@@ -217,3 +212,151 @@ def test_delete_report_view_should_delete_report_on_post(self):
...
+        response = self.client.get(path=self.url)
+        queryset = response.context_data["object_list"]
+        self.assertEqual(len(queryset), 3)
+        self.assertFalse(other_user_report in queryset)
+        for report in self.user.projects.all():
+            self.assertContains(response, report)
+
+    def test_custom_report_list_add_project_method_should_register_current_user_as_project_member(self):
+        self.new_project.members.add(self.user)
+        self.assertTrue(self.user in self.new_project.members.all())
```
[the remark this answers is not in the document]
rwrzesien: Yes, that is 100% framework test ;)

employees/tests/test_views.py
```diff
@@ -217,3 +212,151 @@ def test_delete_report_view_should_delete_report_on_post(self):
...
+
+        self.project = ProjectFactory()
+        self.project.members.add(self.user)
+
+        self.report = ReportFactory(author=self.user, project=self.project)
+        self.report.full_clean()
+        self.report.save()
+        self.url = reverse("custom-report-list")
+        self.client.force_login(self.user)
+        self.new_project = ProjectFactory()
```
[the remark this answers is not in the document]
rwrzesien: I agree, `setUp()` should cover most of the repeatable preparation of the tests. If something is specific for 1-2 tests, it should be directly in test method - it is more readable I think.

employees/views.py
rwrzesien: Its OK then, thanks for explanation.

employees/views.py
```diff
@@ -67,108 +63,60 @@ def perform_create(self, serializer: ReportSerializer) -> None:
...
-        project.save()
-
-    def _create_serializer(self, data: QueryDict = None) -> ReportSerializer:
-        logger.info(f"Create serializer for user with id: {self.request.user.pk}")
-        if data is not None:
-            reports_serializer = ReportSerializer(data=data, context={"request": self.request})
-            reports_serializer.is_valid()
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        user_joined_to_project = "join" in request.POST
+        if user_joined_to_project:
```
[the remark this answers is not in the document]
rwrzesien: Also if possible make second part `elif` instead of `else` and check if there is a submit input name from second form in `request.POST`.

employees/views.py
```diff
@@ -299,6 +257,11 @@ def form_valid(self, form: ReportForm) -> HttpResponseRedirectBase:
         self.object.save()
         return super().form_valid(form)
 
+    def get_form(self, form_class: ReportForm = None) -> ReportForm:
+        if form_class is None:
+            form_class = super().get_form_class()
+        return form_class(user=self.object.author, **super().get_form_kwargs())
+
```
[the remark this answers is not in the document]
rwrzesien: Looks unnecessary.

employees/views.py
```diff
@@ -67,108 +63,60 @@ def perform_create(self, serializer: ReportSerializer) -> None:
...
+    def get_form(self, form_class: ReportForm = None) -> ReportForm:
+        if form_class is None:
+            form_class = super().get_form_class()
+        return form_class(user=self.request.user, **super().get_form_kwargs())
+
+    def get_context_data(self, **kwargs: Any) -> dict:
+        context_data = super().get_context_data(**kwargs)
+        context_data["object_list"] = (
+            super().get_queryset().filter(author=self.request.user).order_by("-date", "project__name", "-creation_date")
+        )
```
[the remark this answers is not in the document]
rwrzesien: Nice idea and it could work, so please try it. However I still think that ultimate solution to this is splitting the view into two - one for list, one for create.

employees/tests/test_views.py
```diff
@@ -217,3 +212,151 @@ def test_delete_report_view_should_delete_report_on_post(self):
...
+            },
+        )
+        self.assertTrue(response.status_code, 302)
+        self.assertIn(self.new_project, Project.objects.all())
+        self.assertTrue(self.user not in self.new_project.members.all())
+
+    def test_custom_report_list_view_should_handle_no_project_being_selected_in_project_form_on_post(self):
+        response = self.client.post(path=self.url, data={"join": "join"})
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(response.context_data["form"].errors)
```
rwrzesien: You could check what error is exactly returned here, and if it is the one that you expect.

employees/tests/test_views.py
```diff
@@ -217,3 +212,151 @@ def test_delete_report_view_should_delete_report_on_post(self):
...
+        )
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(Report.objects.all().count(), 2)
+
+    def test_custom_report_list_view_should_not_add_new_report_on_post_if_form_is_invalid(self):
+        response = self.client.post(
+            path=self.url, data={"description": "Some description", "project": self.project.pk, "work_hours": "8:00"}
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(Report.objects.all().count(), 1)
```
rwrzesien: This looks also a little bit confusing - you check that form is invalid and report should not be created, but here there is some report already. I know this is the one created in `setUp`, but someone not familiar has to dive deep to figure out why. But this is also effect of combining Create and List views together, and tests for them too.

This is not a mistake, but looks a little bit odd.

How about separating tests for create report and listing and joining project to separate classes? This way you could have nice, non-conflicting data created in each `setup` of those test classes. But I leave this up to you, if think it is too much work please leave it as it is, we will then refactor if we decide to split those views.

employees/views.py
```diff
@@ -67,108 +63,60 @@ def perform_create(self, serializer: ReportSerializer) -> None:
     ),
     name="dispatch",
 )
-class ReportList(UserIsManagerOfCurrentReportProjectMixin, UserIsAuthorOfCurrentReportMixin, APIView):
-    serializer_class = ReportSerializer
-    renderer_classes = [renderers.TemplateHTMLRenderer]
+class ReportCreateView(CreateView):
```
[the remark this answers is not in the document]
rwrzesien: Please rename to `ReportListCreateJoinProjectView` to describe all tasks of this view.

employees/forms.py
rwrzesien: > But how we can set user variable to None? I am taking user from request or report author which is required during creating report.

You are but you don't know in which circumstances someone else could use this form, and if he would know/remember to instantiate it correctly. You have to assume that other devs could use this form in wrong way, and don't test it ;) By making it required parameter you enforce them to use it correctly.

> during adapt test I found that if I set queryset only, then my response doesn't have this information about what is in choicefield so I can't test it but when I runserver it is working fine and it displays all user projects.

Which test and what was the problem? I have removed `self.fields["project"].choices = [(project.id, project.name) for project in user.projects.all()]` and all tests passed for me.

dybi:
closing, as the implementation will be done under: https://github.com/Code-Poets/sheetstorm/pull/314

PR #276, opened 2019-06-04
Feature group days and add daily hours to xlsx export

### LINKED ISSUE #268: Group days and add daily hours to exported xlsx file.
Alicja reported a need for an additional column with the sum of hours of the day. In addition, she asked to mention single day only once.

before:
Zrzut ekranu 2019-06-3 o 12 00 08 PM

Should be done:
 ------------
 - 
 - 
 - 
 -

kbeker:
Resolves #268

### DIFF
--- a/employees/common/exports.py
+++ b/employees/common/exports.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from openpyxl import Workbook
 from openpyxl.cell import Cell
 from openpyxl.styles import Alignment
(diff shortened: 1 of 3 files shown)


employees/common/exports.py
```diff
@@ -74,6 +84,24 @@ def set_active_worksheet_name(workbook: Workbook, author: CustomUser) -> Workshe
     return worksheet
 
 
+def get_report_date_and_daily_hours(reports_date: List[str], date: str, reports: Report):
```
rwrzesien: Looks difficult, how about adding some comment(s) to this function explaining what is going on? :)

kbeker: Sure, I'm going to add comment to this :)

Issue #248, opened 2019-05-27
Change users.views.UserUpdateByAdmin do django view


dybi:
Should be done:
 ------------
 - change view
 - adapt tests
 - 
 -

PR #281, opened 2019-06-05
Change `UserUpdateByAdmin` to be Django class based view 

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/248

### DIFF
--- a/users/templates/users_detail.html
+++ b/users/templates/users_detail.html
@@ -1,7 +1,7 @@
 {% extends 'base.html' %}
 
 {% load i18n %}
-{% load rest_framework %}
+{% load crispy_forms_tags %}
 {% load static %}
 
 {% block extra_head %}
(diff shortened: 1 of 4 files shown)


users/tests/test_views.py
```diff
@@ -156,3 +156,56 @@ def test_user_update_view_should_not_update_user_on_post_if_form_is_invalid(self
         self.user.refresh_from_db()
         self.assertEqual(response.status_code, 200)
         self.assertEqual(phone_number_before_request, self.user.phone_number)
+
+
+class UserUpdateByAdminTests(TestCase):
+    def setUp(self):
+        self.user = CustomUser(
```
rwrzesien: Please use factory for this.

users/templates/users_detail.html
```diff
@@ -32,8 +32,8 @@ <h1>
 
   </div>
     <form class="form-horizonal" action="{% url 'custom-user-update-by-admin' pk=user_detail.pk %}" method="POST">
-      {% csrf_token %}
-      {% render_form serializer template_pack='rest_framework/vertical' %}
+        {% csrf_token %}
```
rwrzesien: 2-spaces "tab" is ok in HTML, please align.

users/tests/test_views.py
```diff
@@ -156,3 +156,56 @@ def test_user_update_view_should_not_update_user_on_post_if_form_is_invalid(self
...
+        self.user_admin = AdminUserFactory()
+        self.client.force_login(self.user_admin)
+        self.user.refresh_from_db()
+
+    def test_user_update_by_admin_view_should_display_user_details_on_get(self):
+        response = self.client.get(path=reverse("custom-user-update-by-admin", args=(self.user.pk,)))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, self.user.email)
+        self.assertContains(response, self.user.first_name)
+        self.assertContains(response, self.user.last_name)
```
rwrzesien: You can also add check like `self.assertTemplateUsed('template_name.html')`

users/tests/test_views.py
```diff
@@ -156,3 +156,56 @@ def test_user_update_view_should_not_update_user_on_post_if_form_is_invalid(self
...
+        self.user.refresh_from_db()
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(new_phone_number, self.user.phone_number)
+
+    def test_user_update_by_admin_view_should_not_update_user_on_post_if_form_is_invalid(self):
+        phone_number_before_request = self.user.phone_number
+        invalid_phone_number = generate_random_phone_number(constants.PHONE_NUMBER_MIN_LENGTH - 1)
+        response = self.client.post(
+            path=reverse("custom-user-update-by-admin", args=(self.user.pk,)),
+            data={
```
rwrzesien: Looks like this `data` could be extracted and defined in `setUp` ie.

users/templates/users_detail.html
dybi: :+1:

users/tests/test_views.py
dybi: :+1:

users/tests/test_views.py
dybi: `data` is not exactly the same, so it cannot be a field, but I've made a small helper method for it :)

Issue #216, opened 2019-05-20
Unify AdminReportView and ProjectReportDetail


maciejSamerdak:
The two views are almost identical, at least in Python code, and they serve pretty much identical function. In order to eliminate redundancy, both views should inherit from same view and maybe could even share one template.

Here's the general idea:

Should be done:
 ------------
 - Create base ReportDetail view for inheritance
 - Derive AdminReportView and ProjectReportDetail from base ReportDetail view
 - Use single template and single UI text class in base ReportDetail view
 - Pass return link to template for discard button in view's context

PR #284, opened 2019-06-06
Extract common `ReportDetailBase` for `ProjectReportDetail` and `AdminReportView`

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/216


employees/views.py
```diff
@@ -439,28 +450,17 @@ def post(  # pylint: disable=unused-argument
...
-    def get_context_data(self, **kwargs: Any) -> dict:
-        context_data = super().get_context_data(**kwargs)
-        context_data["UI_text"] = ProjectReportDetailStrings
-        return context_data
+class ProjectReportDetail(UserIsManagerOfCurrentReportProjectMixin, ReportDetailBase):
+    template_post_url = "project-report-detail"
+    template_discard_url = "project-report-list"
+    url_pk = "project"
 
     def get_success_url(self) -> str:
```
rwrzesien: This can be also moved to parent class, because only difference is in "project-report-list" vs "author-report-list".

dybi: good catch, didn't think of it ;)

PR #292, opened 2019-06-06
Bugfix generating xlsx for one month

kbeker:
There was a bug with generating xlsx reports.

### DIFF
--- a/employees/templatetags/data_display_filters.py
+++ b/employees/templatetags/data_display_filters.py
@@ -1,6 +1,8 @@
 from datetime import timedelta
+from typing import List
 
 from django import template
+from django.db.models.functions import datetime
 
 from common.convert import timedelta_to_string
 
@@ -10,3 +12,20 @@ register = template.Library()
 @register.filter
 def duration_field_to_string(data: timedelta) -> str:
     return timedelta_to_string(data)
+
+
+@register.filter
+def extract_year_and_month_from_url(url: str) -> List[str]:
+    splitted_url = url.split("/")
+    while "" in splitted_url:
+        splitted_url.remove("")
+
+    if splitted_url[-2].isdigit() and splitted_url[-1].isdigit():
+        year = splitted_url[-2]
+        month = splitted_url[-1]
+    else:
+        now = datetime.datetime.now()
+        year = now.year
+        month = now.month
+
+    return [year, month]
--- a/employees/tests/test_unit_templatetags.py
+++ b/employees/tests/test_unit_templatetags.py
@@ -1,8 +1,12 @@
-import datetime
+from datetime import timedelta
 
+import pytest
+from assertpy import assert_that
+from django.db.models.functions import datetime
 from django.test import TestCase
 
 from employees.templatetags.data_display_filters import duration_field_to_string
+from employees.templatetags.data_display_filters import extract_year_and_month_from_url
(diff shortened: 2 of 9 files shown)


employees/templatetags/data_display_filters.py
```diff
@@ -10,3 +12,22 @@
 @register.filter
 def duration_field_to_string(data: timedelta) -> str:
     return timedelta_to_string(data)
+
+
+@register.filter
+def extract_year_and_month_from_url(url: str) -> List[str]:
+    splitted_url = url.split("/")
+    blank_elements = True
+    while blank_elements:
```
dybi: it could be simplified:
```python
while "" in spiltted_url:
   splitted_url.remove("")
```

employees/tests/test_unit_templatetags.py
```diff
@@ -21,4 +25,30 @@ def test_get_key_value_function_should_return_empty_string_on_incorrect_key(self
...
     def test_that_duration_field_to_string_should_parse_duration_to_string(self):
-        self.assertEqual(duration_field_to_string(datetime.timedelta(hours=8)), "08:00")
+        self.assertEqual(duration_field_to_string(timedelta(hours=8)), "08:00")
+
+
+class TestExtractionTag:
+    @pytest.mark.parametrize(
+        ("url", "date"),
+        [
+            ("/felicita/felicita/felicita/felicita/2019/4", ["2019", "4"]),
```
dybi: :rofl:

employees/templatetags/data_display_filters.py
kbeker: Oh, right, I didn't notice it :)

employees/tests/test_unit_templatetags.py
kbeker: I had good music during implementation :D

employees/templatetags/data_display_filters.py
kbeker: Done, of course :)

employees/templatetags/data_display_filters.py
```diff
@@ -10,3 +12,20 @@
 @register.filter
 def duration_field_to_string(data: timedelta) -> str:
     return timedelta_to_string(data)
+
+
+@register.filter
+def extract_year_and_month_from_url(url: str) -> List[str]:
```
rwrzesien: This looks a little bit fragile in terms of if url will be modified in `urls.py`. We could use regex for this, which, if possible, includes current url. But I think its optional, so if you think its worth @kbeker  I can create separate issue for this.

employees/views.py
```diff
@@ -465,6 +465,14 @@ class ProjectReportDetail(UserIsManagerOfCurrentReportProjectMixin, ReportDetail
 class ExportUserReportView(DetailView):
     model = CustomUser
 
+    def get_queryset(self) -> QuerySet:
+        return CustomUser.objects.prefetch_related(
+            Prefetch(
```
rwrzesien: :+1: for using `Prefetch`.

employees/templatetags/data_display_filters.py
kbeker: @rwrzesien Sure, create issue, but add there that it is now with low priority, but worth to do :)

rwrzesien: Done https://github.com/Code-Poets/sheetstorm/issues/295

Issue #240, opened 2019-05-27
Delete markdown from report description when export to excel


Szymiks:
Should be done:
 ------------
 - We have to delete markdown because during export all markdown characters.

maciejSamerdak:
But... But we can't just delete markdown! What about all those markdown-based chats on the internet?  And all these programmers all over the world that contributed their time to make it happen! And you just want to delete it all like this? You absolute maniac!

PR #303, opened 2019-06-12
Bugfix exporting report description to xls

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/240

### DIFF
--- a/employees/common/exports.py
+++ b/employees/common/exports.py
@@ -1,5 +1,6 @@
 from typing import List
 
+from bs4 import BeautifulSoup
 from openpyxl import Workbook
 from openpyxl.cell import Cell
 from openpyxl.styles import Alignment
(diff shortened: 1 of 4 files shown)


employees/common/exports.py
```diff
@@ -153,7 +154,8 @@ def generate_xlsx_for_single_user(author: CustomUser) -> Workbook:
                 report.project.name,
                 report.task_activities.name,
                 report.work_hours_str,
-                report.description,
+                convert_markdown_html_to_text(report.markdown_description),
+                # report.description,
```
maciejSamerdak: 👋 Over here!

maciejSamerdak:
Wait, sorry, I just realised something: this conversion eradicates the hyperlinks from description. It might be an issue, but I can't tell for sure whether hyperlinks must be carried over to **.xls** representation. But if convenient fix was possible, I guess it would be great to include it.

dybi:
Seems, that text of an url is preserved in exported **.xls** ;) We can creata another ticket for perserving actual links in a report

PR #308, opened 2019-06-13
Feature add signup email confirmation

kbeker:
No issue for that.


kbeker:
Oh, I see that I forgot to generate migration. I will do it tomorrow.

users/views.py
```diff
@@ -213,3 +227,25 @@ class CustomPasswordResetConfirmView(PasswordResetConfirmView):
...
     template_name = "accounts/password_reset_complete.html"
+
+
+class ActivateAccountView(TemplateView):
+    extra_context = {"MESSAGES": AccountConfirmationText}
+    template_name = "account_confirmation/confirmation.html"
+
+    def activate(self) -> bool:
+        user_id = force_text(urlsafe_base64_decode(self.kwargs["encoded_user_id"]))
+        user = get_user_or_none(user_id=user_id)
```
dybi: I would suggest checing if user exist, and `return False` upon `DoesNotExist` exception

users/tests/test_views.py
```diff
@@ -201,3 +207,65 @@ def _get_user_data(self, new_phone_number):
...
+            },
+        )
+
+    def test_signup_view_should_display_signup_form_on_get(self):
+        response = self.client.get(reverse("signup"))
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, SignUp.template_name)
+
+    def test_signup_view_should_add_new_user_on_post(self):
+        response = self._register_user_using_signup_view()
```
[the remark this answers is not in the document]
kbeker: You're right :) Done

users/views.py
kbeker: Also done :)

rwrzesien:
@kbeker Why not using `all-auth` for this?

kbeker:
@rwrzesien To be honest - I didn't know about this. You think that now it is worth to change it into `all-auth`?

PR #314, opened 2019-06-14
Change `ReportList` to `ReportListCreateProjectJoinView` (DCBV)

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/244

### DIFF
--- a/employees/templates/employees/report_list.html
+++ b/employees/templates/employees/report_list.html
@@ -1,7 +1,7 @@
 {% extends 'base.html' %}
 
 {% load i18n %}
-{% load rest_framework %}
+{% load crispy_forms_tags %}
 {% load static %}
 
 {% load data_display_filters %}
@@ -24,7 +24,7 @@
     {% endfor %}
     <form action="{{ path }}" method="POST">
         {% csrf_token %}
-        {% render_form serializer template_pack='rest_framework/vertical' %}
+        {{ form|crispy }}
         <a href="https://guides.github.com/features/mastering-markdown/" target="_blank">Markdown is supported</a><br/>
        <input type="submit" value="{{ UI_text.CREATE_REPORT_BUTTON.value }}">
     </form>
--- a/employees/views.py
+++ b/employees/views.py
@@ -14,7 +14,9 @@ from django.http.response import HttpResponse
 from django.http.response import HttpResponseRedirectBase
 from django.shortcuts import redirect
 from django.shortcuts import reverse
+from django.utils import timezone
 from django.utils.decorators import method_decorator
+from django.views.generic import CreateView
 from django.views.generic import DeleteView
 from django.views.generic import DetailView
 from django.views.generic import UpdateView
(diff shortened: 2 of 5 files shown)


employees/templates/employees/report_list.html
```diff
@@ -1,7 +1,8 @@
 {% extends 'base.html' %}
 
 {% load i18n %}
-{% load rest_framework %}
+{#{% load rest_framework %}#}
```
MartynaAnnaGottschling: What is it for?

employees/templates/employees/report_list.html
```diff
@@ -98,6 +99,7 @@ <h4><font color="red">{{ error }}</font></h4>
 
 {% block extra_script %}
     {{ month_form.media }}
+{#    {{ project_form.media }}#}
```
MartynaAnnaGottschling: Like above. Is it a comment? If so, why don't you just remove it?

employees/views.py
```diff
@@ -154,6 +156,80 @@ def redirect_to_current_month(self) -> HttpResponseRedirectBase:
         return redirect(self._get_recent_month_url(pk))
 
 
+@method_decorator(login_required, name="dispatch")
+@method_decorator(
+    check_permissions(
+        allowed_user_types=[
+            CustomUser.UserType.EMPLOYEE.name,
+            CustomUser.UserType.MANAGER.name,
+            CustomUser.UserType.ADMIN.name,
```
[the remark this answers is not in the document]
rwrzesien: I would suggest keeping it, if a new user type will be added, and this view should not be available for this new type, we don't need to update it.

employees/views.py
```diff
@@ -154,6 +156,80 @@ def redirect_to_current_month(self) -> HttpResponseRedirectBase:
...
+    def get_context_data(self, **kwargs: Any) -> dict:
+        context_data = super().get_context_data(**kwargs)
+        context_data["UI_text"] = ReportListStrings
+        context_data["object_list"] = self.get_queryset()
+        context_data["daily_hours_sum"] = context_data["object_list"].order_by().get_work_hours_sum_for_all_dates()
+        context_data["monthly_hours_sum"] = context_data["object_list"].order_by().get_work_hours_sum_for_all_authors()
+        context_data["project_form"] = ProjectJoinForm(
+            queryset=Project.objects.exclude(members__id=self.request.user.id).order_by("name")
+        )
+        context_data["hide_join"] = not Project.objects.exclude(members__id=self.request.user.id).exists()
```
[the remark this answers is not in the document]
rwrzesien: Yes, it would be because currently there are two database queries, and it can be done with one. However, this would require modifying `ProjectJoinForm` to accept list instead of queryset. Then if you evaluate first one (by making it into a `list()`) and assign it to variable, you can pass variable to form. Then in line 207 you can check if list is not empty. 

But as I have said, it would require modifying how `ProjectJoinForm` is instantiated here and in other places, so if we have some-kind-of standard that we pass queryset to forms, we would break it.

Still nice catch!

employees/templates/employees/report_list.html
```diff
@@ -24,7 +25,7 @@ <h4><font color="red">{{ error }}</font></h4>
     {% endfor %}
     <form action="{{ path }}" method="POST">
         {% csrf_token %}
-        {% render_form serializer template_pack='rest_framework/vertical' %}
+      {{ form|crispy }}
```

employees/views.py
```diff
@@ -154,6 +156,80 @@ def redirect_to_current_month(self) -> HttpResponseRedirectBase:
...
+    template_name = "employees/report_list.html"
+    project_join_form = ProjectJoinForm
+    model = Report
+    form_class = ReportForm
+    url_name = "custom-report-list"
+    object = None
+
+    def get_initial(self) -> dict:
+        initial = super().get_initial()
+        initial.update({"date": timezone.now(), "user": self.request.user})
```
rwrzesien: Why do you need `date` in initial data?

rwrzesien:
The problem with validation found by @maciejSamerdak is caused by overriding `form_valid` and setting `object.author` after `form.is_valid()` was called. I think the best solution would be:
1. Add `author` as field to `ReportForm` and set its widget to `HiddenInput`, so it won't be displayed.
2. Pass `request.user` to `ReportForm` in `get_initial` method as initial `author` field value.
3. Remove custom `form_valid` method on `ReportForm`.

The validation for 24 hours should be then checked when `is_valid` method on form is called and all necessary data (author) should already be set.

employees/views.py
dybi: to prepopulate the form with `today`. It was my impression that was the way the old view has been nahdling the form, so I wanted to keep it.

PR #317, opened 2019-06-17
Add captcha to registration view

kbeker:
No issue for that. It is security feature for registration.

### DIFF
--- a/users/forms.py
+++ b/users/forms.py
@@ -1,4 +1,5 @@
 from bootstrap_datepicker_plus import DatePickerInput
+from captcha.fields import CaptchaField
 from django import forms
 from django.contrib.auth.forms import UserChangeForm
 from django.contrib.auth.forms import UserCreationForm
@@ -67,7 +68,8 @@ class CustomUserChangeForm(UserChangeForm):
 
 class CustomUserSignUpForm(UserCreationForm):
     email = forms.EmailField(max_length=constants.EMAIL_MAX_LENGTH, required=True, label="id_email")
+    captcha = CaptchaField()
 
     class Meta:
         model = CustomUser
-        fields = ("email", "password1", "password2")
+        fields = ("email", "password1", "password2", "captcha")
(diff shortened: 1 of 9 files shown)


users/forms.py
```diff
@@ -67,7 +68,8 @@ class Meta:
 
 class CustomUserSignUpForm(UserCreationForm):
     email = forms.EmailField(max_length=constants.EMAIL_MAX_LENGTH, required=True, label="id_email")
+    captcha = CaptchaField()
 
     class Meta:
         model = CustomUser
-        fields = ("email", "password1", "password2")
+        fields = ("email", "password1", "password2", "captcha")
```
maciejSamerdak: If I'm not mistaken, this should contain all the fields that we want to import from the source model. So why is `captcha` mentioned here? It isn't a model field, is it? 😕

users/static/users/scripts/captcha-refresh.js
```diff
@@ -0,0 +1,9 @@
+$(function () {
+    $("#captcha-refresh").click(function()  {
+        $.getJSON("/captcha/refresh/", {}, function(json) {
+            document.getElementsByClassName('captcha')[0].attributes.src.value = json.image_url;
+            document.getElementById("id_captcha_0").value = json.key;
+        });
+        return false;
+    });
+});
```
maciejSamerdak: Was this the source of that major frustration of yours that you manifested so verbally last Friday? 😆

users/forms.py
rwrzesien: Also non-model fields should be listed here.

users/static/users/scripts/captcha-refresh.js
rwrzesien: @kbeker You are mixing jQuery style with vanilla Javascript style here, works for me but if you want you could unify this :)

kbeker: @maciejSamerdak exactly that was it. It didn't want to cooperate with me last friday :P @rwrzesien when I was doing this it was driving me crazy so when it finally started to work I just left it and did not touch :D But I'll try to unify this :)

kbeker: @rwrzesien oh, now it did work from first shot. Changed completely to `jQuery`.

PR #318, opened 2019-06-17
Chore base e2e testing mechanism

### LINKED ISSUE #23: BaseTests mechanism
Currently we are testing our Application only by unit tests. We have to change it and implement mechanism for end-to-end tests. This mechanism should allow us to join new test steps without additional effort for writing methods like `send/receive/check` again and again.  This blueprint contain few steps ahow to create fully operational tool for testing.

## Basic call-send-receive-check mechanism

Should be created basic class which will contain mechanism for calling testing, receiving messages from application and checking their correctness.

1. Function calling should receive functions to call - in this case it will be a function, which will have test instruction which it have to create. It can look like this:

   - Tuple which contains test function to call:

   ```pyt
   tests = (
       BaseTests.stepInitManager,
       BaseTests.stepInitUser,
       BaseTests.stepCreateProject
       BaseTests.stepAddUserToProject
       BaseTests.cleanAfterTests
   )
   
   
   def call_function(self)
   	for test in tests:
       	if isinstance(test, Callable):
           	call(test)
       	else:
           	raise
   ```

   ^ something like this, but this is just a preview example, it might be changed during the implementation.

2. Test function like `BaseTests.stepInitManager` should always return data which will be send to App:
   - data like login, password etc. It should be whole json which will be send to end-point
   - end-point URL where above data should be send
   -  `send()` and `receive()` functions should be called every test
   - After every test must be called function for checking flow correctness:
     - if status_code is equaled to assumed
     - if returned message is equaled to assumed (message returned from send must be also checked)
     - Rest things which can be checked - if you want to test anything more, just add additional check
3. Mechanism should be created in directory `/sheetstorm/e2e-tests/` - the same place where `run-test-coverage.sh` is. Every new TestCase shluld be created in new file and inherit from the base class.
4. In order to test the correct operation of the above mechanism there should be created a simple test, which will:
   - check if administrator account exists and if login and password which will be stored in mechanism are correct
   - create Manager account by administrator account
   - create Employee account by administrator account
   - delete all account which have been created during tests

maciejSamerdak:
I know it still requires a lot of polish, like unit tests, proper logging or some additional data flow during tests. This is more of a preview of work done so far and a proposition of base structure that I prepared according to @kbeker instructions. Please see whether you'd consider this an appropriate approach and leave any suggestions regarding current implementation, naming conventions and future development. According to @kbeker it's meant to be reusable in future projects, so I aim to make it as intuitive and clean as I can.

Done:
1. Created base structure for E2E testing, consisting of following components:
- Step: It defines all actions that are performed as a single, thematically concise step. The base class contains helper methods to ease the implementation of the script.
- E2ETest: A form of _TestCase_, it contains a list of steps that are meant to be performed as a single test. It supplies them with necessary data and then runs each one of them in given order.
- ClientHandler: a dummy class wrapping up a client used for handling requests, which will vary from framework to framework.
- DatabaseHandler: same as above, but regarding database operations
2. Also introduced basic InitUser step as part of the `base` module.
3. Added handlers for Django framework
4. I separated **app-related** modules and **general-use** modules by moving the latter to `utils/` directory.

Resolves #23

### DIFF
--- a/sheetstorm/e2e-tests/steps.py
+++ b/sheetstorm/e2e-tests/steps.py
@@ -1,0 +1,80 @@
+from django.urls import reverse
+
+from users.models import CustomUser
+from utils.e2e.base.base_components import BaseE2ETest
+from utils.e2e.base.base_components import BaseStep
+from utils.e2e.django_components import DjangoClientHandler
+from utils.e2e.django_components import DjangoDatabaseHandler
+
+
+class E2ETest(BaseE2ETest):
+    client_handler = DjangoClientHandler()
+    database_handler = DjangoDatabaseHandler()
+    attributes = {
+        "user_class": CustomUser,
+        "username_field": "email",
+        "password_field": "password1",
+        "signup_url": "/signup/",
+        "login_url": "/accounts/login/",
+    }
+
+
+class InitManager(BaseStep):
+    required_attributes = [
+        "user_class",
+        "user_attributes",
+        "username_field",
+        "password_field",
+        "signup_url",
+        "login_url",
+        "extra_signup_attributes",
+        "admin_username",
+        "admin_password",
+        "target_field",
+        "target_value",
+    ]
+
+    def get_attributes(self):
+        attributes = super().get_attributes()
+        attributes["target_field"] = "user_type"
+        attributes["target_value"] = "MANAGER"
+        return attributes
+
+    def run_tests(self):
+        print("### Initializing manager ###")
+        print("1. Registering new user:")
+        self._create_object(
+            model_class=self.attributes["user_class"],
+            attributes=self.attributes["user_attributes"],
+            url=self.attributes["signup_url"],
+            non_model_fields=self.attributes["extra_signup_attributes"],
+        )
+        print("2. Logging in as newly created user:")
+        self._sign_in_user(
+            username=self._get_field_from_model_attributes("user_attributes", "username_field"),
+            password=self._get_field_from_model_attributes("extra_signup_attributes", "password_field"),
+            url=self.attributes["login_url"],
+        )
+        print("3. Logging in as an administrator:")
+        self._client = self._client.__class__()
+        self._sign_in_user(
+            username=self.attributes["admin_username"],
+            password=self.attributes["admin_password"],
+            url=self.attributes["login_url"],
+        )
+        user_edit_url = reverse(
+            "custom-user-update-by-admin",
+            kwargs={
+                "pk": self._database_handler.retrieve_object_id(
+                    self.attributes["user_class"], self.attributes["user_attributes"]
+                )
+            },
+        )
+        print("4. Changing user type:")
+        self._edit_object(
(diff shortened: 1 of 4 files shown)


rwrzesien:
> DatabaseHandler: same as above, but regarding database operations

@dybi Correct me if I am wrong but should there be access to database in e2e tests? I think that those tests should be run as separate process, even outside of the server machine. I think that is why https://github.com/Code-Poets/sheetstorm/issues/23 doesn't mention database access. Looks like everything in those tests needs to be set up using requests.

dybi:
One general remark, before I dive into the review -  why not use existing frameworks/solutions? I.e [behave](https://github.com/behave/behave)?

kbeker:
@dybi 
> One general remark, before I dive into the review - why not use existing frameworks/solutions? I.e [behave](https://github.com/behave/behave)?

When I was describing this task I hadn't knowledge about something like [behave](https://github.com/behave/behave) and also I'm kind of a fan of golem-node-tests because they looks transparent for me. It also might be prepared and working "test-framework" customized especially for us and with needed for our needs methods and classes. In other way if there will be implemented tests for django views it might be reusable for any further project where almost always we need to test registration, login etc.

maciejSamerdak:
@rwrzesien 

1. I figured we would need it to clear the database or check for objects being successfully created. I figured the latter would belong to an integration tests, since (in my understanding that is) we check the cooperation between all components of the app and that would include communication between the views and the database. So I thought at least.
Regarding the former, @kbeker told me I shouldn't worry about clean ups and that I should use factories, so that same test could be ran multiple times without the need for cleaning. So I guess I could drop the database handler altogether.

**ALTHOUGH:** To access certain urls, I need object's primary key and I can only obtain it from the database, as far as I know.

2, 6. I tried to use `requests` module, but I couldn't get past the CSRF verification. I researched possible approaches with little luck. The most common and only approach I found for passing CSRF token didn't work for me and I couldn't find any solution. I dropped it and created the handler class, so that client could be easily replaced if necessary without major changes in the rest of the code.

5. I moved all of the base and abstract components to single file as, well, "base components". I guess I could move the BaseStep to a separate file. The current `steps.py` contains only implemented sheetstorm-specific steps. I created the InitUser step as a multi-purpose step, I thought of it like of Django's generic views and forms which provide ready-to-use contextual implementations.
I wanted to clearly separate the multi-purpose and app-specific modules to ease an export to another project.

7. That was just for trial runs with `manage.py test` on isolated database. I will change this shortly.

9. The `E2ETest` (not to be confused with `BaseE2ETest` 😉 ) is where I provide the handlers and constant values that will be used throughout all of the integration tests for Sheetstorm and Sheetstorm alone. This is basically where I can configure the base for all following tests by using inheritance.
So now the `BaseE2ETest` is the base implementation to be used in each project that will be using this solution.

maciejSamerdak:
This solution will most likely be discontinued in favour of pre-existing testing frameworks.

Issue #311, opened 2019-06-14
Enable export reports to CSV file


Karrp:
@Szymiks created this functionality, but has not added it yet.

Should be done:
 ------------
 - export reports to CSV
 - add "export to CSV" button next to export to XLSX

PR #326, opened 2019-06-21
Feature enable export reports to csv file

rwrzesien:
Resolves https://github.com/Code-Poets/sheetstorm/issues/311

### DIFF
--- a/employees/common/exports.py
+++ b/employees/common/exports.py
@@ -1,5 +1,8 @@
 import csv
+import zipfile
 from datetime import timedelta
+from io import BytesIO
+from io import StringIO
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -16,6 +19,7 @@ from openpyxl.styles import Border
 from openpyxl.styles import Font
 from openpyxl.styles import Side
 from openpyxl.utils import get_column_letter
+from openpyxl.workbook import _writer
 
 from common.convert import convert_string_work_hours_field_to_hour_and_minutes
 from employees.common.constants import ExcelGeneratorSettingsConstants as constants
@@ -231,7 +235,7 @@ def convert_markdown_html_to_text(html: str) -> str:
     return "".join(BeautifulSoup(html, features="html.parser").findAll(text=True))
 
 
-def save_work_book_as_csv(writer: csv.DictWriter, work_book: Workbook) -> None:
+def save_work_book_as_csv(writer: _writer, work_book: Workbook) -> None:
     sheet = work_book.active
     is_last_row = False
     total_hours = timezone.timedelta()
@@ -267,3 +271,20 @@ def save_work_book_as_csv(writer: csv.DictWriter, work_book: Workbook) -> None:
                 next_row.append(cell.value)
 
         writer.writerow(next_row)
+
+
+def save_work_book_as_zip_of_csv(work_book: Workbook) -> BytesIO:
+    zip_file_data = BytesIO()
+    zip_file = zipfile.ZipFile(zip_file_data, "w")
+
+    for sheet_index, sheet_name in enumerate(work_book.sheetnames):
+        work_book.active = sheet_index
+        csv_file_data = StringIO()
+        writer = csv.writer(csv_file_data)
+        save_work_book_as_csv(writer, work_book)
+        file_name = f'{sheet_name.replace(" ", "_").replace(".", "").lower()}-reports.csv'
+        zip_file.writestr(file_name, csv_file_data.getvalue())
+
+    zip_file.close()
+
+    return zip_file_data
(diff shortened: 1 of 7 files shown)


rwrzesien:
@dybi Yes, it is intended, as it is not technically possible for CSV file to have more than 1 sheet, so I have assumed that we want to export only single user report. If we want to export multi-sheet project report too, I would need a sample of output file structure. @Karrp do we want that and could you provide that sample?

kbeker:
@rwrzesien I think that @Karrp is now unreachable so I think that structure of this output may be as follows:
```
archive named by project_name
project_name_creation_date.zip
| name_firstLetterOfLastName1.csv
| name_firstLetterOfLastName2.csv
.
.
```
I think that this may looks like this. What do you think guys @rwrzesien @dybi ?

rwrzesien:
@kbeker Nice idea, I like it! Will require some extra work but definitely doable.

rwrzesien:
@dybi @kbeker Updated and ready for review.

employees/common/exports.py
```diff
@@ -216,3 +224,58 @@ def _get_report_date_and_daily_hours(self, current_report: Report, reports_subse
...
+                    total_hours += timezone.timedelta(hours=int(hours), minutes=int(minutes))
+            elif is_last_row and cell.data_type is TYPE_FORMULA:
+                next_row.append(str(total_hours)[:-3])
+            else:
+                next_row.append(cell.value)
+
+        writer.writerow(next_row)
+
+
+def save_work_book_as_zip_of_csv(work_book: Workbook) -> zipfile.ZipFile:
```
kbeker: Don't you return here `BytesIO` object?

employees/templates/employees/project_report_list.html
```diff
@@ -23,7 +23,10 @@ <h1>
 </h1>
 {% with date=path|extract_year_and_month_from_url %}
 <a href="{% url 'export-project-data-xlsx' pk=object.pk year=date.0 month=date.1 %}" class="btn btn-info">
-        Export Reports
+    Export Reports
```
kbeker: Please add here that this button gives `XLSX` export

employees/templates/employees/report_list.html
```diff
@@ -100,6 +100,9 @@ <h4><font color="red">{{ error }}</font></h4>
             <a href="{% url 'export-data-xlsx' pk=user.pk year=date.0 month=date.1 %}" class="btn btn-info">
                 Export {{ date.1 | convert_to_month_name }} {{ date.0 }} Reports
```
kbeker: Please add here that this gives `XLSX` export :)

employees/common/exports.py
```diff
@@ -216,3 +224,58 @@ def _get_report_date_and_daily_hours(self, current_report: Report, reports_subse
...
+            continue
+
+        # Check if is last row by comparing first cell value to `TOTAL`.
+        if isinstance(row[0].value, str) and row[0].value == constants.TOTAL.value:
+            is_last_row = True
+
+        next_row = []
+        for cell_number, cell in enumerate(row, start=1):
+            if isinstance(cell.value, str) and cell.value.lower().startswith("=timevalue"):
+                hours_as_string = cell.value[len('=timevalue("') : -len('")')]
```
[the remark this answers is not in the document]
rwrzesien: This is being set by `black`, so we can either disable it `black` config, or disable it in `pylint` config, or disable it only for this single line, what do you think would be best in this case?

Btw. where is `black` config file in our repo? I couldn't find it.

employees/common/exports.py
```diff
@@ -216,3 +224,58 @@ def _get_report_date_and_daily_hours(self, current_report: Report, reports_subse
 
 def convert_markdown_html_to_text(html: str) -> str:
     return "".join(BeautifulSoup(html, features="html.parser").findAll(text=True))
+
+
+def save_work_book_as_csv(writer: csv.DictWriter, work_book: Workbook) -> None:
+    sheet = work_book.get_active_sheet()
```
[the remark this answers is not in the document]
rwrzesien: :+1 done

employees/common/exports.py
rwrzesien: Right, ZIP was returned in previous version of the code. :+1: done

employees/templates/employees/project_report_list.html
rwrzesien: :+1: done

employees/templates/employees/report_list.html
rwrzesien: :+1: done

employees/common/exports.py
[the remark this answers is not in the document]
kbeker: @rwrzesien Actually we don't have black config. We give parameters explicitly to `black` command. Can we by `black` config disable only one line in whole file?

Issue #29, opened 2019-02-08
Default project for new report


Karrp:
If user is adding new report, alphabetical first project is default.
If second project is using more often it is uncomfortable.

rwrzesien:
@Karrp How this should be resolved then from business logic point of view? Should more often used project be selected by default? If yes, you need to define 'more often'.

Or maybe there should be a project that was used by this user last time?

Karrp:
The list of projects should be sorted according to the last use.
Probably the first will be automatically selected as the default one.

PR #296, opened 2019-06-09
Chore remove django rest framework

rwrzesien:
This is currently in progress, but if someone would like to finish this next week this is good starting point.

Requires https://github.com/Code-Poets/sheetstorm/pull/314 to be merged to master first.

PR #327, opened 2019-06-21
Feature default project for new report

rwrzesien:
Resolves https://github.com/Code-Poets/sheetstorm/issues/29

TODO: Add test that ReportForm.projects field sets initial to project in which last report was created after https://github.com/Code-Poets/sheetstorm/pull/296 is merged.

### DIFF
--- a/employees/tests/test_unit_report_model.py
+++ b/employees/tests/test_unit_report_model.py
@@ -317,3 +317,41 @@ class TestGetProjectsWorkPercentage(TestCase):
 
         self.assertEqual(len(result), 1)
         self.assertEqual(result, {self.project_1: 100.0})
+
+
+class TestGetProjectOrderedByLastReportCreationDate(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.user = UserFactory()
+
+        self.project_1 = ProjectFactory()
+        self.project_2 = ProjectFactory()
+        self.project_3 = ProjectFactory()
+
+        self.project_1_report_1 = ReportFactory(author=self.user, project=self.project_1)
+        self.project_2_report_1 = ReportFactory(author=self.user, project=self.project_2)
+        self.project_3_report_1 = ReportFactory(author=self.user, project=self.project_3)
+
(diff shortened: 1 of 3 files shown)


users/models.py
```diff
@@ -153,6 +157,13 @@ def is_employee(self) -> bool:
     def get_reports_created(self) -> QuerySet:
         return self.report_set.select_related("task_activities").order_by("-date", "project__name")
 
+    def get_project_ordered_by_last_report_creation_date(self) -> QuerySet:
+        return self.projects.annotate(
+            last_report_creation_date=Coalesce(
+                Max("report__creation_date", filter=Q(report__author=self)), Value("1970-01-01 00:00:00")
```
dybi: I am not sure if I understand why we need `Coalesce` and `Max` here... Could you please explain?

employees/tests/test_unit_report_model.py
```diff
@@ -255,3 +256,65 @@ def test_that_added_task_activity_must_be_accepted_as_correct_input(self):
         test_activity_type.full_clean()
         test_activity_type.save()
         self.field_should_accept_input("task_activities", TaskActivityType.objects.get(name="test"))
+
+
+class TestGetProjectOrderedByLastReportCreationDate(TestCase):
+    def setUp(self):
```
dybi: personally I would prefer having projects build with `ProjectFactory`, i.e
```python
project1 = ProjectFactory()
```
and then have reports named like
```python
self.project1_report1 = ReportFactory(author=self.user, project=self.project1)
```
this way the tests would be easier to grasp in my opinion.

rwrzesien: :+1: done, more code, but as you have said, easier to read :)

users/models.py
rwrzesien: If given project has no reports created for current user, the returned value from `Max()` is `None, which then is being put at the beginning of the list (because we use DESC ordering). So to avoid that I am using `Coalesce` to replace `None` with oldest date known to human civilization and get the desired order.

Issue #310, opened 2019-06-14
Better export of reports to XLSX


Karrp:
Suggestions arising during the interview with Adrian.

Should be done:
 ------------
 - [x] employees mentioned only by name and first letter of the surname
 - [x] employee with no report shouldn't be mentioned
 - [x] sort sheets by first name of employees
 - [x] sort reports by date rising (first day of month on the top)

maciejSamerdak:
|_Better export reports to XLSX_

Oh yeah? And what if I don't? 😛

PR #331, opened 2019-06-24
Feature better export of reports

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/310

### DIFF
--- a/employees/common/exports.py
+++ b/employees/common/exports.py
@@ -96,8 +96,8 @@ def summarizing_reports(worksheet: Worksheet, last_row: int, hours_column: int, 
 
 
 def get_employee_name(author: CustomUser) -> str:
-    if author.last_name is not None and author.first_name is not None:
-        return f"{author.first_name} {author.last_name}"
+    if author.last_name and author.first_name:
+        return f"{author.first_name} {author.last_name[0]}."
     else:
         return f"{author.email}"
 
--- a/employees/common/exports.py
+++ b/employees/common/exports.py
@@ -102,10 +102,21 @@ def get_employee_name(author: CustomUser) -> str:
         return f"{author.email}"
 
 
-def set_active_worksheet_name(workbook: Workbook, author: CustomUser) -> Worksheet:
+def set_active_worksheet_name(workbook: Workbook, employee_name: str) -> Worksheet:
+    if workbook.active.title != "Sheet":
+        _set_employees_worksheet_active(employee_name, workbook)
     worksheet = workbook.active
-    worksheet.title = get_employee_name(author)
+    worksheet.title = employee_name
     return worksheet
+
+
+def _set_employees_worksheet_active(employee_name, workbook):
+    try:
+        employee_sheet_index = workbook.worksheets.index(employee_name)
+    except ValueError:
+        workbook.create_sheet()
+        employee_sheet_index = -1
+    workbook.active = workbook.worksheets[employee_sheet_index]
 
 
 def get_report_date_and_daily_hours(reports_date: List[str], date: str, reports: Report):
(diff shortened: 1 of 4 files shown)


employees/common/exports.py
```diff
@@ -8,217 +6,212 @@
...
-    total_hours_cell.value = formula.format(last_row - 1)
-    total_hours_cell.number_format = ExcelGeneratorSettingsConstants.TOTAL_HOURS_FORMAT.value
-    set_format_styles_for_main_cells(total_hours_cell, is_header=False)
+    cell.number_format = constants.HOURS_FORMAT.value
 
 
 def get_employee_name(author: CustomUser) -> str:
-    if author.last_name is not None and author.first_name is not None:
-        return f"{author.first_name} {author.last_name}"
+    if author.last_name and author.first_name:
```
kbeker: I prefer checking if `value is not None` when we have value which normally is not `bool`. But it's only my opinion :) I leave the choice to you :)

employees/common/exports.py
```diff
@@ -8,217 +6,212 @@
...
+        return f"{author.first_name} {author.last_name[0]}."
```
kbeker: Why we have to get first element from `last_name`?

employees/common/exports.py
```diff
@@ -8,217 +6,212 @@
...
+        self._prepare_worksheet(employee_name)
         for report in reports:
-            (reports_date, report_date, daily_hours) = get_report_date_and_daily_hours(
-                reports_date, report.date, reports
+            self._fill_single_report(report, reports)
+
+        self._summarize_user_reports()
+
+    def _fill_single_report(self, report, reports):
+        report_date, daily_hours = self._get_report_date_and_daily_hours(report, reports)
```
kbeker: I personally prefer to add tuple as a return when I have more than one return element, because I had once problem with this kind of case and I learned that this is like a little safety check :D

employees/common/exports.py
```diff
@@ -8,217 +6,212 @@
...
+        )
+        cell.value = constants.EMPLOYEE_NAME.value.format(employee_name)
+
+    def _set_column_width(self, col_num: int, column_name: str):
+        column_letter = get_column_letter(col_num)
+        column_dimensions = self._active_worksheet.column_dimensions[column_letter]
+        column_dimensions.width = self._headers_settings[column_name].width
+
+    def _fill_current_report_data(self, storage_data: dict):
+        for column_name, cell_value in storage_data.items():
```
kbeker: Same situation with tuple here :)

employees/common/exports.py
```diff
@@ -8,217 +6,212 @@
...
+        # other way returns None
+        date = current_report.date
+        if self._last_date == date:
+            daily_hours = None
+            report_date = None
+        else:
+            daily_hours = reports_subset.get_report_work_hours_sum_for_date(date)
+            report_date = date
+        self._last_date = date
+        return report_date, daily_hours
```
kbeker: And here too - tuple as a return statement :)

employees/common/exports.py
Szymiks: Wiem, że w adresie e-mailowym mamy również nazwisko, ale Adrian powiedział, że powinniśmy wyświetlać go w ten sposób ponieważ mamy RODO. :P

Szymiks: I know that we also have a name in the e-mail address, but Adrian said we should display it this way because we have the RODO.

dybi: It's the first letter of the surname

employees/common/exports.py
dybi: as discussed - it will never be `None`. As it is a text field, it will contain either `""` or a user's valid name.

PR #332, opened 2019-06-24
Add possibility to export monthly reports for every user

kbeker:
The idea of this feature came after way of delivering monthly reports has been changed. Every user should have possibility to export his own reports.

### DIFF
--- a/employees/templates/employees/report_list.html
+++ b/employees/templates/employees/report_list.html
@@ -94,6 +94,13 @@
         </table>
     </div>
 </div>
+    {% if object_list %}
+        {% with date=path|extract_year_and_month_from_url %}
+            <a href="{% url 'export-data-xlsx' pk=user.pk year=date.0 month=date.1 %}" class="btn btn-info">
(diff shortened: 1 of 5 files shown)


employees/templates/employees/report_list.html
```diff
@@ -94,6 +94,13 @@ <h4><font color="red">{{ error }}</font></h4>
         </table>
     </div>
 </div>
+    {% if object_list|length != 0 %}
```
rwrzesien: `{% if object_list %}` should be enough.

employees/templatetags/data_display_filters.py
```diff
@@ -29,3 +30,11 @@ def extract_year_and_month_from_url(url: str) -> List[str]:
         month = now.month
 
     return [year, month]
+
+
+@register.filter
+def convert_to_month_name(month_number: str) -> str:
```
dybi: maybe a small unit test here? ;)

kbeker: Sure, added :)

Issue #295, opened 2019-06-07
Change `extract_year_and_month_from_url` template filter to use regexp


rwrzesien:
Original comment:
https://github.com/Code-Poets/sheetstorm/pull/292#discussion_r291612121

Current implementation of `extract_year_and_month_from_url` is a little bit fragile. If url change, it might not work. We could change it to use url definition (please verify if this is possible) and regex for matching needed values.

This is low priority.

Szymiks:
I will try to do this

PR #333, opened 2019-06-25
Chore change `extract_year_and_month_from_url`

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/295


Szymiks:
> Looks good to me, but please add more tests (for example for 5 digits as year, etc)

giphy

Issue #203, opened 2019-05-14
Default task activity for employee


maciejSamerdak:
Should be done:
 ------------
 - Each employee should have a default task activity, assigned according to his role
 - On report creation, the default task activity option should be selected

Karrp:
similar to #29 
I think that would be enough:
List of task activities should be sorted according to the last use.
Probably the first will be automatically selected as the default one.

PR #337, opened 2019-06-25
Feature default task activity for new report

dybi:
Resolves: #203 

should be merged after #327

### DIFF
--- a/employees/forms.py
+++ b/employees/forms.py
@@ -57,9 +57,13 @@ class ReportForm(forms.ModelForm):
 
     def __init__(self, *args: Any, **kwargs: Any):
         super(ReportForm, self).__init__(*args, **kwargs)
-        self.fields["project"].queryset = kwargs["initial"]["author"].get_project_ordered_by_last_report_creation_date()
+        author = kwargs["initial"]["author"]
+        self.fields["project"].queryset = author.get_project_ordered_by_last_report_creation_date()
         if self.fields["project"].queryset.exists():
             self.initial["project"] = self.fields["project"].queryset.first()
+        report_set = author.report_set.order_by("-creation_date")
+        if report_set:
+            self.initial["task_activities"] = report_set.first().task_activities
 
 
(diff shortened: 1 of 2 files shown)


employees/forms.py
```diff
@@ -57,7 +57,11 @@ class Meta:
 
     def __init__(self, *args: Any, **kwargs: Any):
         super(ReportForm, self).__init__(*args, **kwargs)
-        self.fields["project"].queryset = kwargs["initial"]["author"].projects.all()
+        author = kwargs["initial"]["author"]
```
rwrzesien: What do you think about making `author` a required parameter for this `init`? I think it should be always present, otherwise this will fail easily. If you want I can add it in https://github.com/Code-Poets/sheetstorm/pull/327.

employees/tests/test_views.py
```diff
@@ -181,16 +196,22 @@ def test_custom_report_list_view_should_redirect_to_another_month_if_month_switc
...
-            date=current_date + relativedelta(years=-1), author=self.user, project=self.report.project
+            date=current_date + relativedelta(years=-1), author=self.user, project=self.second_report.project
         )
         response = self.client.get(self.url)
         self.assertEqual(response.status_code, 200)
         self.assertNotContains(response, other_report.description)
         self.assertNotContains(response, yet_another_report.description)
 
+    def test_form_task_activity_is_initialized_with_latest_activity(self):
+        response = self.client.get(self.url)
```
rwrzesien: Wouldn't it make tests simpler to add creating of extra report just here, instead of adding it in `setup` and modifying other tests?

employees/forms.py
dybi: ok, please add it and I will rebase

employees/tests/test_views.py
dybi: Ok, good point. One thing needs to be changed in `setUp` though. I will add named `task_activities` for `self.report` just to make sure it is different from one in newly created report.

Issue #250, opened 2019-05-28
Unify validator usage in models


dybi:
Right now there is no structured way of validators' usage. Part of the validation is done by passing `validator`s to models' fields arguments, part of validation is done in `clean` method.

Should be done:
 ------------
 - Single field related validation should be done via `validator` passed as argument. ie.
```python
email = models.EmailField(
    CustomUserModelText.EMAIL_ADDRESS, 
    max_length=constants.EMAIL_MAX_LENGTH, 
    unique=True,
    validators=[custom_validate_email_function]
)
```
 - Validation related to more fields (or related to inter-dependency between fields) should be done in `clean()` method, ie.
```python
    def clean(self) -> None:
        super().clean()

        if (
            hasattr(self, "author")
            and isinstance(self.work_hours, Decimal)
            and self.author.report_set.get_report_work_hours_sum_for_date(self.date, self.pk) + self.work_hours > 24
        ):
            raise ValidationError(
                message=ReportValidationStrings.WORK_HOURS_SUM_FOR_GIVEN_DATE_FOR_SINGLE_AUTHOR_EXCEEDED.value
            )
```

kbeker:
@dybi @rwrzesien I think, that only validatores provided by django should be passed as a parameter `validators` in field description. When we will add there custom validatores it might cause problems in future if we want to modify this validator or even delete it. All custom validation should be done in `clean` method because it won't be visible in any migration and this is simply in modification.

rwrzesien:
> When we will add there custom validatores it might cause problems in future if we want to modify this validator or even delete it. 

Yes, that's the "feature" I forgot about, and its really annoying. After thinking it through I think it can be resolved by modifying migration which uses it, but not everyone would like this approach. How about discussing this on live?

There is also a case - sometimes validation is needed for model, which means it will be always triggered, and sometimes it is specific for current view, so it can be implemented in form only. We should remember about this possibility too.

kbeker:
After discuss we decided to keep validators in model fields. We proved, that deleting validators from old migrations will not affect on database. If any of custom validations will go beyond normal field validations then should be placed in `clean` method inside model class.

PR #342, opened 2019-06-26
Chore unify validator models

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/250

**Done**
- Unify only in users app because other validators were included in way we wanted.


users/validators.py
```diff
@@ -23,3 +23,21 @@ def __call__(self, users_birth_date: Optional[datetime.date]) -> Optional[dateti
             if not self.MINIMAL_ACCETABLE_AGLE <= age <= self.MAXIMAL_ACCEPTABLE_AGE:
                 raise ValidationError(ValidationErrorText.VALIDATION_ERROR_AGE_NOT_ACCEPTED)
         return users_birth_date
+
+
+@deconstructible
+class UserDomainEmailValidation:
```
kbeker: If we have now here validation, do we really still need validation in `CustomUserManager` in `_create_user` method?

users/validators.py
```diff
@@ -23,3 +23,21 @@ def __call__(self, users_birth_date: Optional[datetime.date]) -> Optional[dateti
...
+    def __init__(self, domain_list: list) -> None:
+        self.domain_list = domain_list
```
kbeker: Personally I would prefer to delete from here `__init__` method and get acceptable domain_list from settings. I am not sure that this is possible, but I think it is worth to try.

users/validators.py
dybi: good question. Do we even need `CustomUserManager`?

Szymiks: I was wondering whether to delete it, but we have several tests that use this method, so I have to check what it is for and if we need it.

users/validators.py
rwrzesien: It can also work both ways - if not list is provided, get one from settings.

users/validators.py
rwrzesien: I would keep it, it contains two methods which simplifies creation of user and superuser. I would just remove extra validation from there, and add `user.full_check` before `user.save()`.

Szymiks:
I think I made those changes and add test to email validator,

users/validators.py
Szymiks: Done

Szymiks:
> There are some questions to answer / some research to be done

users/validators.py
```diff
@@ -5,6 +5,7 @@
 from django.core.exceptions import ValidationError
 from django.utils.deconstruct import deconstructible
 
+from users.common.constants import VALID_EMAIL_DOMAIN_LIST
```
dybi: please move it do `settings`

Szymiks: I moved it, however not sure if I import it in good way.

users/common/strings.py
```diff
@@ -3,7 +3,7 @@
 
 from django.utils.translation import ugettext_lazy as _
 
-from users.common import constants
+from sheetstorm.settings import VALID_EMAIL_DOMAIN_LIST
```
rwrzesien: Please read https://github.com/Code-Poets/coding-style/wiki/python-style#django and fix the import.

users/tests/test_unit_customuser_model.py
```diff
@@ -6,6 +6,7 @@
 from django.test import TestCase
 from freezegun import freeze_time
 
+from sheetstorm.settings import VALID_EMAIL_DOMAIN_LIST
```
rwrzesien: Same here about importing settings.

users/validators.py
rwrzesien: What about other validations in `create_user` method? There are two more, the one for `email` could be Django default raised by `full_clean()`, but for password can't because this field is not defined in our `CustomUser` model directly. Maybe we should treat this as special case and allow it either here or in `CustomUser.clean()`? What do you guys think @dybi @kbeker ?

users/common/strings.py
Szymiks: Thanks, @dybi has show me how to do this, but during implementation I forgot how to make this import. Done.

users/tests/test_unit_customuser_model.py
Szymiks: Done

users/validators.py
Szymiks: But I didn't add to email validators django `EmailValidator` so we are not checking if email is None and then test is going to fail without this validation, I could add this validator, too but then, I my opininon, we would have few same validations twice, like `is email contains "@"`.

kbeker: > What about other validations in `create_user` method? There are two more, the one for `email` could be Django default raised by `full_clean()`, but for password can't because this field is not defined in our `CustomUser` model directly. Maybe we should treat this as special case and allow it either here or in `CustomUser.clean()`? What do you guys think @dybi @kbeker ?

@rwrzesien 
I think that this comming up under special case which we were talking about and should be done in `clean` method in `CustomUser` model :)

sheetstorm/settings/base.py
```diff
@@ -268,3 +268,5 @@
 ]
 
 CRISPY_TEMPLATE_PACK = 'bootstrap4'
+
+VALID_EMAIL_DOMAIN_LIST = ["codepoets.it"]
```
kbeker: Please set here `None` value and add comment what this settings is for. Real value of this setting please add to `development.py` and `production.py` settings. I think that this settings should be customizable and in base.py should be `None`.

users/common/strings.py
```diff
@@ -1,9 +1,9 @@
 # pylint: disable=line-too-long
 from enum import Enum
 
+from django.conf import settings
```
kbeker: :+1:

sheetstorm/settings/base.py
kbeker: Also for this setting should be added `system_check` that this setting is correctly set. [Here](https://github.com/golemfactory/concent/blob/master/concent_api/concent_api/system_check.py) you can see how `system_check`s should be created. I think, that we should also add system checks for other settings. I can see that there are two settings which can be checked - `SITE_ID` and `EMAIL_BACKEND` which are used by `production.py` settings. What do you think guys about this @rwrzesien @dybi ?

users/tests/test_unit_validators.py
```diff
@@ -44,3 +47,24 @@ def _assert_date_of_birth_is_valid(self, date_of_birth):
...
+        ]
+    )
+    def test_validator_should_throw_exception_with_invalid_email(self, email, exception_message):
+        with self.assertRaises(ValidationError) as exception:
+            self.user_email_validation(email)
+        self.assertEqual(exception_message, exception.exception.message)
+
+    def test_validator_should_accept_valid_email(self):
+        valid_email = self.user_email_validation("valid_email@codepoets.it")
+        self.assertEqual(valid_email, "valid_email@codepoets.it")
```
kbeker: This is good time for you to learn how `override_settings` decorator works. Please add positive test for different value in `VALID_EMAIL_DOMAIN_LIST` for example that, when this setting will have more than one element in list etc.

sheetstorm/settings/base.py
rwrzesien: Maybe empty list instead of None? This way the want fail without setting it.

rwrzesien: Sounds like a good exercise, lets do it, but maybe in separate issue (but assigned to @Szymiks too, so he can learn this)?

users/validators.py
rwrzesien: @Szymiks Why does the test fail exactly? Maybe it need rewriting.

It should fail on `.full_clean()` because the CustomUser model requires `email` to be not-null (`null=True` is the default). So you don't need to set anything extra for this. So in fact what is here https://github.com/Code-Poets/sheetstorm/pull/342/files#diff-d06152d304f72502a60bd8fabb2f792eR47 looks duplicated (unless we need this custom error message, but personally I think that default Django error message for missing field should be enough).

Szymiks: Yes It was because of message, django return `This field cannot be blank`

users/tests/test_unit_validators.py
Szymiks: Done

sheetstorm/settings/base.py
Szymiks: So after this PR I will make this issue

Szymiks: Done

kbeker: Sure, empty list is better :)

kbeker: Separate issue sounds good :)

sheetstorm/settings/base.py
```diff
@@ -268,3 +268,6 @@
 ]
 
 CRISPY_TEMPLATE_PACK = 'bootstrap4'
+
+# List of valid email domains which are allowed in https://s.codepoets.it
```
kbeker: Maybe not in `https://s.codepoets.it` because you are pointing on domain which should use it. I think it better comment for this might be: 
```
# List of valid email domains which are allowed in SheetStorm
```

users/common/strings.py
```diff
@@ -45,13 +45,13 @@ class CustomUserModelText:
 
 
 class ValidationErrorText:
-    VALIDATION_ERROR_EMAIL_MESSAGE = "The given email must be set"
+    VALIDATION_ERROR_EMAIL_MESSAGE = "Enter your email address, please."
```
kbeker: This sounds for me like a begging for correct email address :D Maybe ```Please enter correct email address``` ?

users/common/strings.py
```diff
@@ -45,13 +45,13 @@ class CustomUserModelText:
...
     VALIDATION_ERROR_EMAIL_EXISTING_MESSAGE = "User with this Email address already exists."
-    VALIDATION_ERROR_PASSWORD_MESSAGE = "The given password must be set"
+    VALIDATION_ERROR_PASSWORD_MESSAGE = "Enter you password, please."
```
kbeker: Same here: ```Please enter your password```

users/migrations/0005_auto_20190629_0830.py
```diff
@@ -0,0 +1,24 @@
...
+            model_name='customuser',
+            name='email',
+            field=models.EmailField(error_messages={'blank': 'Enter your email address, please.'}, max_length=255, unique=True, validators=[users.validators.UserEmailValidation()], verbose_name='email address'),
+        ),
+        migrations.AlterField(
+            model_name='customuser',
+            name='password',
+            field=models.CharField(error_messages={'null': 'Enter you password, please.'}, max_length=128),
+        ),
+    ]
```
kbeker: Please delete this two migrations and generate one with all operations. Or maybe what you doing requires two migrations. If yes please explain why :)

users/tests/test_unit_validators.py
```diff
@@ -44,3 +48,41 @@ def _assert_date_of_birth_is_valid(self, date_of_birth):
...
+            ("valid@fewdomain.it", ["test.pl", "fewdomain.it"]),
+        ]
+    )
+    def test_validator_should_allow_all_email_domains_with_empty_valid_domain_list(self, email, valid_email_domains):
+        try:
+            with override_settings(VALID_EMAIL_DOMAIN_LIST=valid_email_domains):
+                valid_email = self.user_email_validation(email)
+        except ValidationError as exception:
+            self.fail(exception.messages)
+        self.assertEqual(valid_email, email)
```
kbeker: You don't have to add here `try...except` block. If method raise exception test will fail anyway.
You can make it shorter:

```python
        with override_settings(VALID_EMAIL_DOMAIN_LIST=valid_email_domains):
            valid_email = self.user_email_validation(email)
        self.assertEqual(valid_email, email)
```

users/tests/test_unit_validators.py
```diff
@@ -44,3 +48,41 @@ def _assert_date_of_birth_is_valid(self, date_of_birth):
...
+    )
+    def test_validator_should_throw_exception_with_invalid_email(self, email, exception_message):
+        with self.assertRaises(ValidationError) as exception:
+            self.user_email_validation(email)
+        self.assertEqual(exception_message, exception.exception.message)
+
+    @parameterized.expand(
+        [
+            ("valid@codepoets.it", ["codepoets.it"]),
+            ("valid@otherdomain.it", ["otherdomain.it"]),
```
kbeker: This test case check exactly the same thing as one with domain `codepoets.it`. You should check here what happen (until we don't have system checks) what will happened if someone forgot to add any valid domain to this list.

kbeker: Oh, I see same test what I mentioned in comment above is at the bottom of this file. So please just delete this one with `otherdomain.it`

sheetstorm/settings/base.py
Szymiks: Done

users/common/strings.py
Szymiks: Done:)

users/common/strings.py
Szymiks: Done

users/migrations/0005_auto_20190629_0830.py
Szymiks: Done

users/tests/test_unit_validators.py
Szymiks: Done

users/tests/test_unit_validators.py
Szymiks: Yes, I deleted this and one from test below.

Issue #251, opened 2019-05-28
Unify error display on forms


dybi:
Should be done:
 ------------
 - all field related validation (`class="invalid-feedback"`) should be displayed in one way
 - all general validation (i.e. `alert`s) should be displayed in one way
 - 
 - 
 -

PR #344, opened 2019-06-26
Chore unify display errors in form

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/251

All other form are displayed by crispy filter so we have to remember about it during frontend refactor

### DIFF
--- a/users/templates/registration/login.html
+++ b/users/templates/registration/login.html
@@ -20,6 +20,7 @@
     <form method="post" class="login">
         {% csrf_token %}
       <input type="text" id="username" class="fadeIn second" name="username" placeholder="Email" required>
+      {% include 'partials/errors_in_field.html' with errors=form.email.errors %}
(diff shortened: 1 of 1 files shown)


users/templates/registration/login.html
```diff
@@ -30,19 +31,6 @@
         <span><a class="underlineHover" href="{% url 'password_reset' %}">Forgot your password?</a></span>
     </div>
   </div>
-     {% if form.errors %}
```
dybi: It seems that those errors are missing - only `form.non_field_errors` seem to be handled in changed code.

Szymiks:
> Looks good please just align (remove extra space at the beginning).
You mean spaces before partial? Pycharm made those spaces after pressing enter. However, done :).

rwrzesien:
@Szymiks Yea, Pycharm likes to mixup spacing in non Python files, unless you configure it, but its a lot of effort.

Issue #315, opened 2019-06-14
Improve the way of entering work hours when creating/editing report


Karrp:
Should be done:
 ------------
 - 99:99 mask for entering hours
 - h:ss placeholder

PR #345, opened 2019-06-26
Feature add mask for work hours field in reports form

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/315

### DIFF
--- a/employees/forms.py
+++ b/employees/forms.py
@@ -32,13 +32,13 @@ class ProjectJoinForm(forms.Form):
 
 class DurationInput(TextInput):
     def format_value(self, value: Optional[str]) -> str:
-        if value is not None:
+        if value is not None and value != ":":
             return timedelta_to_string(parse_duration(value))
         return ""
 
 
 class DurationFieldForm(forms.DurationField):
-    widget = DurationInput
+    widget = DurationInput(attrs={"data-mask": "09:99", "placeholder": "H:MM"})
 
     def clean(self, value: str) -> str:
         if value is None:
--- a/employees/templates/employees/report_list.html
+++ b/employees/templates/employees/report_list.html
@@ -106,7 +106,7 @@
 
 {% block extra_script %}
     {{ month_form.media }}
-    <script src="//code.jquery.com/jquery-1.12.4.js"></script>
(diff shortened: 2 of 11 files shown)


employees/forms.py
```diff
@@ -32,13 +32,13 @@ def __init__(self, queryset: QuerySet, *args: Any, **kwargs: Any) -> None:
...
     def format_value(self, value: Optional[str]) -> str:
-        if value is not None:
+        if value is not None and value != ":":
             return timedelta_to_string(parse_duration(value))
         return ""
 
 
 class DurationFieldForm(forms.DurationField):
-    widget = DurationInput
+    widget = DurationInput(attrs={"data-mask": "99:99", "placeholder": "H:MM"})
```
kbeker: What does it mean that this field has `99:99` mask?

dybi: https://dobsondev.com/2017/04/14/using-jquery-mask-to-mask-form-input/
> * ‘0’: {pattern: /\d/}
> * ‘9’: {pattern: /\d/, optional: true}

employees/templates/employees/report_list.html
```diff
@@ -105,6 +105,9 @@ <h4><font color="red">{{ error }}</font></h4>
 {% endblock %}
 
 {% block extra_script %}
+    {{ form.media }}
+    <script src="//code.jquery.com/jquery-3.3.1.min.js"></script>
+    <script src="//cdnjs.cloudflare.com/ajax/libs/jquery.mask/1.14.10/jquery.mask.js"></script>
```
rwrzesien: Looks a little bit suspicious that we use two versions of the same library, maybe removing older version below will keep everything still working?

employees/forms.py
rwrzesien: So shouldn't it be `90:00`? Which means that you should have at least one digit before `:` and two after. This way you can type in values like `12:00`, `8:00` and `0:30`.  Or maybe I have misunderstood this completely.

dybi: quite frankly, it must be `09:99` (checked empirically;)).

employees/templates/employees/report_list.html
dybi: good point. changed.

Issue #190, opened 2019-05-09
[UX] day of the week in the timestamp


michalpiasecki1:
on the /reports/ page there is a table with columns:

```
Date | Project | Task Activity | Work hours | Description
```

Date field could include day of the week (Mon/Tue/...) for easier navigation. Weeks could be separated as well.

PR #346, opened 2019-06-26
Feature add day name to reports

kbeker:
Resolves https://github.com/Code-Poets/sheetstorm/issues/190


kbeker:
I know that i create wrong branch name, but I will fix it during merging.

Issue #348, opened 2019-06-26
"-------------" in projects drop down list in report list


dybi:
Should be done:
 ------------
 - remove it
 - 
 - 
 -

PR #349, opened 2019-06-26
Bugfix "---------------" in projects dropdown list in reposr list view

dybi:
resolves: #348


employees/models.py
```diff
@@ -59,7 +59,7 @@ class Report(models.Model):
     creation_date = models.DateTimeField(auto_now_add=True)
     last_update = models.DateTimeField(auto_now=True)
     author = models.ForeignKey(CustomUser, on_delete=models.PROTECT)
-    project = models.ForeignKey(Project, on_delete=models.CASCADE)
+    project = models.ForeignKey(Project, on_delete=models.CASCADE, blank=False, default=1)
```
rwrzesien: Is `default` really needed? I think it might cause problems because you are pointing to ID here that might not necessary exist, ie. on fresh installation or when object with ID 1 was removed. `blank=False`.

Issue #350, opened 2019-06-27
When admin edit someone report become owner of this report


kbeker:
Should be done:
 ------------
 - Add test for this scenario
 - Fix bug

Issue #351, opened 2019-06-27
When admin edit someone report need to be member in this project to change it


kbeker:
Admin does not has to be in any project to edit any report.

kbeker:
It is also merged.

PR #352, opened 2019-06-27
Bugfix when admin edit someone report become owner of this report

kbeker:
Resolves https://github.com/Code-Poets/sheetstorm/issues/350 https://github.com/Code-Poets/sheetstorm/issues/351

### DIFF
--- a/employees/views.py
+++ b/employees/views.py
@@ -214,7 +214,7 @@ class ReportDetailBase(UpdateView):
 
     def get_initial(self) -> dict:
         initial = super().get_initial()
-        initial.update({"author": self.request.user})
+        initial.update({"author": self.object.author})
         return initial
 
     def get_context_data(self, **kwargs: Any) -> dict:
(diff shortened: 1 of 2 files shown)


employees/views.py
```diff
@@ -214,7 +214,7 @@ class ReportDetailBase(UpdateView):
 
     def get_initial(self) -> dict:
         initial = super().get_initial()
-        initial.update({"author": self.request.user})
+        initial.update({"author": Report.objects.get(pk=self.kwargs["pk"]).author})
```
rwrzesien: Report should be already available as `self.object`. If it is not, please get it with `self.get_object()`. This method takes into consideration all updates made ie. in `self.queryset` or `get_queryset` method.

employees/views.py
```diff
@@ -262,7 +262,7 @@ class ReportDetailView(
...
```
rwrzesien: Same here, please use `self.object` or `self.get_object()` (preferably first).

employees/views.py
kbeker: Done

employees/views.py
kbeker: Also done

PR #354, opened 2019-06-28
Bugfix MultipleObjectsReturned exception when manager wants to edit his reports

kbeker:


employees/tests/test_views.py
```diff
@@ -338,6 +338,17 @@ def test_manager_should_be_able_to_update_his_reports_in_project_in_which_he_is_
         self.assertEqual(response.status_code, 302)
         self.assertEqual(report_manager.description, data["description"])
 
+    def test_custom_report_detail_view_should_return_only_one_report_when_are_more_than_one_managers_in_project(self):
```
rwrzesien: I think it is more about "when manager is both in project and is author of current report".

PR #375, opened 2019-07-01
Variable default date value on report creation

### LINKED ISSUE #366: Default current day value when displaying other months in report list
When we're displaying other months, the default date value in report creation form is still the current day. Previously it was the first day of the month.

Should be done:
 ------------
- If the month is current, then the value is current day.
- If there are no reports on this month, the value is the first day of the month.
- If there are reports on this month, the value is the next day from the most recent report.
- If the most recent report is from the last day of the month, the value is also the last day of the month.

maciejSamerdak:
Resolves #366 

Added method for deriving default date value in report creation form. This is how it works:
1. If the month is current, then the value is current day.
2. If there are no reports on this month, the value is the first day of the month.
3. If there are reports on this month, the value is the next day from the most recent report.
4. If the most recent report is from the last day of the month, the value is also the last day of the month.

This can be useful if the user is late on filling reports. The date on previous months defaults to the day after the last report (if possible), so after each post the default date value is raised by one day.

Done:
- Added method for calculating appropriate default date for selected month, based on existing reports.

### DIFF
--- a/employees/views.py
+++ b/employees/views.py
@@ -155,9 +155,23 @@ class ReportListCreateProjectJoinView(MonthNavigationMixin, ProjectsWorkPercenta
     url_name = "custom-report-list"
     object = None
 
+    def _get_default_date(self) -> datetime.date:
+        year = int(self.kwargs["year"])
+        month = int(self.kwargs["month"])
+        queryset = self.get_queryset()
+        if not queryset.exists():
+            return datetime.date(year=year, month=month, day=1)
+        last_report = queryset.first()
+        if last_report.creation_date.date() == timezone.now().date():
+            return last_report.date
+        date = last_report.date + timezone.timedelta(days=1)
+        if date.month != month:
+            return last_report.date
+        return date
+
     def get_initial(self) -> dict:
         initial = super().get_initial()
-        initial.update({"date": timezone.now(), "author": self.request.user})
+        initial.update({"date": self._get_default_date(), "author": self.request.user})
         return initial
 
     def get_queryset(self) -> QuerySet:
--- a/employees/tests/test_views.py
+++ b/employees/tests/test_views.py
@@ -4,6 +4,7 @@ from dateutil.relativedelta import relativedelta
 from django.shortcuts import reverse
 from django.test import TestCase
 from django.utils import timezone
+from freezegun import freeze_time
 
 from employees.common.strings import AuthorReportListStrings
 from employees.common.strings import ProjectReportListStrings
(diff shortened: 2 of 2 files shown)


maciejSamerdak:
Now that I think of this, most people might be filling multiple reports on same day. That means the **one day after the last report** default might be counter-intuitive. What do you think of this?

rwrzesien:
@maciejSamerdak This plan looks good, only one problem
```
1. If the month is current, then the value is current day.
vs.
3. If there are reports on this month, the value is the next day from the most recent report.
```

Choose one :) Also, you should move this description to original issue.

employees/views.py
```diff
@@ -155,9 +155,23 @@ class ReportListCreateProjectJoinView(MonthNavigationMixin, ProjectsWorkPercenta
...
+        month = int(self.kwargs["month"])
+        current_date = timezone.now()
+        if current_date.year == year and current_date.month == month:
+            return current_date.date()
+        queryset = self.get_queryset()
+        if not queryset.exists():
+            return datetime.date(year=year, month=month, day=1)
+        date = queryset.first().date + timezone.timedelta(days=1)
+        if date.month != month:
+            return queryset.first().date
```
rwrzesien: You should evaluate queryset only once.

maciejSamerdak:
@rwrzesien So for all months it will behave the same? As in: if there are no reports in current month then the date is the first day?

It might work most of the time, except for few instances. But same can be said about the current date. I think @Karrp should give some insight on which option he would find preferable.

employees/views.py
maciejSamerdak: I forgot this would be an issue. At this point I treated it more like list-based structure returning reference to it's head.

Is it enough to reduce `.first()` calls to one or should I maybe call `.first()` on the `get_queryset()` and call `.exists()` on that?

rwrzesien:
> @rwrzesien So for all months it will behave the same? As in: if there are no reports in current month then the date is the first day?

Yes, this is what I have meant. This is just my opinion based on my experience in adding reports via Sheetstorm. I don't think I have added the report for previous month so far, and it would be great if I have the date set to next day from last report in current month. But I understand it won't be best solution for everyone.

maciejSamerdak:
> > @rwrzesien So for all months it will behave the same? As in: if there are no reports in current month then the date is the first day?
> 
> Yes, this is what I have meant. This is just my opinion based on my experience in adding reports via Sheetstorm. I don't think I have added the report for previous month so far, and it would be great if I have the date set to next day from last report in current month. But I understand it won't be best solution for everyone.

How about taking creation date under consideration? If the last report from the month wasn't filled today, the default date will be the next day. But if the last report was filled today, then default date is today.

So if we display any list and we haven't edited it today, we get the next day (for example: today, if it's the current month and we filled some reports yesterday). Then once we add new report, the date remains the same, allowing us to conveniently fill out other reports from this single day, without constantly changing the date one day back.

This will work like a charm, if we want to fill in today's report or if we want to fill last reports from previous month. Seems like most common use cases.

maciejSamerdak:
Recent fixups apply changes I described in the previous comment.

rwrzesien:
> How about taking creation date under consideration? If the last report from the month wasn't filled today, the default date will be the next day. But if the last report was filled today, then default date is last report's date.

Makes sense, I think it is good trade of between moving to next date and staying on the same day.

employees/views.py
rwrzesien: Not big issue, but rather optimization. The way you did it now is OK.

Issue #367, opened 2019-07-01
H:M input shouldn't be allowed in hours' mask


maciejSamerdak:
The default behaviour is that H:M is passed as H:0M. This is troublesome and inconsistent with the mask. We should block out the H:M format and enforce two digits input for minutes.

Should be done:
 ------------
 - H:M input shouldn't be validated

rwrzesien:
Sorry for being picky, but I think it should be:
``
H:M input should be validated as invalid.
```

PR #378, opened 2019-07-02
Bugfix add more restrict setting work_hours value

kbeker:
Resolves https://github.com/Code-Poets/sheetstorm/issues/367

### DIFF
--- a/employees/common/strings.py
+++ b/employees/common/strings.py
@@ -4,6 +4,8 @@ from django.utils.translation import ugettext_lazy as _
 
 from utils.mixins import NotCallableMixin
 
+from .constants import ReportModelConstants
+
 
 class ReportListStrings(NotCallableMixin, Enum):
     PAGE_TITLE = _("Reports")
@@ -94,11 +96,12 @@ class ProjectReportDetailStrings(NotCallableMixin, Enum):
 
 
 class ReportValidationStrings(NotCallableMixin, Enum):
+    WORK_HOURS_MINUTES_ARE_INCORRECT = _("Possible values for minutes are 0, 15, 30 and 45")
     WORK_HOURS_SUM_FOR_GIVEN_DATE_FOR_SINGLE_AUTHOR_EXCEEDED = _(
         "Sum of work hours of all reports from a given day for single author must not exceed 24."
     )
     WORK_HOURS_MIN_VALUE_NOT_EXCEEDED = _(
-        "Minimum value for work hours for single report must not be less than 1 minute"
+        f"Minimum value for work hours for single report must not be less than {int(ReportModelConstants.MIN_WORK_HOURS.value.total_seconds() / 60)} minutes"
     )
     WORK_HOURS_FIELD_NOT_TIMEDELTA_INSTANCE = _("Work hours field must be instance of timedelta")
     WORK_HOURS_WRONG_FORMAT = _("Acceptable format for work hours field is: HH:MM and HH")
--- a/employees/models.py
+++ b/employees/models.py
@@ -79,7 +79,11 @@ class Report(models.Model):
 
         if hasattr(self, "author"):
             if not isinstance(self.work_hours, timedelta):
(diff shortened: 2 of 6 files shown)


employees/common/strings.py
```diff
@@ -94,11 +94,12 @@ class ProjectReportDetailStrings(NotCallableMixin, Enum):
 
 
 class ReportValidationStrings(NotCallableMixin, Enum):
+    WORK_HOURS_MINUTES_ARE_INDIVISIBLE_BY_FIFTEEN = _("Possible values for minutes are 0, 15, 30 and 45")
     WORK_HOURS_SUM_FOR_GIVEN_DATE_FOR_SINGLE_AUTHOR_EXCEEDED = _(
         "Sum of work hours of all reports from a given day for single author must not exceed 24."
     )
     WORK_HOURS_MIN_VALUE_NOT_EXCEEDED = _(
-        "Minimum value for work hours for single report must not be less than 1 minute"
+        "Minimum value for work hours for single report must not be less than 15 minutes"
```
rwrzesien: Why not using `MIN_WORK_HOURS` constant here? Like this `f"must not be less than{ReportModelConstants.MIN_WORK_HOURS.total_seconds() / 60}"`.

employees/models.py
```diff
@@ -79,13 +79,17 @@ def clean(self) -> None:
 
         if hasattr(self, "author"):
             if not isinstance(self.work_hours, timedelta):
-                raise ValidationError(message=ReportValidationStrings.WORK_HOURS_FIELD_NOT_TIMEDELTA_INSTANCE.value)
+                raise ValidationError(message=ReportValidationStrings.WORK_HOURS_WRONG_FORMAT.value)
+            if self.work_hours < ReportModelConstants.MIN_WORK_HOURS.value:
+                raise ValidationError(message=ReportValidationStrings.WORK_HOURS_MIN_VALUE_NOT_EXCEEDED.value)
+            if self.work_hours.seconds % timedelta(minutes=15).seconds != 0:
```
rwrzesien: `seconds` should work with our cases but using `total_seconds()` is always safer.

employees/common/strings.py
```diff
@@ -94,11 +94,12 @@ class ProjectReportDetailStrings(NotCallableMixin, Enum):
 
 
 class ReportValidationStrings(NotCallableMixin, Enum):
+    WORK_HOURS_MINUTES_ARE_INDIVISIBLE_BY_FIFTEEN = _("Possible values for minutes are 0, 15, 30 and 45")
```
rwrzesien: I would give it more generic name, because the requirement might change in the future, then the message will have to change, and then the enum item name will have to change too.

employees/tests/test_unit_report_model.py
```diff
@@ -118,27 +118,41 @@ def test_report_model_author_field_should_not_be_empty(self):
...
 
+    def test_report_model_work_hours_field_should_not_accept_value_which_is_not_divisible_by_15(self):
+        self.field_should_not_accept_input(
+            "work_hours",
+            datetime.timedelta(minutes=16),
+            error_message=str(ReportValidationStrings.WORK_HOURS_MINUTES_ARE_INDIVISIBLE_BY_FIFTEEN.value),
+        )
+
+    def test_report_model_work_hours_field_should_accept_value_which_is_divisible_by_15(self):
+        self.field_should_accept_input("work_hours", datetime.timedelta(minutes=30))
```
rwrzesien: You can easily parametrize this test with all 4 possible values.

employees/common/strings.py
kbeker: Done

employees/models.py
kbeker: You're right. Done :)

employees/tests/test_unit_report_model.py
kbeker: Done

employees/common/strings.py
kbeker: Okay, done :)

PR #379, opened 2019-07-03
Fill context on invalid form response

maciejSamerdak:
On project joining in report list view, when the form is invalid the response is `render_to_response`.
The problem is, the provided context is insufficient and also the joining feature `project_form` replaces the report's `form`.
This causes an error in `test_custom_report_list_view_should_handle_no_project_being_selected_in_project_form_on_post` unit test at branch `feature-rebuild-user-reports-interface`, because during render the template refers to non-existing fields (since the forms are switched).

**NOTE:** This is not something that can be experienced via the regular use of Sheet Storm. This use case is merely an extra safety measure to prevent app from crashing on post with insufficient form data. Normally this is impossible.

### DIFF
--- a/employees/views.py
+++ b/employees/views.py
@@ -218,7 +218,7 @@ class ReportListCreateProjectJoinView(MonthNavigationMixin, ProjectsWorkPercenta
             logger.debug(f"User with id: {request.user.pk} join to the project with id: {project.pk}")
             return redirect(self.get_success_url())
         else:
-            return self.render_to_response(context={"form": form})
+            return self.render_to_response(context=self.get_context_data().update({"project_form": form}))
 
 
 class ReportDetailBase(UpdateView):


employees/views.py
```diff
@@ -201,7 +201,7 @@ def _handle_join_request(self, request: HttpRequest) -> HttpResponse:
             logger.debug(f"User with id: {request.user.pk} join to the project with id: {project.pk}")
             return redirect(self.get_success_url())
         else:
-            return self.render_to_response(context={"form": form})
+            return self.render_to_response(context=self.get_context_data())
```
Szymiks: I think that now you are not returning invalid project form and user will not get any information what happened. I am not sure what solution would be better but maybe you could `update` this dictionary ?

maciejSamerdak: It's a good point. Truth be told, I don't know how this would affect the view though. This was created in case an empty form is submitted. Normally this is impossible, since empty form is created when there are no projects to join to. But in such case, the form is hidden from user in the view, so form's error feedback is useless. Although in case of a different validation error (which I can't think of right now, but since we're considering manual requests here I guess more are possible) it could be different.

I suppose at the very least for formality's sake, the validated form should be passed to response.
@rwrzesien @kbeker Would that be the most appropriate approach?

rwrzesien: I see this is resolved already?

PR #383, opened 2019-07-04
Chore add type annotations in exports.py file

### LINKED ISSUE #382: Chore type annotations in `export.py` file
Should be done:
 ------------
 - Add type annotations in `export.py` file

kbeker:
Resolves #382

### DIFF
--- a/employees/common/exports.py
+++ b/employees/common/exports.py
@@ -1,4 +1,11 @@
+from decimal import Decimal
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
+
 from bs4 import BeautifulSoup
+from django.utils.datetime_safe import datetime
 from openpyxl import Workbook
 from openpyxl.cell import Cell
 from openpyxl.styles import Alignment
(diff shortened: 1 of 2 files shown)


employees/common/exports.py
```diff
@@ -199,7 +206,9 @@ def _summarize_user_reports(self):
         total_hours_cell.number_format = constants.TOTAL_HOURS_FORMAT.value
         set_format_styles_for_main_cells(total_hours_cell, is_header=False)
 
-    def _get_report_date_and_daily_hours(self, current_report: Report, reports_subset: ReportQuerySet):
+    def _get_report_date_and_daily_hours(
+        self, current_report: Report, reports_subset: ReportQuerySet
+    ) -> Tuple[Optional[datetime], Optional[Decimal]]:
```
maciejSamerdak: I can see we are still using Decimal to represent work hours sum. Should this be changed as well now that we're using the time delta to represent hours?

kbeker: Yes, I saw it too, but for this PR I did not change it. If you can please add issue for that @maciejSamerdak

PR #384, opened 2019-07-04
Bugfix missing work hours during updating reports

kbeker:
No issue for that

### DIFF
--- a/employees/forms.py
+++ b/employees/forms.py
@@ -10,6 +10,7 @@ from django.db.models import QuerySet
 from django.forms import HiddenInput
 from django.forms import TextInput
 from django.utils.dateparse import parse_duration
+from django.utils.duration import duration_string
 
 from common.convert import convert_string_work_hours_field_to_hour_and_minutes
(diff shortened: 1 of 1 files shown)


employees/forms.py
```diff
@@ -48,7 +49,10 @@ def clean(self, value: str) -> str:
         return f"{hours}:{minutes}:00"
 
     def prepare_value(self, value: str) -> str:
-        return f"{value}:00"
+        if isinstance(value, datetime.timedelta):
+            return duration_string(value)
+        else:
+            return f"{value}:00"
```
maciejSamerdak: What is happening here? When the value isn't considered a timedelta?

kbeker: Because this value is passed to form in html so it need to be string :)

maciejSamerdak: Ok, so I understand that the error only occurs when we're writing in full hours, since it's not recognised as the timedelta? Hence the `f"{value}:00"`

PR #390, opened 2019-07-05
Bugfix add auto assigning manager if he create project

### LINKED ISSUE #118: As a manager, when I create a project and don't assign myself to it, I can't edit it.
As a manager, when I create a project and don't assign myself to it, I can't edit it. Is this how it should work?

kbeker:
Resolves #118

### DIFF
--- a/managers/forms.py
+++ b/managers/forms.py
@@ -1,17 +1,38 @@
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+
 from bootstrap_datepicker_plus import DatePickerInput
 from django import forms
+from django.forms import SelectMultiple
 
 from common.constants import CORRECT_DATE_FORMAT
 from managers.models import Project
+from users.models import CustomUser
+
+
+class ManagerSelectMultiple(SelectMultiple):
+    def optgroups(self, name: str, value: List, attrs: Optional[Dict] = None) -> List:
+        self.choices.queryset = CustomUser.objects.exclude(user_type=CustomUser.UserType.EMPLOYEE.name).exclude(
+            pk=self.user_pk if hasattr(self, "user_pk") else None
+        )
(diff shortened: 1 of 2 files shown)


managers/views.py
```diff
@@ -102,6 +102,12 @@ def get_success_url(self) -> str:  # pylint: disable=no-self-use
         return reverse("custom-projects-list")
 
     def form_valid(self, form: ProjectAdminForm) -> HttpRequest:
+        if self.request.user not in form.cleaned_data["managers"]:
+            assert not self.request.user.is_employee
```
maciejSamerdak: It seems overly cautious to me, since AFAIK we already successfully restrict access for unauthorised personel.
But, you know, personally I don't mind it much, really. 😁

managers/forms.py
```diff
@@ -1,8 +1,20 @@
...
 
 from common.constants import CORRECT_DATE_FORMAT
 from managers.models import Project
+from users.models import CustomUser
+
+
+class ManagerSelectMultiple(SelectMultiple):
+    def optgroups(self, name: str, value: List, attrs: Optional[Dict] = None) -> List:
+        self.choices.queryset = CustomUser.objects.exclude(user_type=CustomUser.UserType.EMPLOYEE.name)
+        return super().optgroups(name, value, attrs)
```
maciejSamerdak: Since the user is assigned either way, the user should be excluded from the field and the field should be marked as optional.

kbeker: Done

managers/views.py
kbeker: I gave assert here only for safety-check purposes. It already gave me feed back when I set there `assert self.request.user.is_manager` and I completely forgot about `ADMIN` role. So it might be helpful :)

PR #391, opened 2019-07-05
Chore change type annotation from Decimal to timedelta

### LINKED ISSUE #387: Change work hours sum type from Decimal to timedelta.
The sum of work hours still returned as Decimal. I guess we should switch it to timedelta. As long as displaying sum as timedelta in HTML won't require some extra code for formatting, we will cut down on some extra code for sure (I'm thinking about functions replacing doth with colon in the Decimal representation).

Should be done:
 ------------
 - The sum should be returned in timedelta.
 - 
 - 
 -

kbeker:
Resolves #387

### DIFF
--- a/employees/common/exports.py
+++ b/employees/common/exports.py
@@ -1,4 +1,4 @@
-from decimal import Decimal
+from datetime import timedelta
 from typing import Dict
 from typing import List
 from typing import Optional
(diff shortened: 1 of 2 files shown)


Szymiks:
> You're not giving me much to moan about here, you know...

@maciejSamerdak you can always look at my code

maciejSamerdak:
@Szymiks But this is not your pull request now, is it? 😛

PR #408, opened 2019-07-08
Feature Add notifications models

### LINKED ISSUE #394: Notifications: Create NotificationType, Notification and NotificationEnabledForUser models
Requires https://github.com/Code-Poets/sheetstorm/issues/393.

Create `NotificationType`, `Notification` and `NotificationEnabledForUser` models described in https://github.com/Code-Poets/sheetstorm/issues/76.

kbeker:
Resolves #394


notifications/models.py
```diff
@@ -0,0 +1,39 @@
...
+    type = CharField(choices=NotificationType.choices(), max_length=NotificationTypeConstants.length.value)
+    created_at = models.DateTimeField(auto_now_add=True)
+    is_read_at = models.DateTimeField(null=True)
+
+
+class NotificationEnabledForUser(models.Model):
+    user = ForeignKey(CustomUser, on_delete=models.CASCADE, related_name="notification_user_enabled")
+    type = CharField(choices=NotificationType.choices(), max_length=NotificationTypeConstants.length.value)
+    is_enabled = BooleanField(default=False)
+    values = JSONField()
```
kbeker: @rwrzesien I was thinking about this field and I think that instead of `JSONField` I would prefer separate database model for all values needed for notifications. What do you think?

kbeker: It's just because I don't like fields which are like data blobs :P

kbeker:
Closing because changed in requirements.

Issue #412, opened 2019-07-09
Change adding employees and manager method in project edition and creation.


Karrp:
Choosing users with pressed `Ctrl` key is unintuitive and can cause uncheck all users if you forget about `Ctrl` key.

I propose something like is in github with labels:
![image](https://user-images.githubusercontent.com/37847362/60870891-f8a26400-a231-11e9-9e45-3d74f04e7860.png)

You can check, uncheck, search and see checked fields on the top of list when open again. And maybe more.

Should be done:
 ------------
 -  change input method to more intuitive - widget? bootstrap?

PR #416, opened 2019-07-09
Feature improve managers and employees selection

dybi:
resolves: #412


managers/forms.py
```diff
@@ -5,14 +5,22 @@
...
 from common.constants import CORRECT_DATE_FORMAT
 from managers.models import Project
 from users.models import CustomUser
 
 
-class ManagerSelectMultiple(SelectMultiple):
+class Select(Select2MultipleWidget):
+    def _get_media(self) -> forms.Media:
+        media = super()._get_media()
+        return media + forms.Media(css={"all": "employees/style.css"})
```
rwrzesien: Personally I would be for separating as much as possible frontend and backend layers. You could add it to the template in which this form will used. Maybe that would sound like less reusability but with current way you tightly couple this stylesheet with this form, and if someone wants to replace the frontend with different one, he still needs to mess up with backend code.

But as I have said, this is my personal approach, so I am just writing this as a suggestion.

If you still decide to go this way, you could still extract styles needed by this field/form to separate stylesheet and give it name related to the field/form.

managers/forms.py
```diff
@@ -5,14 +5,22 @@
...
+
+    media = property(_get_media)
+
+
+class ManagerSelectMultiple(Select):
```
rwrzesien: :+1: nice way to cover both of them with minimum code.

sheetstorm/urls.py
```diff
@@ -19,4 +19,5 @@
     url(r"^", include("users.urls")),
     url(r"^managers/", include("managers.urls")),
     url(r"^", include("employees.urls")),
+    url(r"^select2/", include("django_select2.urls")),
```
rwrzesien: According to docs (https://django-select2.readthedocs.io/en/latest/django_select2.html#module-django_select2.urls), it is only need when "you use any ‘Model’ fields". I have no idea what the mean by that - that when we use it for `ModelForms`? Anyway, it would be great if you could check if this is really needed (I am very careful when it comes to adding new urls from 3rd party libs).

employees/static/employees/style.css
```diff
@@ -208,3 +208,7 @@ textarea.form-control, #id_description {
 .description-column > p > a {
     color: deepskyblue;
 }
+
+.select2-results__option[aria-selected=true]{
```
Szymiks: Add space before the curly bracket, please.

sheetstorm/urls.py
dybi: it seems to work without it - good catch! :+1:

managers/forms.py
dybi: done!

sheetstorm/urls.py
```diff
@@ -19,4 +19,5 @@
     url(r"^", include("users.urls")),
     url(r"^managers/", include("managers.urls")),
     url(r"^", include("employees.urls")),
+    # url(r"^select2/", include("django_select2.urls")),
```
rwrzesien: @dybi why not removed? :)

PR #417, opened 2019-07-09
Bugfix work hours display in report update

dybi:
No issue for that - quick fix for problem found by @Karrp  (hours mask was not applied in all views)

### DIFF
--- a/employees/templates/employees/project_report_detail.html
+++ b/employees/templates/employees/project_report_detail.html
@@ -26,5 +26,7 @@
 {% endblock %}
 
 {% block extra_script %}
-{{ form.media }}
+  {{ form.media }}
+  <script src="//code.jquery.com/jquery-3.3.1.js"></script>
+  <script src="//cdnjs.cloudflare.com/ajax/libs/jquery.mask/1.14.10/jquery.mask.js"></script>
 {% endblock %}
--- a/employees/templates/employees/project_report_detail.html
+++ b/employees/templates/employees/project_report_detail.html
@@ -27,6 +27,5 @@
 
 {% block extra_script %}
   {{ form.media }}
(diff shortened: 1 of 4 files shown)


employees/templates/employees/project_report_detail.html
```diff
@@ -26,5 +26,7 @@ <h1>{{ UI_text.PAGE_TITLE.value }}{{ report.project }} ({{ report.date }})</h1>
 {% endblock %}
 
 {% block extra_script %}
-{{ form.media }}
+  {{ form.media }}
+  <script src="//code.jquery.com/jquery-3.3.1.js"></script>
```
rwrzesien: This one is already in `base.html` template, so it should be loaded in every template that extends `base.html` template. Are you sure it is needed too? If yes, it means something is fundamentally wrong with our scripts loading in templates.

employees/templates/employees/report_detail.html
```diff
@@ -33,12 +33,13 @@ <h1>{{ UI_text.PAGE_TITLE.value }}{{ report.project }} ({{ report.date }})</h1>
...
   {{ form.media }}
-    <script src="//code.jquery.com/jquery-3.3.1.js"></script>
-    <script src="//code.jquery.com/ui/1.12.1/jquery-ui.js"></script>
-    <script type="text/javascript">
-      var discard_text = "{{ UI_text.DELETE_POPUP_NO.value }}";
-      var confirmation_text = "{{ UI_text.DELETE_POPUP_YES.value }}";
-      var redirect_url =﻿"{% url 'custom-report-delete' pk=report.pk %}";
-    </script>
-    <script src="{% static 'employees/scripts/basic_popup_window.js' %}"></script>
+  <script src="//code.jquery.com/jquery-3.3.1.js"></script>
```
rwrzesien: Same here, it should work without this.

PR #421, opened 2019-07-10
Feature add notifications

kbeker:
I don't assign as resolve blueprint which @rwrzesien created because now it is not done what is in this blueprint


users/templates/user_notifications_list.html
```diff
@@ -0,0 +1,39 @@
...
+</h1>
+<br>
+<ul>
+    <div class="container">
+        <div class="table-responsive">
+            {% if users_days_without_report %}
+                <table class="table table-bordered">
+                    <thead>
+                        <tr>
+                            <td>{% trans 'Id' %}</td>
```
dybi: maybe we could make headers **in bold**?

users/views.py
```diff
@@ -210,3 +211,41 @@ def get_context_data(self, **kwargs: Any) -> dict:
...
+            days_without_report_list = []
+            latest_report = user.report_set.order_by("-date").first()
+
+            if latest_report is not None:
+                latest_report_date = latest_report.date
+
+                for days_delta in range(1, (today - latest_report_date).days):
+                    no_report_day = latest_report_date + datetime.timedelta(days=days_delta)
+
+                    if no_report_day.weekday() not in [5, 6]:
```
dybi: just one suggestion here:
maybe use `.isoweek()` and add comment, that Saturday and Sunday should be skipped

users/views.py
```diff
@@ -210,3 +211,41 @@ def get_context_data(self, **kwargs: Any) -> dict:
...
+    template_name = "user_notifications_list.html"
+    model = CustomUser
+
+    def get_queryset(self) -> QuerySet:
+        return CustomUser.objects.filter(projects__managers__pk=self.request.user.pk).distinct()
+
+    def get_context_data(self, *, _object_list: Any = None, **kwargs: Any) -> dict:
+        context_data = super().get_context_data(**kwargs)
+        users_queryset = self.get_queryset()
+        today = datetime.date.today()
```
dybi: this could be inlined

users/views.py
```diff
@@ -210,3 +211,41 @@ def get_context_data(self, **kwargs: Any) -> dict:
...
+        users_days_without_report_dict = {}
+        for user in users_queryset:
+            days_without_report_list = []
+            latest_report = user.report_set.order_by("-date").first()
```
dybi: this could be inlined

users/tests/test_views.py
```diff
@@ -272,3 +276,66 @@ def test_registered_user_with_wrong_activation_link_will_not_activate_account(se
...
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, NotificationUserListView.template_name)
+
+    def test_employee_should_not_be_able_get_notification_view(self):
+        self.client.force_login(self.employee)
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 302)
+
+    def test_manager_should_get_only_notification_how_many_days_employee_did_not_add_report(self):
```
dybi: maybe use `subTest` here?

users/templates/user_notifications_list.html
rwrzesien: I think we should use CSS for this, it is easier to change it there then.

users/templates/base.html
```diff
@@ -76,6 +77,19 @@
             </li>
             {% endif %}
             {% if user.user_type == manager or user.user_type == admin %}
+            <li>
+                {% if request.path|startswith:custom_users_notifications %}
```
rwrzesien: Please don't do it this way. Use `{% with %}` to assign `if` condition result to variable and write two small `ifs` to set active class and static file path.

How about making a partial html template and reuse it for other menu items?

users/templates/user_notifications_list.html
```diff
@@ -0,0 +1,39 @@
+{% extends 'base.html' %}
+
+{% load i18n %}
+{% load data_structure_element_selectors %}
+
+{% block content %}
+<h1>
+    &nbsp; {% trans 'Notifications' %} &nbsp;
```
rwrzesien: Why the extra spaces? :)

users/tests/test_views.py
rwrzesien: Or use `parametrized`.

users/views.py
```diff
@@ -210,3 +211,41 @@ def get_context_data(self, **kwargs: Any) -> dict:
...
+)
+class NotificationUserListView(ListView):
+    template_name = "user_notifications_list.html"
+    model = CustomUser
+
+    def get_queryset(self) -> QuerySet:
+        return CustomUser.objects.filter(projects__managers__pk=self.request.user.pk).distinct()
+
+    def get_context_data(self, *, _object_list: Any = None, **kwargs: Any) -> dict:
+        context_data = super().get_context_data(**kwargs)
```
rwrzesien: As discussed, I think this can be done with single queryset.

users/views.py
kbeker: Sure, changed :)

users/views.py
kbeker: Done

users/templates/user_notifications_list.html
kbeker: Done

users/tests/test_views.py
kbeker: Done :)

users/templates/user_notifications_list.html
kbeker: I just copied it from other template :) Fixed :)

users/templates/base.html
kbeker: Partial html template for menu items sound good, but I think that it might be another issue instead of doing this in this PR. Could you please create an issue with task @rwrzesien ?

users/views.py
kbeker: Done

users/templates/user_notifications_list.html
```diff
@@ -0,0 +1,39 @@
...
+
+{% load i18n %}
+{% load data_structure_element_selectors %}
+
+{% block content %}
+<h1>
+    &nbsp;{% trans 'Notifications' %}&nbsp;
+</h1>
+<br>
+<ul>
```
rwrzesien: Why is it wrapped in <ul>?

users/static/users/center_style.css
```diff
@@ -433,3 +433,7 @@ span {
     padding: 2%;
     float: left;
 }
+
+#top-table-part {
```
rwrzesien: It shouldn't be id, but class, as it is repeatable multiple times. If you want an id, create it for whole table, as it is used once, ie `#notification-table` and access to cells using xpath like `#notification-table thead tr td`.

users/templates/base.html
rwrzesien: Added https://github.com/Code-Poets/sheetstorm/issues/430

But please still do this using `{% with %}`, as current way simply duplicates the code.

users/templates/user_notifications_list.html
rwrzesien: Still there :)

users/views.py
```diff
@@ -210,3 +213,42 @@ def get_context_data(self, **kwargs: Any) -> dict:
...
+    check_permissions(allowed_user_types=[CustomUser.UserType.ADMIN.name, CustomUser.UserType.MANAGER.name]),
+    name="dispatch",
+)
+class NotificationUserListView(ListView):
+    template_name = "user_notifications_list.html"
+    model = CustomUser
+
+    def get_queryset(self) -> QuerySet:
+        return (
+            CustomUser.objects.filter(projects__managers__pk=self.request.user.pk)
```
rwrzesien: Use `self.get_queryset` instead `CustomUser.objects`.

users/views.py
rwrzesien: Then you don't need the whole calculation below this comment. You already have date of last report for every user. You can easily calculate the difference between today and that date. In fact, when I was writing about doing this with single queryset, I have meant that it can be done in that queryset too using `F` object. Rough example:

```
for x in CustomUser.objects.annotate(last_report_date=Coalesce(Max("report__date"), None)).annotate(z=timezone.now() - F('last_report_date')): 
    ...:     print (x.z) 
    ...:                                                                                                                                                 
None
10 days, 21:49:57.549116
43 days, 21:49:57.549116
19 days, 21:49:57.549116
43 days, 21:49:57.549116
43 days, 21:49:57.549116
800 days, 21:49:57.549116
None
None
```

The only thing missing is skipping weekends but that can be added as simple template tag called `substract_weekdays_amount_from_last_days(days)` where `days` will be int taken from `z`.

Btw. I think that weekdays should count too, as the result might be ambiguous, for example there will be 4 days which can mean that there were really 4 days, or 6 but there was a weekend in meantime. Also, it is easier to notice that someone didn't report for example for whole 14 days but it is known he was on 2 weeks holidays, than to see 14 days without being sure if that is more than 2 weeks because weekends are cut off.

users/templates/base.html
rwrzesien: OK nevermind, I have tried to apply this, but there were too many differences between menu items.

users/views.py
kbeker: Done

users/templates/user_notifications_list.html
kbeker: Now I don't really know about which spaces are you talking about? :P

users/views.py
kbeker: Done. I left skipping weekends as request from client.

users/templates/user_notifications_list.html
rwrzesien: What is the `&nbsp;` there then? :)

users/static/users/center_style.css
kbeker: Done

users/templates/user_notifications_list.html
kbeker: Deleted :)

users/views.py
```diff
@@ -210,3 +216,41 @@ def get_context_data(self, **kwargs: Any) -> dict:
...
+class NotificationUserListView(ListView):
+    template_name = "user_notifications_list.html"
+    model = CustomUser
+
+    def get_queryset(self) -> QuerySet:
+        return (
+            super()
+            .get_queryset()
+            .filter(
+                Q(is_active=True)
```
rwrzesien: There is no need to use `Q` if and`&` as in this case this is default `.filter()` behaviour.

kbeker: Ahh, I tested different combinations of querysets and this left after this battlefield :P Deleted.

users/templates/user_notifications_list.html
```diff
@@ -0,0 +1,36 @@
...
+
+{% block content %}
+<h1>
+    &nbsp;{% trans 'Notifications' %}&nbsp;
+</h1>
+<div class="container">
+    <div class="table-responsive">
+        {% if users_days_without_report %}
+            <table class="table table-bordered">
+                <thead class="notification_table_head">
```
rwrzesien: This class is not needed anymore.

users/templates/user_notifications_list.html
rwrzesien: What about this one? `&nbsp;` mean non-breaking space in HTML. I am not saying its wrong to add, I was just curious why you have added them?

users/views.py
rwrzesien: Still there, you don't need to wrap the each filter parts in `Q`.

users/tests/test_views.py
```diff
@@ -272,3 +277,67 @@ def test_registered_user_with_wrong_activation_link_will_not_activate_account(se
...
+        self.project.members.add(self.employee)
+        self.url = reverse("custom-users-notifications")
+
+    def _check_response(self, response, status_code, contains):
+        self.assertTemplateUsed(response, NotificationUserListView.template_name)
+        self.assertEqual(response.status_code, status_code)
+        for element in contains:
+            self.assertContains(response, element)
+
+    def test_manager_should_be_able_get_notification_view(self):
```
rwrzesien: You should also add this check to `AccessPermissionsTestCase`.

users/templates/user_notifications_list.html
kbeker: To be honest I didn't know what it is and what it does. I just copied it from different template. I see now what it is for and this is not needed here. Deleting :)

users/views.py
kbeker: Okaaay, so to sum up we need to wrap filters with `Q` when we use some combinations with `AND` and `OR`?

users/templates/user_notifications_list.html
kbeker: Right, deleted

users/tests/test_views.py
kbeker: You're right. I've deleted them from `test_views` an`d moved to `test_access_permissions`

users/views.py
rwrzesien: Yes, also `AND` is used by `filter()` by default, so it is used only in some special cases. So `Q` is mostly used when you want to apply `OR`.

Issue #370, opened 2019-07-01
OpenOffice does not display full description in .xlsx view


maciejSamerdak:
If at all possible, the rows' should be formatted in a way that would allow the full description to be displayed.

Should be done:
 ------------
 - The rows should have variable height when .xls is opend in OpenOffice

PR #422, opened 2019-07-10
Bugfix does not display full description in .xlsx

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/370


Szymiks:
After push I found that it not works as should so I closed it, but after this I pushed it with force and I can't reopen it

PR #423, opened 2019-07-10
Bugfix OpenOffice does not display full description

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/370

There is PR which is related to this one.
Related: https://github.com/Code-Poets/sheetstorm/pull/426

### DIFF
--- a/employees/common/exports.py
+++ b/employees/common/exports.py
@@ -40,7 +40,7 @@ def set_format_styles_for_main_cells(cell: Cell, is_header: bool) -> None:
     )
 
 
-def set_and_fill_description_cell(cell: Cell, cell_value: str) -> None:
+def set_and_fill_cell(cell: Cell, cell_value: str) -> None:
     wrapped_alignment = Alignment(vertical=constants.VERCTICAL_TOP.value, wrap_text=True)
     cell.alignment = wrapped_alignment
     cell.value = cell_value
@@ -51,6 +51,8 @@ def set_and_fill_hours_cell(cell: Cell, cell_value: str) -> None:
         cell.value = constants.TIMEVALUE_FORMULA.value.format(cell_value)
     else:
         cell.value = cell_value
+    alignment = Alignment(vertical=constants.VERCTICAL_TOP.value)
+    cell.alignment = alignment
     cell.number_format = constants.HOURS_FORMAT.value
 
 
@@ -124,6 +126,7 @@ class ReportExtractor:
             constants.DESCRIPTION_HEADER_STR.value: convert_markdown_html_to_text(report.markdown_description),
         }
         self._fill_current_report_data(storage_data)
+        self._set_row_height(str(storage_data[constants.DESCRIPTION_HEADER_STR.value]))
         self._current_row += 1
 
     def _prepare_worksheet(self, employee_name: str) -> None:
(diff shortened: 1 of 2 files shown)


employees/common/exports.py
```diff
@@ -124,6 +126,8 @@ def _fill_single_report(self, report: Report, reports: ReportQuerySet) -> None:
             constants.DESCRIPTION_HEADER_STR.value: convert_markdown_html_to_text(report.markdown_description),
         }
         self._fill_current_report_data(storage_data)
+        self._set_row_height(str(storage_data[constants.DESCRIPTION_HEADER_STR.value]))
+
```
rwrzesien: Remove new extra line.

employees/common/exports.py
```diff
@@ -230,6 +232,24 @@ def _get_report_date_and_daily_hours(
...
+    def _set_row_height(self, description: str) -> None:
+        # If the description has \ n or text exceeds 100 characters, it is automatically moved to the new line.
+
+        new_lines_in_description = description.count("\n")
+        splitted_description = description.split("\n")
+        rows_in_xlsx = 0
+
+        for row in splitted_description:
+            length_of_row = len(row)
+            rows_in_xlsx = int(length_of_row / 100)
```
rwrzesien: I don't get why you need iterate over all `splitted_description` to set this value, as it will have values set in last iteration of `spitted_description`. You can simply just select the last element with `splitted_description[-1]`.

Szymiks: I iterate over all because over description column width is 100 chars and then We have wrapped text, so I have to check if there is no some rows with more than 100 chars. Because we could write reports like that:
```
First row
Second row
Third row
```
or just in one line without any enters.
``` 
First row Second row Third row
```
How will help me taking just last element ? I found that this is not working in all options so I fixed it up but there is again for and one more if condition

employees/common/exports.py
Szymiks: Done

employees/common/exports.py
```diff
@@ -230,6 +231,26 @@ def _get_report_date_and_daily_hours(
...
+
+        new_lines_in_description = description.count("\n")
+        splitted_description = description.split("\n")
+        rows_in_xlsx = 0
+
+        if len(splitted_description) == 1:
+            rows_in_xlsx = int(len(splitted_description[0]) / 100)
+        else:
+            for row in splitted_description:
+                rows_in_xlsx = rows_in_xlsx + int(len(row) / 100)
```
rwrzesien: You can simplify too `rows_in_xlsx += int(len(row) / 100)

employees/common/exports.py
rwrzesien: Now after the fix it makes more sense. Previously the code was working like:
```
for x in y:
    a = something(x)
```

so the result was the same as `a = something(y[-1])`.

Szymiks: Yes I know, I just realized this when I look for it this morning

employees/common/exports.py
Szymiks: Yea... I was wondering why it doesn't work but I have tried `=+` ... thanks. Done

employees/common/exports.py
```diff
@@ -230,6 +231,26 @@ def _get_report_date_and_daily_hours(
...
+        rows_in_xlsx = 0
+
+        if len(splitted_description) == 1:
+            rows_in_xlsx = int(len(splitted_description[0]) / 100)
+        else:
+            for row in splitted_description:
+                rows_in_xlsx += int(len(row) / 100)
+
+        row_height = (
+            constants.DEFAULT_ROW_HEIGHT.value * (1 + new_lines_in_description + rows_in_xlsx)
```
dybi: this `if ... else` seems unnecessary.

employees/common/exports.py
```diff
@@ -230,6 +231,26 @@ def _get_report_date_and_daily_hours(
...
+
+        new_lines_in_description = description.count("\n")
+        splitted_description = description.split("\n")
+        rows_in_xlsx = 0
+
+        if len(splitted_description) == 1:
+            rows_in_xlsx = int(len(splitted_description[0]) / 100)
+        else:
+            for row in splitted_description:
+                rows_in_xlsx += int(len(row) / 100)
```
dybi: wouldn't you need to round up? casting to int would round down.

Szymiks: I wanted to round down, because if some string has 150 chars then `150 / 100 = 1.5` so I need to extend height only for one because second part is still fitting to the width.

employees/common/exports.py
Szymiks: Yes you are right, It was needed when I was trying to implement it on first time, but now it is useless. Thanks.!

Issue #409, opened 2019-07-08
Variable default date value on report creation - skipping weekends


Karrp:
related to #366 

When creating new report and last report is from Friday and today is Monday application suggest you Saturday but should suggest Monday.

Should be done:
 ------------
 - skipping weekends

PR #427, opened 2019-07-11
Bugfix weekends should be skipped in reports date suggestion

dybi:
resolves: #409


employees/views.py
```diff
@@ -181,6 +181,8 @@ def _get_default_date(self) -> datetime.date:
         date = last_report.date + timezone.timedelta(days=1)
         if date.month != month:
             return last_report.date
+        while date.isoweekday() in [6, 7]:
+            date += timezone.timedelta(days=1)
```
rwrzesien: Wouldn't that move to Sunday when Saturday? ;)

PR #428, opened 2019-07-12
Fix urls namespace and add reverse to missing usages in tests

rwrzesien:
While cleaning up I have seen that we are still using the same url namespace for two Django apps. This should be avoided if not needed, so I have added the fix. Due to this I have discovered there were some places in which we were using hardcoded urls, so I have switched them to use `reverse`.

### DIFF
--- a/employees/tests/test_unit_custom_mixins.py
+++ b/employees/tests/test_unit_custom_mixins.py
@@ -126,9 +126,16 @@ class MonthNavigationMixinContextDataTests(TestCase):
             "path": self.mixin.request.path,
             "navigation_text": MonthNavigationText,
             "month_form": MonthSwitchForm(initial_date=datetime.date(year=2019, month=3, day=1)),
-            "next_month_url": "/reports/project/1/2019/4/",
-            "recent_month_url": f"/reports/project/1/{current_date.year}/{current_date.month}/",
-            "previous_month_url": "/reports/project/1/2019/2/",
+            "next_month_url": reverse(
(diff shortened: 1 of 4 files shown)


employees/tests/test_unit_custom_mixins.py
```diff
@@ -126,9 +126,11 @@ def test_month_navigator_get_context_data_method_should_return_all_data_necessar
             "path": self.mixin.request.path,
             "navigation_text": MonthNavigationText,
             "month_form": MonthSwitchForm(initial_date=datetime.date(year=2019, month=3, day=1)),
-            "next_month_url": "/reports/project/1/2019/4/",
-            "recent_month_url": f"/reports/project/1/{current_date.year}/{current_date.month}/",
-            "previous_month_url": "/reports/project/1/2019/2/",
+            "next_month_url": reverse("project-report-list", args=(self.mixin.kwargs["pk"], 2019, 4)),
```
dybi: I see that sometimes we use `args` and sometimes `kwargs` in url building. I would suggest that we don't mix them up. My personal choice would be `kwargs` (greater readability). @rwrzesien, what do you think?

kbeker: I would also prefer `kwargs` because we can see all keys and values associated with them. If you agree @rwrzesien I can change it for you and merge today :)

rwrzesien: Yes, I prefer `kwargs` too, I just do it in little rush yesterday, go ahead and fix it if you want :)

rwrzesien: I have updated it.

Issue #414, opened 2019-07-09
Change name of a "recent month" button


lukasz-tarka:
In reports view change misleading text.

Should be done:
 ------------
 - "Recent month" -> "current month"

PR #429, opened 2019-07-12
Change "Recent month" to "Current month" in month navigation bar

dybi:
resolves: #414


employees/common/strings.py
```diff
@@ -110,4 +110,4 @@ class ReportValidationStrings(NotCallableMixin, Enum):
 
 class MonthNavigationText(NotCallableMixin, Enum):
     SWITCH_MONTH = _("Go")
-    RECENT_MONTH = _("Recent month")
+    RECENT_MONTH = _("Current month")
```
Szymiks: Variable should not be change to `CURRENT_MONTH` as well ?

dybi: Good point:)

PR #437, opened 2019-07-15
Chore add behave based integration tests

### LINKED ISSUE #23: BaseTests mechanism

maciejSamerdak:
Integration tests implementation based on **behave** and **Selenium** frameworks.
Currently features 3 scenarios regarding user sign up feature.

Selenium uses **Chrome** browser for both headless and non-headless run support (headless run boosts performance, but looks less cool). To successfully run tests, a Chrome browser must be installed and [Chrome WebDriver](http://chromedriver.chromium.org/) must be included in PATH.

Resolves #23


maciejSamerdak:
I'm currently struggling with writing proper test for user signup, due to multiple issues with Selenium (it seems like the signup view is special and doesn't want to cooperate nicely). The errors are mostly focused around found elements not being considered interactive by Selenium (from `Sign up` link to form fields in signup view). What's more _funny_ is that everything is executed perfectly fine when performing steps with debugger.
Anyway, I'll push it as soon as I figure it out.

rwrzesien:
@maciejSamerdak 

> Selenium uses Chrome browser for both headless and non-headless run support (headless run boosts performance, but looks less cool). To successfully run tests, a Chrome browser must be installed and Chrome WebDriver must be included in PATH.

If this can't be automated or put into a script, you should add it to `README`.

> What's more funny is that everything is executed perfectly fine when performing steps with debugger.

Maybe the timeout is needed? I mean, maybe in regular tests it wants to perform action before page is fully loaded, while in debugger it takes more time.

features/environment.py
```diff
@@ -0,0 +1,17 @@
+# pylint: disable=W0601, W0613
+from behave import use_fixture
+
+from features.fixtures import selenium_browser
+
+
+def setup_debug_on_error(userdata):
+    global BEHAVE_DEBUG_ON_ERROR
```
rwrzesien: Why do you need this to be `global`?

sheetstorm/settings/development.py
```diff
@@ -14,3 +14,5 @@
 CAPTCHA_TEST_MODE = True
 
 VALID_EMAIL_DOMAIN_LIST = ["codepoets.it"]
+
+EMAIL_SIGNUP_VERIFICATION = False
```
rwrzesien: Looks like a separate feature, do we really want it this way in production?

features/steps/user_initialisation_steps.py
```diff
@@ -0,0 +1,65 @@
...
+from users.factories import AdminUserFactory
+from users.factories import UserFactory
+
+
+@given("an admin exists in database")
+def step_impl(context):
+    context.admin = AdminUserFactory()
+
+
+@when("the user signs up")
```
rwrzesien: `signs up` => `signs in`.

rwrzesien:
Looks nice at first sight, I especially like the part written in behave. I am just curious about implementation changes with user sign up - I thought we can't currently send emails?

features/environment.py
```diff
@@ -0,0 +1,17 @@
...
+    BEHAVE_DEBUG_ON_ERROR = userdata.getbool("BEHAVE_DEBUG_ON_ERROR")
```
dybi: maybe this can be part of Django `testting` settings? @rwrzesien, what do you think?

sheetstorm/settings/development.py
dybi: I agree with @rwrzesien. Eveit if it was to say I suggest we should have it `True` by default.

e2e.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
```
dybi: :+1:

features/steps/user_initialisation_steps.py
dybi: signs up = creates account,
signs in = logs is,
which case is implemented here?

features/steps/user_initialisation_steps.py
```diff
@@ -0,0 +1,65 @@
...
+    sign_in(context)
+
+
+@when("admin changes user's employee type to '{user_type}'")
+def step_impl(context, user_type):
+    other_browser = get_chrome_browser()
+    sign_in(context=context, login=context.admin.email, browser=other_browser)
+    other_browser.find_element_by_link_text("Employees").click()
+    table = other_browser.find_element_by_class_name("table")
+    button = table.find_element_by_xpath(".//tbody/child::tr[3]/child::td[7]//a")
```
dybi: isn't this `tr[3]`  and `td[7]` a bit risky hardcoding? Couldn't we have some `id` there (or sth else that would make those test less fragile)?

features/steps/user_initialisation_steps.py
```diff
@@ -0,0 +1,65 @@
...
+    update = other_browser.find_element_by_id("opener_user_account_update")
+    update.click()
+    confirm = other_browser.find_element_by_id("id_confirm_update_button")
+    confirm.click()
+
+    other_browser.quit()
+
+
+@then("the user should have an employee status")
+def step_impl(context):
```
dybi: the three steps implementations look very similar to each other (especially `manager` and `admin`). Maybe we could have them extracted to a generalized function?

features/environment.py
maciejSamerdak: The code was taken directly from behave docs, as a proposed way to handle debugging by devs. https://behave.readthedocs.io/en/latest/tutorial.html#debug-on-error-in-case-of-step-failures

features/steps/user_initialisation_steps.py
maciejSamerdak: The focus is sign up, since that's the feature intended for testing here. That's why factory-based creation is used here as a placeholder for proper code. I just have the user signed in afterwards as confirmation of successful sign up procedure.

features/steps/user_initialisation_steps.py
maciejSamerdak: From what I understood, according to @rwrzesien the Xpath should be used with majority of elements and that we should assume for testing purposes, that the precise layout is known. He also said, and I quote:

> księga ulicy mówi jasno - zmieniasz kody, zmieniasz testy

features/steps/user_initialisation_steps.py
maciejSamerdak: The problem is I'm also checking the access to key exclusive functionalities (`Projects` link for `Manager` and `Employees` for `Admin`), so it's hard to pass those arguments in step name. Unless I extract this part to separate step? Like:
```
the user should have "Manager" status
And the user should have access to "Projects" page
```
I guess it's the quickest and cleanest solution. That way we can also easily test admin's access to `Projects` if we wanted to.

sheetstorm/settings/development.py
maciejSamerdak: But I thought these are development-only settings? This setting is already introduced in `base.py` with default `True` value, therefore the production shouldn't be affected?
Or did I miss something? 😕

maciejSamerdak:
@rwrzesien 

> I am just curious about implementation changes with user sign up - I thought we can't currently send emails?

On development, the emails are displayed in console, from what I was told. Therefore the whole verification process still applies and I must redirect to messaged URL to activate an account.
But I can't access it from browser, so I had to disable this feature to even consider performing tests for sign up, without fiddling with database.

> Maybe the timeout is needed? I mean, maybe in regular test wants to perform action before page is fully loaded, while in debugger it takes more time.

Yeah, I managed to figure it out yesterday: I've set the timeout based on elements visibility and elements interactivity. The problem is the both conditions are satisfied as soon as the element is rendered, but the animation is still playing. Therefore the clicks are performed on an object that changed it's position. I'll have to write a custom condition for it.

features/steps/user_initialisation_steps.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@given("an admin exists in database")
+def step_impl(context):
+    context.admin = AdminUserFactory()
+
+
+@when("the user signs up")
+def step_impl(context):
+    # TODO: HAVE TO SOLVE CAPTCHA AND EMAIL VERIFICATION ISSUES
```
kbeker: Haven't you already solve captcha issue? Does not `CAPTCHA_TEST_MODE` setting work?

features/steps/user_initialisation_steps.py
kbeker: I love this quote

lint.sh
```diff
@@ -6,7 +6,7 @@ flake8
     --jobs=4                                                                                                \
     --max-line-length=120                                                                                   \
     --ignore=E124,E126,E128,E131,E156,E201,E221,E222,E241,E265,E271,E272,E701,F405,E501,W503                \
-    --per-file-ignores=users/common/strings.py:E501
```
kbeker: 1. Why you are deleting this file from flake ignores?
2. Please explain why do you add new one to ignore and what do you want to ignore in this file.

sheetstorm/settings/base.py
```diff
@@ -44,6 +44,7 @@
     'django_countries',
     'users',
     # 3rd party
+    'behave_django',
```
kbeker: I don't really like that this is added in base django settings. I would move it to `testing.py`

sheetstorm/settings/development.py
kbeker: @rwrzesien @dybi this setting is set in `development.py`. On server we have `local_settings.py` which inherits from `production.py` which inherits from `base.py`. There is no chance that we use setting which is set in this file. However I think that in `base.py` should be this set on `None` or even commented to enforce someone who wants to run server see what setting is really valid for him, but if you guys wants to have `True` as a value in `base.py` I would fight with this decision :). In this file for me everything is fine.

requirements.lock
```diff
@@ -4,6 +4,8 @@ astroid==2.2.5
 atomicwrites==1.3.0
```
kbeker: I think that after this PR, or maybe I will have a chance to do it during this PR to separate requirements into requirements which are necessary for proper server work and for testing purposes because this file is too big and too much unnecessary requirements is installed to virutal enviroment on server.

e2e.sh
maciejSamerdak: @kbeker At the same time it runs all the tests with appropriate configuration for behave itself (e.g. `--no-capture` for debugging) , so it is directly used to perform the tests the way we intended.

features/fixtures.py
```diff
@@ -0,0 +1,17 @@
...
+    context.browser = get_chrome_browser()
+    yield context.browser
+    context.browser.quit()
+
+
+def get_chrome_browser():
+    options = webdriver.ChromeOptions()
+    if settings.HEADLESS_BROWSER_RUN:
+        options.add_argument("headless")
+    return webdriver.Chrome(chrome_options=options)
```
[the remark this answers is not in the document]
maciejSamerdak: Yes, both Chrome browser and ChromeDriver must be independently installed. Same applies to any other browser that Selenium supports.

features/steps/user_initialisation_steps.py
```diff
@@ -0,0 +1,65 @@
+# pylint: disable=E0102, E0602
+from behave import *
```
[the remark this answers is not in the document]
maciejSamerdak: But that's exactly what the devs told me to do. 😢 
Who am I, a simple junior developer, to question the devs' wisdom? 😭 
https://behave.readthedocs.io/en/latest/tutorial.html#

features/steps/user_initialisation_steps.py
maciejSamerdak: It does, I forgot to change this comment. I guess it should be more like: `Write proper implementation reflecting user's behaviour` or something like that. However this most likely won't get merged until I do that.
Otherwise I'm requesting changes myself! 😛

features/environment.py
```diff
@@ -0,0 +1,17 @@
+# pylint: disable=W0601, W0613
```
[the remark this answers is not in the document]
maciejSamerdak: `W0613: Unused argument 'scenario' (unused-argument)` - well that's the definition. I figured it's better to silence then trying to add `_` to the argument. The latter solution messed up `get` and `post` not work in Django, so I thought it's safer to just mute pylint on this.

`W0601: Global variable 'BEHAVE_DEBUG_ON_ERROR' undefined at the module level (global-variable-undefined)` - don't blame me, that's what the devs suggested. I don't know this magic too well.

features/steps/user_initialisation_steps.py
```diff
@@ -0,0 +1,65 @@
+# pylint: disable=E0102, E0602
```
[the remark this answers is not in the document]
maciejSamerdak: `features/steps/user_initialisation_steps.py:58:0: E0102: function already defined line 11 (function-redefined)` - well that's just how `behave` works.

`Undefined variable 'when/then/given' (undefined-variable)` - I dunno whether this has to do with the import. And that import was the devs' instruction to begin with, so plz no h8 😿  I'm telling mom otherwise

lint.sh
maciejSamerdak: 1. Well, it's funny: I tried to just ADD another file to this rule, but I couldn't figure out the proper syntax to separate the files.
I mean it was supposed to be the newline character, but I couldn't get it to work.
But I checked it and `users/common/strings.py:E501` is not raised anymore, most likely due to some refactoring along the way.
So since it's useless, I just replaced it...

2. `F403 'from behave import *' used; unable to detect undefined names` - so same as before, so up for debate.
`F811 redefinition of unused 'step_impl' from line 10` - so again, that's how `behave` is expected to work.

features/environment.py
maciejSamerdak: My bad, I forgot I got rid of this functionality, since I figured it wasn't really useful for me to call debugger after the fact. So this is a part of the code that I forgot to remove.

Unless you guys think this could be useful and I should bring this functionality back?

features/environment.py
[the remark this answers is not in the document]
maciejSamerdak: I don't need that global value anymore, so it this `W0601` exception can be removed.

maciejSamerdak:
> If this can't be automated or put into a script, you should add it to `README`.

Done

features/environment.py
maciejSamerdak: I removed it for now.

features/environment.py
maciejSamerdak: Done, it got removed entirely

features/steps/user_initialisation_steps.py
maciejSamerdak: Done, applied the concept above

features/steps/user_initialisation_steps.py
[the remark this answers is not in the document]
maciejSamerdak: Done

features/environment.py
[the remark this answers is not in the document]
maciejSamerdak: Both warnings are removed now.

features/steps/user_initialisation_steps.py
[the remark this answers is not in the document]
maciejSamerdak: The latter got replaced with E0611, since pylint can't seem to find those decorators in behave package, somehow.

features/steps/user_initialisation_steps.py
maciejSamerdak: Done, the step got implemented

sheetstorm/settings/base.py
maciejSamerdak: Done

features/steps/helpers.py
```diff
@@ -0,0 +1,49 @@
...
+    submit_button.click()
+
+
+def cautious_click(browser, xpath, iterations=5, timeout=2):
+    iterations = iterations - 1
+    element = browser.find_element_by_xpath(xpath)
+    try:
+        sleep(timeout)
+        element.click()
+    except ElementClickInterceptedException as Error:
```
maciejSamerdak: That's the best thing I could come up with to keep this thing somewhat deterministic. Seems to work alright, as long as there's a sleep before first action (otherwise it would fail radomnly, I'm not sure why, _no comprendo el frontendo_).

features/environment.py
rwrzesien: Lets don't use globals if not needed (which is - never ;) )

features/steps/user_initialisation_steps.py
rwrzesien: Ok, I thought its signing in (loggin in).

sheetstorm/settings/development.py
rwrzesien: What I have meant is that we were not sending emails at all (I think so?) and now we are adding this functionality in not really related pull request. So the question was rather - do we want this feature? If yes than maybe it would be easier to provide it in separate pull request.

features/environment.py
rwrzesien: Depending on if we want to run this in Django context. If yes, then it is OK, but I thought originally that we want to run it independently. Do we?

features/steps/user_initialisation_steps.py
rwrzesien: I understand @dybi doubts but I also have another rule - never ever apply code to tests (meaning if something is difficult to test, still test it the hardway, don't change the code).  Why? Because if you start doing this you will be tended to do more and more shortcuts. 

I also think that it won't change that often. HTML should be written in a way that it can be easily modified with CSS for better/different looking. I think we didn't change the order of those elements since they were implemented for the first time? We only added some columns and each of those adds were rather big features.

Still, I think it could be tested in different way, at least for column. You could determine which column index has the element in header (`<thead><tr><th>`) basing on its text, and then apply same index here.

features/steps/user_initialisation_steps.py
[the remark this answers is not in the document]
rwrzesien: Listen to your fellows co-workers, not some random people from Internet ;)

sheetstorm/settings/development.py
maciejSamerdak: @rwrzesien The emails are sent to the console output, from what I was told. The activation proces still applies anyway, so I need this setting to bypass this. Otherwise I can't activate an account during tests without fiddling with database, which was undesirable in e2e.

features/steps/user_initialisation_steps.py
maciejSamerdak: Done, except:
The columns containing links don't have any names, so I search for the row containing user name and then I pick the first column after named columns (headers count).

features/steps/helpers.py
```diff
@@ -1,20 +1,49 @@
...
-    username = login_form.find_element_by_xpath(".//input[@id='username']")
+    cautious_send_keys(browser=login_form, xpath=".//input[@id='username']", sequence=login)
     password = login_form.find_element_by_xpath(".//input[@id='password']")
-    username.send_keys(login)
     password.send_keys("userpasswd")
     submit_button = login_form.find_element_by_xpath(".//input[@type='submit']")
     submit_button.click()
+
+
+def cautious_click(browser, xpath, iterations=10, timeout=1):
```
dybi: personally I would suggest using simple `while` loop over recursion ;)

features/steps/helpers.py
```diff
@@ -1,20 +1,49 @@
...
+        sleep(timeout)
+        element.click()
+    except ElementClickInterceptedException as Error:
+        if iterations > 0:
+            cautious_click(browser, xpath, iterations, timeout)
+        else:
+            raise Error
+
+
+def cautious_send_keys(browser, xpath, sequence, iterations=10, timeout=1):
```
dybi: it is almost identical to `cautious_click`. Please generalize it to `cautious_action` and add callable argument `action` that is to be performed on `element`

features/steps/user_initialisation_steps.py
```diff
@@ -26,40 +47,34 @@ def step_impl(context, user_type):
...
-@then("the user should have an employee status")
-def step_impl(context):
-    sidebar = context.browser.find_element_by_id("sidebar")
-    sidebar_header = sidebar.find_element_by_xpath(".//div[@class='sidebar-header']")
-    assert "Employee" in sidebar_header.text
-
-
-@then("the user should have a manager status")
-def step_impl(context):
+@then("the user should have '{user_type}' status")
```
dybi: :+1:

features/steps/user_initialisation_steps.py
```diff
@@ -26,40 +47,34 @@ def step_impl(context, user_type):
...
     sidebar_header = sidebar.find_element_by_xpath(".//div[@class='sidebar-header']")
-    sidebar_components = sidebar.find_element_by_xpath(".//ul[contains(@class, 'components')]")
-    assert "Manager" in sidebar_header.text
-    assert "Projects" in sidebar_components.text
+    assert user_type in sidebar_header.text
 
 
-@then("the user should have an admin status")
-def step_impl(context):
+@then("the user should have access to '{page_name}' page")
```
dybi: :+1:

features/steps/helpers.py
maciejSamerdak: Done

features/steps/helpers.py
maciejSamerdak: Done

features/steps/helpers.py
```diff
@@ -23,27 +27,40 @@ def sign_in(context, login=None, browser=None):
...
+    cautious_action(browser, xpath, "click", [], ElementClickInterceptedException, iterations, timeout)
+
+
+def cautious_send_keys(
+    browser, xpath, sequence, iterations=CAUTIOUS_ACTION_ITERATIONS, timeout=CAUTIOUS_ACTION_TIMEOUT
+):
+    cautious_action(browser, xpath, "send_keys", [sequence], ElementNotInteractableException, iterations, timeout)
+
+
+def cautious_action(
```
dybi: it can be simplified a bit - it is just a suggestion ;)
```python
def cautious_action(
    browser,
    xpath,
    action,
    expected_exception,
    iterations,
    timeout,
    *args,
    **kwargs,
):
    received_exception = None
    while iterations > 0:
        try:
            element = browser.find_element_by_xpath(xpath)
            sleep(timeout)
            getattr(element, action)(*args, **kwargs)
            break
        except expected_exception as error:
            received_exception = error
            iterations -= 1
    else:
        raise received_exception
```
please not, that we can decide to use any of `*args`, `**kwargs`

maciejSamerdak: Thanks!
Wasn't aware that `*args, **kwargs` solution would work here. Thought it's some higher level magic involving Django modules.
Never heard of **while-else** before! 👀 What a peculiar mechanism!
Also, at some point I was advised to avoid breaks in general. I was told they can make it harder on following the code and that it's better to just write proper condition for the loop. What's your opinion on this?

kbeker:
When @maciejSamerdak will sqash all fixUps and rebease on newest master this PR is ready to merge.

PR #438, opened 2019-07-16
Bugfix login and home urls

kbeker:
No issue for that

### DIFF
--- a/sheetstorm/settings/base.py
+++ b/sheetstorm/settings/base.py
@@ -12,6 +12,8 @@ https://docs.djangoproject.com/en/1.11/ref/settings/
 
 import os
 
+from django.urls import reverse_lazy
+
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
@@ -232,6 +234,7 @@ STATIC_ROOT = os.path.join(BASE_DIR, 'static')
 STATIC_URL = '/static/'
 
 AUTH_USER_MODEL = 'users.CustomUser'
+LOGIN_URL = reverse_lazy("login")
 LOGIN_REDIRECT_URL = 'home'
 LOGOUT_REDIRECT_URL = 'login'
 
--- a/sheetstorm/tests/test_views.py
+++ b/sheetstorm/tests/test_views.py
@@ -1,0 +1,23 @@
+from django.conf import settings
+from django.test import TestCase
+from django.urls import reverse
+
+from sheetstorm.views import Index
+from users.factories import UserFactory
+
+
+class IndexPageTests(TestCase):
+    def setUp(self):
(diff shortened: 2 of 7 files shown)


sheetstorm/settings/base.py
```diff
@@ -232,6 +232,7 @@
 STATIC_URL = '/static/'
 
 AUTH_USER_MODEL = 'users.CustomUser'
+LOGIN_URL = "users/accounts/login"
```
dybi: this seems to be unused

sheetstorm/urls.py
```diff
@@ -14,7 +14,10 @@
 from django.contrib import admin
 from django.urls import include
 
+from users import views
+
 urlpatterns = [
+    url(r"^$", views.index, name="home"),
```
Szymiks: I am just wondering, about changing this view to TemplateView because  we are using CBV, everywhere. However, maybe it is not needed.

kbeker: Good point. Changed :+1:

sheetstorm/settings/base.py
kbeker: This is not unused. This is used by django auth and this is using `settings.LOGIN_URL` which normally is hardcoded for "accounts/login"

rwrzesien: Still please use `reverse` or `reverse_lazy`.

sheetstorm/tests/test_views.py
```diff
@@ -0,0 +1,23 @@
...
+from sheetstorm.views import Index
+from users.factories import UserFactory
+
+
+class IndexPageTests(TestCase):
+    def setUp(self):
+        self.url = reverse("home")
+        self.user = UserFactory()
+
+    def test_not_logged_user_got_redirected_to_login_page(self):
```
rwrzesien: Isn't that testing framework functionality?

kbeker: But these are simple 2 unit tests to be sure that we have access to home page :)

rwrzesien: Lets keep it then, but please also add to test `test_access_permissions`.

sheetstorm/settings/base.py
kbeker: Oh, I forgot about `reverse_lazy`. Done

sheetstorm/urls.py
```diff
@@ -14,7 +14,10 @@
 from django.contrib import admin
 from django.urls import include
 
+from sheetstorm import views
```
dybi:
```suggestion
from . import views
```
I guess it might by more django-way ;)

kbeker: Sure, changed :)

sheetstorm/tests/test_views.py
kbeker: Also added :)

Issue #435, opened 2019-07-15
Feature columns in xslx export separated by lines


Karrp:
Client suggestion to add line separation to columns in reports xslx export

PR #439, opened 2019-07-17
Feature add lines separators in columns

Szymiks:
Resolves:https://github.com/Code-Poets/sheetstorm/issues/435
I also change horizontal alignment in hours field because it wasn't look good in printing mode.

### DIFF
--- a/employees/common/constants.py
+++ b/employees/common/constants.py
@@ -2,6 +2,8 @@ from datetime import timedelta
 from enum import Enum
 from typing import NamedTuple
 
+from openpyxl.styles import Side
+
 
 class MonthNavigationConstants(Enum):
     MAX_MONTH_VALUE = 12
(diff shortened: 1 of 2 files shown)


employees/common/constants.py
```diff
@@ -77,3 +79,4 @@ class ExcelGeneratorSettingsConstants(Enum):
     ZIP_CONTENT_TYPE_FORMAT = "application/zip"
     ZIP_EXPORTED_FILE_NAME = 'attachment; filename="{}_{}-reports.zip"'
     BORDER = "double"
```
dybi: this constant seems not to be used except for the line below; please inline it

Szymiks:
> If there are tests to be added, please make sure to add them ;)

I didn't see any test for settings like border or alignment but I think we could test it like creating some workbook and check for each cell value settings, I am not sure is it worth to do ?

PR #441, opened 2019-07-17
Chore add pipenv

kbeker:
There is no issue for that. I changed our requirements to pipenv. Please check if everything works on yours systems guys. Also I have cleaned packages needed for SheetStorm project because there were at least 5 packages which are no longer in use - that's why please check if everything is working on yours systems :)

If you want to install all requirements form pipenv you have to type `pipenv install`.

If you want to install also requirements for development purposes type `pipenv ins'tall --dev`

There is also message for @bartoszbetka : We need to change our deployment steps because now we won't install requirements from `requirements.lock` file. I will create issue and assign it to you.


kbeker:
> Was `black = "==19.3b0"` installed without any problems for you?

@rwrzesien I just had to type in statically and after this all works. But installing `black` without specified version does not work.

PR #443, opened 2019-07-18
Bugfix update requirements in pipenv

kbeker:
@Szymiks reported that current installation of pipenv does not work. I checked it and fixed up what I think might be wrong on debian. @Szymiks please check if this pipfile configuration works for you. If no - please contact me.


Szymiks:
success
Works fine ! :)

Issue #444, opened 2019-07-18
Inconsistent side-panel link highlights on report-related pages


maciejSamerdak:
When we navigate to `/reports/author/*/*/*/` or `/reports/management/*/`, `/reports/author/*/*/*/`, `/reports/author/*/*/*/`, `/users/*/` nothing is highlighted in the side-panel.

Should be done:
 ------------
 - Displaying each page should trigger proper highlight in the side-panel

PR #448, opened 2019-07-19
Bugfix set active actual url

Szymiks:
Resolve: https://github.com/Code-Poets/sheetstorm/issues/444

### DIFF
--- a/users/templates/base.html
+++ b/users/templates/base.html
@@ -24,9 +24,12 @@
 {% url 'custom-users-list' as custom_users_list_url %}
 {% url 'custom-projects-list' as custom_projects_list_url %}
 {% url 'custom-users-notifications' as custom_users_notifications %}
+{% url 'logout' as logout_url %}
 {% now "Y" as year_for_urls %}
 {% now "n" as month_for_urls %}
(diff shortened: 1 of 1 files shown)


users/templates/base.html
```diff
@@ -37,39 +38,39 @@
...
+                <a href="{{ home_url }}">
                     <img src="{% static 'users/images/grey/ico-02.png' %}">
                     <span>{% trans 'Home' %}</span>
                 </a>
                 {% endif %}
             </li>
             <li>
-                {% if request.path == account_url %}
-                <a class="active" href="{% url 'custom-user-update' %}">
+                {% if request.path == account_url or request.path|startswith:'/users/accounts/password_change/'%}
```
rwrzesien: Is `request.path|startswith:account_url` working? If yes please use this instead of hardcoded url that might change.

Szymiks: With `account_url` it didn't work but I have used `password_change url`

rwrzesien: @Szymiks I think you should then use those reversed urls instead of hardcoded ones, but I see you have left 3 of them. Would it be possible to fix them in separate pull request?

Issue #449, opened 2019-07-22
Duplicating task_activities while loading initial data


MartynaAnnaGottschling:
After deleting database and running `load_initial_data.sh`, I noticed that some of task_activities are duplicated (or even sometimes some mysterious and weird / random names). This is not a significant bug, but this phenomenon is annoying and hindering product demo presentations.

After small investigation I think the reason is loading ReportFactories without giving `task_activities` (because when I run `load_initial_data` command without previous `loaddata employees/fixtures/task_activities.yaml` only these duplicate tasks were in the database).

Should be done:
 ------------
 - add `task_activities` to every ReportFactory in `load_initial_data.py`

MartynaAnnaGottschling:
I just noticed adding `task_activities = factory.SubFactory("employees.factories.TaskActivityTypeFactory")` in ReportFactory. But I think this is still the cause of duplication.

PR #450, opened 2019-07-22
Bugfix add task activities to Report in initial data

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/449

When we create reports in initial data we didn't pass task activities, due to this `TaskActivityTypeFactory` was creating new task activity object from this values `"Review", "Backend Development", "Frontend Development", "Meeting"`

### DIFF
--- a/load_initial_data.sh
+++ b/load_initial_data.sh
@@ -1,2 +1,3 @@
+#!/usr/bin/env bash
 python manage.py loaddata employees/fixtures/task_activities.yaml
 python manage.py load_initial_data
--- a/users/management/commands/load_initial_data.py
+++ b/users/management/commands/load_initial_data.py
@@ -4,6 +4,7 @@ from django.db import transaction
 from django.utils import timezone
 
 from employees.factories import ReportFactory
+from employees.models import TaskActivityType
 from managers.factories import ProjectFactory
 from managers.models import Project
 from users.factories import UserFactory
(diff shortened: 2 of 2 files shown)


load_initial_data.sh
```diff
@@ -1,2 +1,3 @@
+#!/usr/bin/env bash
 python manage.py loaddata employees/fixtures/task_activities.yaml
 python manage.py load_initial_data
```
dybi: we might want to add a new line here ;)

users/management/commands/load_initial_data.py
```diff
@@ -127,6 +128,15 @@ def handle(self, *args, **options):
         self._add_member_to_project_if_not_added_yet(project_terminated, user_manager_1)
         self._add_member_to_project_if_not_added_yet(project_terminated, user_manager_2)
 
+        #
+        # Task activities
+        #
+
+        task_activity_review = TaskActivityType.objects.get(name="Review")
```
rwrzesien: If for some reason those objects would not exist (ie. someone will run `python manage.py flush` to clear database) this code will fail, so I think it is better to use `objects.get_or_create()` here and below.

rwrzesien:
Looks good, but please apply `get_or_create` pattern. It would make this completely safe in case of database flush. After adding it should be ready to merge.

PR #451, opened 2019-07-22
Change CustomUser's manager to enable migration making

maciejSamerdak:
No issue for this.

Fixes an issue where `makemigrations` call fails with an error, related to CustomUsers's manager class not being found.


users/migrations/0006_auto_20190722_1131.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('users', '0005_auto_20190703_1737'),
+    ]
+
+    operations = [
+        migrations.AlterModelManagers(
+            name='customuser',
+            managers=[
```
maciejSamerdak: I guess it's okay, since the manager is now created dynamically?

users/models.py
```diff
@@ -39,8 +39,6 @@ def get_with_prefetched_reports(self, reports: QuerySet) -> QuerySet:
 
 
 class CustomUserManager(BaseUserManager):
-    use_in_migrations = True
```
maciejSamerdak: The usage of this property is obscure. According to [Django documentation](https://docs.djangoproject.com/en/2.2/topics/migrations/#model-managers) it has to do with serialisation. @Szymiks reported seeing this on forums regarding REST framework. This line was created by @MartynaAnnaGottschling back in October 2018. `full-check` did not raise any errors after it got removed.

Is it therefore safe to assume that this is a leftover from the REST implementation and can be safely deleted?

users/migrations/0006_auto_20190722_1131.py
rwrzesien: Looks like thanks to this `CustomUser` managers will no longer be tracked in migrations, so it should be ok.

users/models.py
rwrzesien: Not related with REST exactly, but with possibility to use custom manager methods in migrations. As we don't use them, lets remove it.

Issue #374, opened 2019-07-01
Markdown support should be entirely removed from the app


maciejSamerdak:
Markdown was considered unnecessary for daily report filling. Formatting in description field should only consider newlines, whitespaces and hyperlinks. Hyperlinks should be blue (customer decision).
There also should be migration to cut out all markdown related syntax from existing reports.

Should be done:
 ------------
 - Markdown support should be removed
 - Description field should display newlines, whitespaces
 - Description field should display clickable **hyperlinks** in blue font (in .xlsx should be black)
 - Markdown syntax should be cut out from existing reports

dybi:
please, don't kill `markdown`!!!11oneoneone

PR #456, opened 2019-07-23
Chore remove markdown

dybi:
resolves: #374


rwrzesien:
@dybi Looks good, however the issue mentions data migration to remove markdown from current data (I am wondering if this is even possible without some super complex regex). Are you planning to do this or are you skipping this on purpose? Whatever you have chosen, please just drop a comment.

dybi:
@rwrzesien 
> Btw. what about clickable hyperlinks?

quite frankly I forgot about this ;) I will add them

When it comes to migration, I am wondering how much of it (markdown) is in the current data. I think it should be checked first and then the decision should be made whether it is worth to implement ;)

dybi:
Added hyperlinks in blue

rwrzesien:
> When it comes to migration, I am wondering how much of it (markdown) is in the current data. I think it should be checked first and then the decision should be made whether it is worth to implement ;)

I think it won't be possible to cover all cases (or the migration doing this would be overkill).

Btw. interesting solution for those links :+1:

dybi:
@kbeker @rwrzesien: migrations are solved in: https://github.com/Code-Poets/sheetstorm/pull/457

Issue #204, opened 2019-05-14
Project should have customisable task activity list


maciejSamerdak:
Should be done:
 ------------
 - Project should have it's own, customisable list of task activities
 - When creating or editing a report, the employee cannot select a task activity, that's outside of project's list
 - Project can have special task activities

rwrzesien:
My idea for that was:
1. Add nullable field `project` to `TaskActivityType` which will be FK to `Project`.
2. There should be a list of default `TaskActivityTypes` which `project` set to `null`.
3. When Manager/Admin creates a project, a copy of those defaults is created for project.
4. Then Manager/Admin can modify the `TaskActivityTypes` within related to project, without modifying defaults. He can CRUD them in any way without making changes to other projects too.

@maciejSamerdak @dybi What do you guys think?

Karrp:
It seems to me that it can cause more problems than benefits.

dybi:
@rwrzesien I would need to research a bit on that
@Karrp - like all new features, right? ;)

rwrzesien:
@dybi Sure, I would be happy to discuss the idea or its improvements, but this is just an idea, which still requires decision ;)
@Karrp Could you please be more precise - what problems exactly do you mean?

Karrp:
If the employee does some work which type of activity does not appear in the default list, the problem will appear.
I understand that then he must set `TaskActivityType` as `other` and contact the manager?
Maybe you have better solution.

I do not remember any more problems so I think we can implement this idea.

 ----
By the way, you will need to give the manager opportunity to add new activities in such a way that two managers do not add two activities with different names but significant ones.

kbeker:
@Karrp is this issue still valid? @maciejSamerdak is this a request from client?

Karrp:
Issue is valid.
It's nice to have this feature.

PR #98, opened 2019-03-13
Bug automatic user type

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/65

PR #460, opened 2019-07-29
Feature project should have customisable task activity list

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/204

**This PR will be change because @Karrp want this feature in differemt way.**
**Related**:https://github.com/Code-Poets/sheetstorm/pull/469

In this PR I just created few things:
**Should be done**
- create fields in `task_activite_models` - `projects - m2m` and `is_default` 
- refactor singals to signals.py in manager app because I had problems with import `TaskActivityType` in models in manager app and import `Project` in models in employee app.
- view when Admin and manager could manage project task activities or create new one.

**Will be done in next PR**
- add view for Admin to manage all task activities
- add dynamic filter during filling report

**Question**
I have never done something like this, so when I should make some script or something to set default task activity list for all existing projects?


Szymiks:
@Karrp Could you say what task activities should be default for all projects or maybe talk about it with Adrian ? :)

rwrzesien:
> I have never done something like this, so when I should make some script or something to set default task activity list for all existing projects?

For existing projects, in data migration.
For new ones, in function connected to `post_save` signal to `Project` model.

employees/models.py
```diff
@@ -15,8 +15,17 @@
...
 
+class TaskActivityTypeQuerySet(models.QuerySet):
+    def filter_default_task_activities(self) -> QuerySet:
+        return self.filter(is_default=True)
+
+
 class TaskActivityType(models.Model):
     name = models.CharField(max_length=TaskActivityTypeConstans.TASK_ACTIVITIES_MAX_LENGTH.value)
+    is_default = models.BooleanField(default=False)
+    projects = models.ManyToManyField(Project, related_name="project_activities")
```
rwrzesien: So how it is going to work now? One task activity type can be assigned to multiple projects? What if in one project manager will remove or rename it?

employees/models.py
```diff
@@ -15,8 +15,17 @@
 from users.models import CustomUser
 
 
+class TaskActivityTypeQuerySet(models.QuerySet):
+    def filter_default_task_activities(self) -> QuerySet:
```
rwrzesien: I think `get_defaults` should be enough.

managers/apps.py
```diff
@@ -3,3 +3,6 @@
 
 class ManagersConfig(AppConfig):
     name = "managers"
+
+    def ready(self) -> None:
+        import managers.signals  # noqa, flake8 F401 issue  # pylint: disable=unused-import
```
rwrzesien: Why do you need to import it this way?

managers/signals.py
```diff
@@ -0,0 +1,65 @@
...
+
+from employees.models import TaskActivityType
+from managers.models import Project
+from users.models import CustomUser
+
+logger = logging.getLogger(__name__)
+
+
+@receiver(m2m_changed, sender=Project.managers.through)
+def update_user_type(sender: Project, action: str, pk_set: Set, **kwargs: Any) -> None:
```
rwrzesien: Those are not exactly signals, but signals handlers. What was the case for them being in models, like it is usually done?

managers/signals.py
```diff
@@ -0,0 +1,65 @@
...
+            continue
+
+
+@receiver(post_save, sender=Project)
+def add_default_task_activities(sender: Project, **kwargs: Any) -> None:
+    assert sender == Project
+    if kwargs["created"]:
+        project = kwargs["instance"]
+        for task_activity in TaskActivityType.objects.filter_default_task_activities():
+            if task_activity not in project.project_activities.all():
```
rwrzesien: It definitely won't be there, as the object was just created.

managers/signals.py
```diff
@@ -0,0 +1,65 @@
...
+                task_activity.projects.add(project)
```
rwrzesien: You can do this like `project.project_activities.set(TaskActivityType.objects.filter_default_task_activities())`.

managers/templates/managers/project_task_activities.html
```diff
@@ -0,0 +1,74 @@
...
+                <form method="get" action=".">
+                    <select class="multiselect-task-activities" multiple="multiple" name="task_activities" size="6"
+                            title="{% trans 'If you want add more than one press ctrl' %}">
+                        {% for task in activities_not_connected %}
+                            <option value="{{ task.id }}">{{ task.name }}</option>
+                        {% endfor %}
+                    </select>
+                <p>
+                    <input type="submit" class="btn btn-primary" value="Add to project">
+                </p>
```
rwrzesien: Align! :)

managers/templates/managers/project_task_activities.html
```diff
@@ -0,0 +1,74 @@
...
+                        <th class="task-activities-column">{% trans 'Default' %}</th>
+                        <td class=Invisible></td>
+                    </tr>
+                    {% for task_activity in task_activities %}
+                        <tr>
+                            <td class="task-activities-column">{{ forloop.counter }}</td>
+                            <td class="task-activities-column">{{ task_activity.name }}</td>
+                            <td class="task-activities-column">{{ task_activity.is_default }}</td>
+                            <td class=Invisible align="center">
+                                <a href=".?task_activity={{ task_activity.id }}"
```
rwrzesien: Don't do this, use `{% url 'xxx' %}?task_activity={{ task_activity.id }}` instead of `.`.

managers/views.py
```diff
@@ -160,3 +165,47 @@ def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpRespons
...
+    form_class = TaskActivityForm
+
+    def _delete_task_activity_from_relation(self) -> None:
+        task_activity = get_object_or_404(TaskActivityType, pk=self.request.GET.get("task_activity"))
+        task_activity.projects.remove(self.object)
+
+    def _add_task_activities_to_relation(self, list_of_task_activities: list) -> None:
+        for task_activity_id in list_of_task_activities:
+            task_activity = get_object_or_404(TaskActivityType, pk=task_activity_id)
+            task_activity.projects.add(self.object)
```
rwrzesien: Same here, you can query for all `TaskActivityType` in single request, and set result using `set()`.

managers/views.py
```diff
@@ -160,3 +165,47 @@ def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpRespons
...
+    def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        self.object = self.get_object()  # pylint: disable=attribute-defined-outside-init
+        if "task_activities" in request.GET:
+            self._add_task_activities_to_relation(request.GET.getlist("task_activities"))
+        elif request.GET.get("task_activity"):
+            self._delete_task_activity_from_relation()
+        return super().get(request, args, kwargs)
+
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        self.object = self.get_object()  # pylint: disable=attribute-defined-outside-init
```
rwrzesien: Is this needed?

managers/views.py
```diff
@@ -160,3 +165,47 @@ def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpRespons
...
+    def form_valid(self, form: TaskActivityForm) -> HttpRequest:
+        task_activity = form.save()
+        task_activity.projects.add(self.get_context_data()["project"])
+        return self.render_to_response(self.get_context_data(form=form))
+
+    def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        self.object = self.get_object()  # pylint: disable=attribute-defined-outside-init
+        if "task_activities" in request.GET:
+            self._add_task_activities_to_relation(request.GET.getlist("task_activities"))
+        elif request.GET.get("task_activity"):
```
rwrzesien: Please create separate view for removing task activity.

managers/views.py
```diff
@@ -160,3 +165,47 @@ def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpRespons
...
+    def get_context_data(self, **kwargs: Any) -> dict:
+        context = super().get_context_data()
+        context["project"] = self.object
+        context["activities_not_connected"] = TaskActivityType.objects.exclude(projects__in=self.kwargs["pk"])
+        context["task_activities"] = self.object.project_activities.all().order_by("name")
+        return context
+
+    def form_valid(self, form: TaskActivityForm) -> HttpRequest:
+        task_activity = form.save()
+        task_activity.projects.add(self.get_context_data()["project"])
```
rwrzesien: Why not `self.object`? It doesn't make sense to get whole context just to get one object out of it.

managers/views.py
```diff
@@ -160,3 +165,47 @@ def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpRespons
...
+
+    def _add_task_activities_to_relation(self, list_of_task_activities: list) -> None:
+        for task_activity_id in list_of_task_activities:
+            task_activity = get_object_or_404(TaskActivityType, pk=task_activity_id)
+            task_activity.projects.add(self.object)
+
+    def get_context_data(self, **kwargs: Any) -> dict:
+        context = super().get_context_data()
+        context["project"] = self.object
+        context["activities_not_connected"] = TaskActivityType.objects.exclude(projects__in=self.kwargs["pk"])
```
rwrzesien: Why `__in`? You have only one item in `self.kwargs["pk"]`.
What is more, if `self.kwargs["pk"]` will be string ie. `"123"`it would exclude objects with id 1, 2 and 3.

Karrp:
> @Karrp Could you say what task activities should be default for all projects or maybe talk about it with Adrian ? :)

@Szymiks 
If you create new project task activities should be None as default. Consequently when creating report employee should be able to chose every possible activity from list.

managers/tests/test_views.py
```diff
@@ -225,3 +229,89 @@ def test_delete_project_function_view_should_not_delete_project_on_non_admin_req
...
+        task_activity_name = "New activity"
+        response = self.client.post(self.url2, data={"name": task_activity_name})
+
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, task_activity_name)
+        task_activity = TaskActivityType.objects.get(name=task_activity_name)
+        self.assertIn(task_activity, self.project.project_activities.all())
+
+    def test_manage_task_activities_view_should_add_existing_task_activity_to_project_on_add_to_project_function(self):
+        length_of_default_task_activities = TaskActivityType.objects.filter_default_task_activities().count()
```
kbeker: Maybe better name for this would be `number_of_all_default_task_activities`?

managers/signals.py
```diff
@@ -0,0 +1,65 @@
...
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from employees.models import TaskActivityType
+from managers.models import Project
+from users.models import CustomUser
+
+logger = logging.getLogger(__name__)
+
+
```
dybi: shouldn't we prevent signals from being duplicated? https://docs.djangoproject.com/en/2.2/topics/signals/#preventing-duplicate-signals
@rwrzesien, what would be your suggestions here?

managers/tests/test_views.py
```diff
@@ -225,3 +229,89 @@ def test_delete_project_function_view_should_not_delete_project_on_non_admin_req
...
+        self.assertEqual(self.project.project_activities.all().count(), length_of_default_task_activities)
+
+        response = self.client.get(self.url2 + f"?task_activities={self.not_default_task_activity.pk}")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, self.not_default_task_activity.name)
+        self.assertIn(self.not_default_task_activity, self.project.project_activities.all())
+        self.assertNotEqual(self.project.project_activities.all().count(), length_of_default_task_activities)
```
kbeker: I always prefer and recommend to checking if something is equal to something instead of not equal if you have opportunity to this, because 1 won't be equal to 2, but also won't be equal to -42109387123812.

managers/tests/test_views.py
```diff
@@ -225,3 +229,89 @@ def test_delete_project_function_view_should_not_delete_project_on_non_admin_req
...
+        length_of_project_task_activities_before_remove = self.project.project_activities.all().count()
+        self.assertIn(self.default1_task_activity, self.project.project_activities.all())
+
+        response = self.client.get(self.url2 + f"?task_activity={self.default1_task_activity.pk}")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertNotIn(self.default1_task_activity, self.project.project_activities.all())
+        self.assertNotEqual(
+            self.project.project_activities.all().count(), length_of_project_task_activities_before_remove
+        )
```
kbeker: Same here

managers/tests/test_views.py
```diff
@@ -225,3 +229,89 @@ def test_delete_project_function_view_should_not_delete_project_on_non_admin_req
...
         self.assertEqual(response.status_code, 302)
         self.assertEqual(Project.objects.all().count(), 1)
+
+
+class ManageTaskActivitiesInProjectView(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.user_admin = AdminUserFactory()
+        self.client.force_login(self.user_admin)
+        self.default1_task_activity = TaskActivityTypeFactory(is_default=True)
```
kbeker: We use snake case style of naming variables so instead of `default1....` please change it to `default_task_activity_1`

managers/tests/test_views.py
```diff
@@ -225,3 +229,89 @@ def test_delete_project_function_view_should_not_delete_project_on_non_admin_req
...
+        self.default2_task_activity = TaskActivityTypeFactory(is_default=True)
```
kbeker: Same here

managers/tests/test_views.py
```diff
@@ -225,3 +229,89 @@ def test_delete_project_function_view_should_not_delete_project_on_non_admin_req
         response = self.client.post(self.url)
         self.assertEqual(response.status_code, 302)
         self.assertEqual(Project.objects.all().count(), 1)
+
+
+class ManageTaskActivitiesInProjectView(TestCase):
```
kbeker: Did you check if our scripts for running test starts this? You missed in this name "Test" and also you give to test class the same name as view name. Please change it :)

managers/views.py
```diff
@@ -160,3 +165,47 @@ def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpRespons
...
+        context["task_activities"] = self.object.project_activities.all().order_by("name")
+        return context
+
+    def form_valid(self, form: TaskActivityForm) -> HttpRequest:
+        task_activity = form.save()
+        task_activity.projects.add(self.get_context_data()["project"])
+        return self.render_to_response(self.get_context_data(form=form))
+
+    def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        self.object = self.get_object()  # pylint: disable=attribute-defined-outside-init
```
kbeker: In `get` method `object` variable is not already initiated?

managers/signals.py
```diff
@@ -0,0 +1,65 @@
...
+@receiver(m2m_changed, sender=Project.managers.through)
+def update_user_type(sender: Project, action: str, pk_set: Set, **kwargs: Any) -> None:
+    assert sender == Project.managers.through
+    project = kwargs["instance"]
+    logger.debug(f"Updates on project with id: {sender.pk} for users with id: {pk_set}")
+    if action in ["pre_remove", "post_remove"]:
```
dybi: shouldn't one action (i.e. "post_remove") suffice here? Isn't signal received twice per same event?

managers/signals.py
```diff
@@ -0,0 +1,65 @@
...
+        change_user_type_to_employee(pk_set)
+    elif action in ["pre_add", "post_add"]:
```
dybi: shouldn't one action (i.e. "post_add") suffice here? Isn't signal received twice per same event?

managers/signals.py
```diff
@@ -0,0 +1,65 @@
...
+        change_user_type_to_manager(project)
+    else:
```
dybi: this `else` seems unnecessary

Karrp:
> Hmm it is quite edge case but we still need to handle this. What to do with reports which have activity which can be deleted by manager from list of project_activities?

We can do 2 things i think.
1. Not delete just inactive (not to choose in new reports -> edition? ) - more changes, need discussion
2. Delete and change to `Other` - easier but changes history

@kbeker what you think?

managers/signals.py
```diff
@@ -0,0 +1,65 @@
...
+        logger.debug(f"User with user type: {manager.user_type} with id: {manager.pk} in managers")
+        if manager.user_type == CustomUser.UserType.EMPLOYEE.name:
+            logger.debug(
+                f"User with user type: {manager.user_type} and id: {manager.pk} is in managers for project with id: {project.pk}"
+            )
+            manager.user_type = CustomUser.UserType.MANAGER.name
+            manager.full_clean()
+            manager.save()
+            logger.debug(f"User with id: {manager.pk} have changed user type to manager")
+        else:
```
dybi: this `else` seems unnecessary

managers/signals.py
```diff
@@ -0,0 +1,65 @@
...
+        user = CustomUser.objects.get(pk=user_id)
+        if not user.manager_projects.exists() and user.user_type != CustomUser.UserType.ADMIN.name:
+            logger.info(f"User: {user} user type change to employee")
+            user.user_type = CustomUser.UserType.EMPLOYEE.name
+            user.full_clean()
+            user.save()
+            logger.info(f"User with id: {user.pk} have changed user type to employee")
+        else:
+            logger.debug(f"User with id: {user.pk} has not changed to employee")
+            continue
```
dybi: this `continue` seems unnecessary

employees/models.py
Szymiks: @rwrzesien Yes, as you said one task activity could be assigned to multiple projects, what is more, in my case I thought to give permission to change name of task activity only for admin. If PM needs to create some special task He could to this in manage view and it will be automaticly added to project. @kbeker Maybe in this situations we could change it to `other`?

managers/apps.py
Szymiks: I couldn't import TaskActivityType in manager app in models so I decided to refactor whole handlers to file `signals.py`. On this page, It is written that we have to do this because if I wouldn't my handlers will not catch signals.
https://simpleisbetterthancomplex.com/tutorial/2016/07/28/how-to-create-django-signals.html

managers/signals.py
Szymiks: @rwrzesien Obsługiwały sytuację, gdy ktoś tworzył projekt i nadawał użytkownikowi `user_type=Manager` lub zabierał prawa jako menedzer jezeli nie mial innych projektów.

managers/tests/test_views.py
```diff
@@ -225,3 +229,89 @@ def test_delete_project_function_view_should_not_delete_project_on_non_admin_req
...
+    def setUp(self):
+        super().setUp()
+        self.user_admin = AdminUserFactory()
+        self.client.force_login(self.user_admin)
+        self.default1_task_activity = TaskActivityTypeFactory(is_default=True)
+        self.default2_task_activity = TaskActivityTypeFactory(is_default=True)
+        self.not_default_task_activity = TaskActivityTypeFactory()
+
+        self.project = ProjectFactory()
+        self.url2 = reverse("project-task-activities", kwargs={"pk": self.project.pk})
```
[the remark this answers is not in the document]
Szymiks: After that bug what I told you I was trying to do everything, and finally I forgot to chagne it ... Sory for that .. :bug:  ....

employees/models.py
Szymiks: Done

employees/fixtures/task_activities.yaml
```diff
@@ -2,6 +2,7 @@
   pk: 1
   fields:
     name: Other
+    is_default: True
```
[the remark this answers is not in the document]
Szymiks: Sure, I will wait for last decission it is only for tests.

managers/signals.py
Szymiks: I will wait for @rwrzesien opinion, because I am not sure what I should have to do now.

managers/signals.py
Szymiks: I created it because once @kbeker  suggest it to me:
https://github.com/Code-Poets/sheetstorm/pull/98#discussion_r270030292

managers/signals.py
Szymiks: I created it because once @kbeker  suggest it to me:
https://github.com/Code-Poets/sheetstorm/pull/98#discussion_r270030292

managers/signals.py
Szymiks: I created it because once @kbeker  suggest it to me:
https://github.com/Code-Poets/sheetstorm/pull/98#discussion_r270030292

managers/signals.py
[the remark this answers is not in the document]
Szymiks: I have choosen @rwrzesien version because`forloop` is not needed there.

managers/views.py
Szymiks: Yes, FormView doesn't have `self.object` so I inherit from `SingleObjectMixin` and use all methods from there to work with object and queryset. What is more it give protection to this view only for manager of current project, because without `SingleObjectMixin` this mixin `IsManagerOfCurrentProjectMixin` doesn't work.

managers/views.py
Szymiks: Yes, FormView doesn't have self.object so I inherit from SingleObjectMixin and use all methods from there to work with object and queryset. What is more it give protection to this view only for manager of current project, because without SingleObjectMixin this mixin IsManagerOfCurrentProjectMixin doesn't work.

managers/views.py
Szymiks: What is more when I wanted to use `self.object` during post I have to init this variable because then in `form_valid` I don't have it .

managers/tests/test_views.py
Szymiks: Yes our scripts run this test too, however I wil change it ASAP... sory :(

managers/tests/test_views.py
Szymiks: Yes, I that it is not precission enough, but I wanted to check if task activity list is different when I add new task than before when it contain only default tasks, however I will change it ASAP

managers/signals.py
Szymiks: I use this name to file because It was explained like that here:
https://simpleisbetterthancomplex.com/tutorial/2016/07/28/how-to-create-django-signals.html

managers/apps.py
rwrzesien: What was the exact problem with imports?

managers/signals.py
rwrzesien: The whole function can be limited to single query 
```
project.managers.filter(
    user_type=CustomUser.UserType.EMPLOYEE.name
).update(
   user_type=CustomUser.UserType.MANAGER.name
)
```

managers/signals.py
rwrzesien: Same here:
```
CustomUser.objects.filter(
    pk__in=pk_set
).exclude(
   user_type=CustomUser.UserType.ADMIN.name,
   projects=project,  # I am not sure about that one but should be reachable in similar way, maybe using annotate + count + filter on it
).update(
   user_type=CustomUser.UserType.EMPLOYEE.name
)
```

employees/models.py
rwrzesien: @Szymiks OK makes sense to limit access only for admin, but what if there will be the same type used in two projects, and in one manager would like to rename it? Even if he create a new one, old reports will still be using old one.

managers/signals.py
rwrzesien: I get the point, but personally I think placing them under the models works fine. I have never experienced any import issues like they mention in this article. I think it is readable when you have signals right under the model definition, as it is actually a part of model "logic".
But if you decide to move them away to separate file it is OK for me too. But honestly in `signals.py` I would rather expect custom signals objects (https://docs.djangoproject.com/en/2.2/topics/signals/#defining-signals).

managers/templates/managers/project_task_activities.html
```diff
@@ -0,0 +1,75 @@
...
+                <form id="delete-task-from-relation" method="post" action="{% url 'remove-task-activity-from-project' pk=project.pk %}">
+                    {% csrf_token %}
+                </form>
+                    {% for task_activity in task_activities %}
+                        <tr>
+                            <td class="task-activities-column">{{ forloop.counter }}</td>
+                            <td class="task-activities-column">{{ task_activity.name }}</td>
+                            <td class="task-activities-column">{{ task_activity.is_default }}</td>
+                            <td class=Invisible align="center">
+                                <button form="delete-task-from-relation" type="submit" value="{{ task_activity.pk }}" name="task_activity_pk" class="btn btn-danger glyphicon glyphicon-remove" title="{% trans 'Remove from project' %}"></button></button>
```
rwrzesien: Will submit outside of the <form> work?

managers/templates/managers/project_task_activities.html
```diff
@@ -0,0 +1,75 @@
...
+        <div class="container">
+            <div class="table-responsive">
+                <table class="table">
+                    <tr>
+                        <th class="task-activities-column">{% trans 'Index' %}</th>
+                        <th class="task-activities-column">{% trans 'Name' %}</th>
+                        <th class="task-activities-column">{% trans 'Default' %}</th>
+                        <td class=Invisible></td>
+                    </tr>
+                <form id="delete-task-from-relation" method="post" action="{% url 'remove-task-activity-from-project' pk=project.pk %}">
```
rwrzesien: This is not correct html syntax. You can't put stuff like that inside table tags. Yes, it will work, but this is not right way to do it. I think it would be better to put into the <td> in which you currently have submit button.

managers/views.py
rwrzesien: What if you combine `FormView` and `DetailView` instead `SingleObjectMixin`?

managers/tests/test_views.py
rwrzesien: It works because it inherits from `TestCase`.

managers/views.py
rwrzesien: I suggest trying to combine `FormView` and `DetailView`.

managers/signals.py
rwrzesien: I think we won't have this problem here because we don't connect it manually, but is nice that you have thought about this case.

employees/models.py
Szymiks: @rwrzesien I understand, but how often can such a case happen? If there is a problem with the name for everyone, all you need to do is change the name of activity  by admin, but if only one manager need to change this he could create new one and use it. It seems to me that this problem will only occur after the other person's feedback, so will old reports with old activity be did it matter? However this is only my small opinion, I think that someone decisional should give his point of view.

managers/apps.py
Szymiks: I hade this exception 
```
Traceback (most recent call last):
  File "manage.py", line 19, in <module>
    execute_from_command_line(sys.argv)
  File "/home/szymi/virtualenv/sheetstorm-SgY9UmI-/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/home/szymi/virtualenv/sheetstorm-SgY9UmI-/lib/python3.7/site-packages/django/core/management/__init__.py", line 357, in execute
    django.setup()
  File "/home/szymi/virtualenv/sheetstorm-SgY9UmI-/lib/python3.7/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/home/szymi/virtualenv/sheetstorm-SgY9UmI-/lib/python3.7/site-packages/django/apps/registry.py", line 114, in populate
    app_config.import_models()
  File "/home/szymi/virtualenv/sheetstorm-SgY9UmI-/lib/python3.7/site-packages/django/apps/config.py", line 211, in import_models
    self.models_module = import_module(models_module_name)
  File "/home/szymi/virtualenv/sheetstorm-SgY9UmI-/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/szymi/Dokumenty/GitRepos/sheetstorm/sheetstorm/managers/models.py", line 9, in <module>
    from employees.models import TaskActivityType
  File "/home/szymi/Dokumenty/GitRepos/sheetstorm/sheetstorm/employees/models.py", line 14, in <module>
ImportError: cannot import name 'Project' from 'managers.models' (/home/szymi/Dokumenty/GitRepos/sheetstorm/sheetstorm/managers/models.py)
```
I think it happen because We imported Project in `employees app` and in `managers app` models we want to import `TaskActivityType` which is needed to work with handlers.

managers/signals.py
Szymiks: Sure I will make research about the problem with this import but for one I don't know how I could get round this problem

managers/views.py
Szymiks: I think that it won't work, I remeber that once I have tried this but It didn't work.
I found some info how it could be resolved.
https://stackoverflow.com/questions/16931901/django-combine-detailview-and-formview
and her in documentation:
https://docs.djangoproject.com/en/1.8/topics/class-based-views/mixins/#using-formmixin-with-detailview
I is quite similar solution as mine but there is written to use `DetailView` with `FormMixin` and override `post` If you think this will be better I will change it ASAP :)

managers/templates/managers/project_task_activities.html
Szymiks: I wasn't surs if this is okay. So our html should generate many forms? If we have 17 task activities we will have 17th forms ? Am I understand good?

managers/templates/managers/project_task_activities.html
Szymiks: Sure :P

managers/views.py
Szymiks: As uppon I said that combine two generic view are very though or maybe even not possible. We should use mixins to get methods we want or make two views. 
https://docs.djangoproject.com/en/1.8/topics/class-based-views/mixins/#using-formmixin-with-detailview

managers/views.py
Szymiks: Sure I will do this

managers/signals.py
dybi: I would insist on removing it - it's redundant. @kbeker what would be the advantages of having it here?

managers/tests/test_views.py
```diff
@@ -265,53 +265,84 @@ def test_manage_task_activities_in_project_view_should_redirect_on_employee_get(
...
-        self.assertContains(response, task_activity_name)
+        self.assertEqual(response.status_code, 302)
         task_activity = TaskActivityType.objects.get(name=task_activity_name)
         self.assertIn(task_activity, self.project.project_activities.all())
 
     def test_manage_task_activities_view_should_add_existing_task_activity_to_project_on_add_to_project_function(self):
-        length_of_default_task_activities = TaskActivityType.objects.filter_default_task_activities().count()
-        self.assertEqual(self.project.project_activities.all().count(), length_of_default_task_activities)
+        number_of_all_task_activities_before_add = self.project.project_activities.all().count()
+        self.assertEqual(self.project.project_activities.all().count(), number_of_all_task_activities_before_add)
```
dybi: ?

managers/tests/test_views.py
```diff
@@ -265,53 +265,84 @@ def test_manage_task_activities_in_project_view_should_redirect_on_employee_get(
...
 
-        response = self.client.get(self.url2 + f"?task_activities={self.not_default_task_activity.pk}")
+        response = self.client.post(self.url, data={"task_activities": self.not_default_task_activity.pk})
+        number_of_all_task_activities_after_add_one_new_task_activity = self.project.project_activities.all().count()
+        new_task_activity = 1
```
dybi: num_of_added_task_activities?

managers/tests/test_views.py
Szymiks: :broken_heart:

managers/signals.py
kbeker: @dybi I don't remember exactly when but we discussed during Concent project how to handle cases where if-elif statemets are and we decided that we should always end statements like this with `else-return` statement. But now when I'm staring and this piece of code I think that this might be deleted I think that I made mistake about case where we should use `else-return` statement and this is sure not good place for it. So @Szymiks as @dybi suggested - please delete it.

managers/apps.py
Szymiks: @rwrzesien When I make import locally it works fine

employees/models.py
rwrzesien: I agree that it not for us to decide on requirements, I just wanted to point out the problem that could be caused with current implementation. But if it is agreed that it will be resolved in the way which you have described, it is ok for me.

managers/apps.py
rwrzesien: I see that you have resolved this by making import directly in function, so it works now right?

managers/models.py
```diff
@@ -57,38 +58,29 @@ def update_user_type(sender: Project, action: str, pk_set: Set, **kwargs: Any) -
...
+        Q(user_type=CustomUser.UserType.ADMIN.name) | ~Q(manager_projects=None)
+    ).update(user_type=CustomUser.UserType.EMPLOYEE.name)
+
+
+@receiver(post_save, sender=Project)
+def add_default_task_activities(sender: Project, **kwargs: Any) -> None:
+    assert sender == Project
+    if kwargs["created"]:
+        project = kwargs["instance"]
+        from employees.models import TaskActivityType
```
rwrzesien: Please place those types of imports at the beginning of the function.

managers/signals.py
rwrzesien: I think it is resolved now?

managers/views.py
rwrzesien: So is this still needed after that?

managers/models.py
```diff
@@ -57,38 +58,29 @@ def update_user_type(sender: Project, action: str, pk_set: Set, **kwargs: Any) -
...
-            user.user_type = CustomUser.UserType.EMPLOYEE.name
-            user.full_clean()
-            user.save()
-            logger.info(f"User with id: {user.pk} have changed user type to employee")
-        else:
-            logger.debug(f"User with id: {user.pk} has not changed to employee")
-            continue
+    CustomUser.objects.filter(pk__in=pk_set).exclude(
+        Q(user_type=CustomUser.UserType.ADMIN.name) | ~Q(manager_projects=None)
+    ).update(user_type=CustomUser.UserType.EMPLOYEE.name)
```
rwrzesien: Nice solution :+1:

managers/templates/managers/project_task_activities.html
rwrzesien: Yes, and then you could use `task_activity.pk` directly in url instead of passing it in body and simplify `RemoveTaskActivityFromProjectView` a little bit (`if` will be no needed).

rwrzesien:
@Szymiks Almost there, I think the one important thing left is passing `task_activity_id` in url and make proper html in the table.

managers/apps.py
Szymiks: @rwrzesien , Yes it works but pylint gives one warning about `cycling import`

managers/tests/test_views.py
```diff
@@ -225,3 +229,119 @@ def test_delete_project_function_view_should_not_delete_project_on_non_admin_req
...
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, self.default_task_activity_1.name)
+        self.assertContains(response, self.not_default_task_activity.name)
+
+    def test_manage_task_activities_view_should_add_new_task_activity_on_post_and_add_it_to_project(self):
+        task_activity_name = "New activity"
+        response = self.client.post(self.url, data={"name": task_activity_name})
+
+        self.assertEqual(response.status_code, 302)
+        task_activity = TaskActivityType.objects.get(name=task_activity_name)
```
kbeker: This is used only one time. You can check it directly in assert: 
```python
self.assertIn(TaskActivityType.objects.get(name=task_activity_name), self.project.project_activities.all())
```

managers/views.py
Szymiks: Yes it is, because I am using `self.object` here -> `_add_task_activities_to_relation`

managers/templates/managers/project_task_activities.html
Szymiks: Done

PR #461, opened 2019-07-30
Chore install python packages by using pipenv instead of pip

### LINKED ISSUE #442: Change requirements installation step in deployment process
Our requirements changed and now instead of installing requirements by `pip` and installing them from `requirements.lock` file we need to have installed `pipenv` on server. Please install and test newest version of this app. After correct installation please install all needed requirements by typing `pipenv install`. 
Step to reproduce installing requirements on system:

1. Create virtualenv
2. Activate virtualenv
3. Install requirements using `pipenv install`.

Different steps might create you virtualenv with strange name and make it impossible to use correctly virtualenv with packages installed by `pipenv`

bartoszbetka:
Resolves #442

### DIFF
--- a/sheetstorm-deployment/roles/install_dependencies_and_basic_configuration/tasks/main.yml
+++ b/sheetstorm-deployment/roles/install_dependencies_and_basic_configuration/tasks/main.yml
@@ -31,8 +31,6 @@
       apt:
         name:
           - git
-          - virtualenv
-          - python3-setuptools
           - wget
 
     - name:  Install dependencies from testing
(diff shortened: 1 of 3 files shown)


sheetstorm-deployment/roles/install_dependencies_and_basic_configuration/tasks/main.yml
```diff
@@ -52,6 +50,9 @@
           - python3-distutils
           - python3-psycopg2
           - python3-certbot-nginx
+          - virtualenv
+          - python3-setuptools
```
kbeker: Andrew, please sort it alphabetically :)

Issue #454, opened 2019-07-23
After enter wrong password during login email shouldn't disappear.


Karrp:
Should be done:
 ------------
 - Somehow don't erase email address

maciejSamerdak:
And how about the failed sign**up**? You also have to refill all the data afterwards.

PR #467, opened 2019-07-31
Bugfix modify template to preserve user's email in case of errors

dybi:
resolves: https://github.com/Code-Poets/sheetstorm/issues/454


dybi:
with special thanks for @rwrzesien ;)

PR #469, opened 2019-08-01
Feature add dynamic filter for task activities per project

Szymiks:
Related: https://github.com/Code-Poets/sheetstorm/pull/460

On this branch I added dynamic filter for task activity list per each project.
After approving this PR: https://github.com/Code-Poets/sheetstorm/pull/460 I will start to create a panel admin view for task activities.

After `full_check` you can notice two failed test but It has been reported because it occured on master branch.

### DIFF
--- a/employees/forms.py
+++ b/employees/forms.py
@@ -18,6 +18,7 @@ from employees.common.constants import MonthNavigationConstants
 from employees.models import Report
 from employees.models import TaskActivityType
 from managers.models import Project
+from users.models import CustomUser
 
 
 class ProjectJoinForm(forms.Form):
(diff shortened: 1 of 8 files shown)


employees/forms.py
```diff
@@ -59,18 +60,31 @@ class ReportForm(forms.ModelForm):
...
         super(ReportForm, self).__init__(*args, **kwargs)
         author = kwargs["initial"]["author"]
         self.fields["project"].queryset = author.get_project_ordered_by_last_report_creation_date()
+
+        self._filter_task_activities_per_project(author)
+        self._set_last_choices_in_report_form(author)
+
+    def _filter_task_activities_per_project(self, author: CustomUser) -> None:
+        if 'project' in self.data:
+            project_id = int(self.data.get('project'))
```
Szymiks: In next commit there is added `try..except` block because if user would put empty string we could get `ValueException`, I made it `pass` because form after this will raise ValidationException and it will be shown to user.

employees/views.py
```diff
@@ -569,3 +572,16 @@ def render_to_response(self, context: dict, **response_kwargs: Any) -> HttpRespo
...
+
+@method_decorator(login_required, name="dispatch")
+class LoadTaskActivitiesInProjectView(TemplateView):
+    template_name = "employees/partial/task_activity_list.html"
+
+    def get_context_data(self, **kwargs: Any) -> dict:
+        context = super().get_context_data(**kwargs)
+        if "project" not in self.request.GET:
+            raise Http404
+        context["project_id"] = self.request.GET.get("project")
```
rwrzesien: You don't need this to be passed in context. Also you should validate the value passed from GET, at least if it is an int.

employees/forms.py
```diff
@@ -59,18 +60,36 @@ class ReportForm(forms.ModelForm):
...
+        self._set_last_choices_in_report_form(author)
+
+    def _filter_task_activities_per_project(self) -> None:
+        if "project" in self.data:
+            try:
+                project_id = int(self.data.get("project"))
+                self.fields["task_activities"].queryset = TaskActivityType.objects.filter(projects=project_id).order_by(
+                    "name"
+                )
+            except (ValueError, TypeError):
```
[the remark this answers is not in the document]
Szymiks: @dybi, when someone will pass to `self.data` something lik `{"project": "abcd" + 2}`

Issue #462, opened 2019-07-31
Problems with managers field in project creation and update


Szymiks:
During project creation user who requested this is assigne to managers automatically and isn't displayed as a choice in managers field, however this field is stil required and can't be empty.

When we want to update project created above, user is not displayed in managers field and when we press update button this user is removed from managers from this project.

Should be done:
 ------------
 -  managers field should be optional or display all managers in managers field
 - show all managers in managers field in project update view

PR #472, opened 2019-08-01
Bugfix managers field in project admin form

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/462

**Should have done**
- managers field is optional, user of request is always set as manager.
- when we go to update the project view, we could see all managers and if we want to update it, we do not remove the manager who created the project.

### DIFF
--- a/managers/forms.py
+++ b/managers/forms.py
@@ -1,7 +1,4 @@
 from typing import Any
-from typing import Dict
-from typing import List
-from typing import Optional
 
 from bootstrap_datepicker_plus import DatePickerInput
 from django import forms
@@ -12,19 +9,17 @@ from managers.models import Project
 from users.models import CustomUser
 
 
-class ManagerSelectMultiple(Select2MultipleWidget):
-    def optgroups(self, name: str, value: List, attrs: Optional[Dict] = None) -> List:
-        self.choices.queryset = CustomUser.objects.exclude(user_type=CustomUser.UserType.EMPLOYEE.name).exclude(
-            pk=self.user_pk if hasattr(self, "user_pk") else None
-        )
-        return super().optgroups(name, value, attrs)
-
-
 class ProjectAdminForm(forms.ModelForm):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
+        user_pk = kwargs.pop("user_pk", None)
         super().__init__(*args, **kwargs)
-        if hasattr(self, "user_pk"):
-            ManagerSelectMultiple.user_pk = self.user_pk
+
+        managers_to_choose = CustomUser.objects.exclude(user_type=CustomUser.UserType.EMPLOYEE.name)
+        if self.instance.pk:
+            self.fields["managers"].queryset = managers_to_choose
+        else:
+            self.fields["managers"].queryset = managers_to_choose.exclude(pk=user_pk)
+            self.fields["managers"].required = False
 
     class Meta:
         model = Project
@@ -32,14 +27,13 @@ class ProjectAdminForm(forms.ModelForm):
         widgets = {
             "start_date": DatePickerInput(options={"format": CORRECT_DATE_FORMAT}),
             "stop_date": DatePickerInput(options={"format": CORRECT_DATE_FORMAT}),
-            "managers": ManagerSelectMultiple(),
+            "managers": Select2MultipleWidget(),
             "members": Select2MultipleWidget(),
         }
 
 
-class ProjectManagerForm(ProjectAdminForm):
+class ProjectManagerForm(forms.ModelForm):
     class Meta:
-        model = ProjectAdminForm.Meta.model
-        fields = ProjectAdminForm.Meta.fields
+        model = Project
         exclude = ("managers",)
         widgets = ProjectAdminForm.Meta.widgets
--- a/managers/tests/test_views.py
+++ b/managers/tests/test_views.py
@@ -9,6 +9,7 @@ from managers.views import ProjectCreateView
 from managers.views import ProjectDetailView
 from managers.views import ProjectsListView
 from managers.views import ProjectUpdateView
+from users.factories import ManagerUserFactory
 from users.factories import UserFactory
 from users.models import CustomUser
 
@@ -137,7 +138,6 @@ class ProjectCreateViewTests(ProjectBaseTests):
             "name": "Another Example Project",
             "start_date": timezone.now().date() - timezone.timedelta(days=30),
             "suspended": False,
-            "managers": [self.user.pk],
             "members": [self.user.pk],
         }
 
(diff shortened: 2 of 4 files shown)


managers/views.py
```diff
@@ -92,9 +92,12 @@ class ProjectCreateView(CreateView):
     model = Project
     template_name = "managers/project_form.html"
 
+    def get_form(self, form_class: ProjectAdminForm = None) -> ProjectAdminForm:
+        ProjectAdminForm.user_pk = self.request.user.pk
```
rwrzesien: This should be passed in `get_form_kwargs` method and handled in form `__init__`.

managers/forms.py
```diff
@@ -12,27 +9,27 @@
...
-        return super().optgroups(name, value, attrs)
-
-
 class ProjectAdminForm(forms.ModelForm):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
-        if hasattr(self, "user_pk"):
-            ManagerSelectMultiple.user_pk = self.user_pk
+
+        if "managers" in self.fields:
```
rwrzesien: This check doesn't make much sense, as this field is in form by its definition.

rwrzesien:
> user of request is always set as manager.

Is this implemented?

managers/forms.py
Szymiks: But manager is able to update project, too. However they don't have rights to update managers in project and  then we are using for them`ProjectManagerForm` which inheritance from `ProjectAdminForm` but `manager` field is excluded in `ProjectManagerForm`

Szymiks:
> > user of request is always set as manager.
> 
> Is this implemented?
It is implement in create view by someone else before that PR.
```
    def form_valid(self, form: ProjectAdminForm) -> HttpRequest:
        if self.request.user not in form.cleaned_data["managers"]:
            assert not self.request.user.is_employee
            managers_pk_list = [manager.pk for manager in form.cleaned_data["managers"]]
            managers_pk_list.append(self.request.user.pk)
            managers_queryset = CustomUser.objects.filter(pk__in=managers_pk_list)
            form.cleaned_data["managers"] = managers_queryset
        project = form.save()
        logger.info(f"New project with id: {project.pk} has been created")
        return super(ModelFormMixin, self).form_valid(form)  # pylint: disable=bad-super-call
```

> > when we go to update the project view, we could see all managers and if we want to update it, we do not remove the manager who created the project.
> 
> I don't see this too.
```
        if "managers" in self.fields:
            managers_to_choose = CustomUser.objects.exclude(user_type=CustomUser.UserType.EMPLOYEE.name)
            if self.instance.pk:
                self.fields["managers"].queryset = managers_to_choose
```
It is here, before we was excluding all the time `user_pk`

managers/views.py
Szymiks: Not sure about my solution but I think i resolve it

managers/forms.py
```diff
@@ -12,27 +9,26 @@
...
+        user_pk = kwargs.pop("user_pk") if "user_pk" in kwargs else None
         super().__init__(*args, **kwargs)
-        if hasattr(self, "user_pk"):
-            ManagerSelectMultiple.user_pk = self.user_pk
+
+        if "managers" in self.fields:
+            managers_to_choose = CustomUser.objects.exclude(user_type=CustomUser.UserType.EMPLOYEE.name)
+            if self.instance.pk:
+                self.fields["managers"].queryset = managers_to_choose
+            else:
```
dybi: could you please explain `else`?

managers/views.py
```diff
@@ -92,9 +92,13 @@ class ProjectCreateView(CreateView):
     model = Project
     template_name = "managers/project_form.html"
 
+    def get_form_kwargs(self) -> dict:
+        kwargs = super(ProjectCreateView, self).get_form_kwargs()
```
dybi: could be simpler: 
```suggestion
        kwargs = super().get_form_kwargs()
```

managers/tests/test_views.py
```diff
@@ -205,6 +209,18 @@ def test_project_update_view_should_not_update_managers_if_user_is_manager(self)
         self.project.refresh_from_db()
         self.assertEqual(self.project.managers.count(), 0)
 
+    def test_project_update_view_should_display_all_connected_managers_on_get(self):
```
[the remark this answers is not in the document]
Szymiks: It means that in managers field we have all managers of current project.

managers/forms.py
Szymiks: Sure.
if instance.pk exist is mean that we user want to update project so we would like to show all available managers in current project  and all managers in db to choose. Furthermore, now this field is required because we must be sure that user of request will not delete all managers from project.
`else` mean that we are in `CreateProjectView` so as it said in issue we wanted to assigne user of request to managers and exclude him from choices, what is more, if we don't want to add any other managers I make this field optional.

managers/forms.py
```diff
@@ -12,27 +9,26 @@
...
-    def optgroups(self, name: str, value: List, attrs: Optional[Dict] = None) -> List:
-        self.choices.queryset = CustomUser.objects.exclude(user_type=CustomUser.UserType.EMPLOYEE.name).exclude(
-            pk=self.user_pk if hasattr(self, "user_pk") else None
-        )
-        return super().optgroups(name, value, attrs)
-
-
 class ProjectAdminForm(forms.ModelForm):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
+        user_pk = kwargs.pop("user_pk") if "user_pk" in kwargs else None
```
rwrzesien: This could be shortened to `kwargs.pop("user_pk", None)`

managers/forms.py
rwrzesien: Then inheritance doesn't make much sense here any more. It should be used if parent class can contain logic from subclasses so you don't have to repeat this logic. Here now those two forms have different logic.

managers/tests/test_views.py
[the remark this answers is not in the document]
rwrzesien: `aggregated` sounds more adequate I think, `connected` suggest that they are related or something like this.

rwrzesien:
@Szymiks 
> It is implemented in create view by someone else before that PR.

Thanks, I thought you will be adding this in this PR and I didn't saw that code in PR that is why I have asked.

> It is here, before we was excluding all the time user_pk.

OK, what I have originally understood from `we do not remove the manager who created the project.` is that we want to block possibility of removing project creatior from managers, but you have explained it in https://github.com/Code-Poets/sheetstorm/pull/472#discussion_r310101727 so it is clear now (that project creator can be removed from managers in update, right?)

Szymiks:
> OK, what I have originally understood from `we do not remove the manager who created the project.` is that we want to block possibility of removing project creatior from managers, but you have explained it in [#472 (comment)](https://github.com/Code-Poets/sheetstorm/pull/472#discussion_r310101727) so it is clear now (that project creator can be removed from managers in update, right?)

@rwrzesien, Yes he can be removed, but before he was allways removed during first updating.

Issue #478, opened 2019-08-05
Add option to turn off notification for project


Karrp:
Inactivity in some projects shouldn't be notified
Very useful for internal projects

Should be done:
 ------------
 - add boolean flag in project model (notifications turned on as default)
 - add this option to project create and edit form

PR #483, opened 2019-08-05
Add option to switch off notifications

dybi:
resolves: #478


managers/models.py
```diff
@@ -37,6 +37,7 @@ class Project(models.Model):
     suspended = models.BooleanField(default=False)
     managers = models.ManyToManyField(CustomUser, related_name="manager_projects")
     members = models.ManyToManyField(CustomUser, related_name="projects")
+    notifications_on = models.BooleanField(default=True)
```
kbeker: Hmm, maybe `notification_enabled`? :)

users/tests/test_views.py
```diff
@@ -324,3 +324,15 @@ def test_manager_should_only_get_notifications_about_employees_from_his_projects
...
+        self.project.notifications_on = False
+        self.project.save()
+
+        with freeze_time("2019-07-08"):
+            ReportFactory(author=self.employee, project=self.project, date="2019-07-08")
+
+        self.client.force_login(self.manager)
+        with freeze_time("2019-07-15"):
+            response = self.client.get(self.url)
+        self._check_response(response, 200, ["No new notifications about employees in your projects"])
```
kbeker: Please use `UserNotificationsText.NO_MORE_NOTIFICATIONS.value`.

managers/models.py
rwrzesien: Or `is_notification_enabled`.

rwrzesien:
I think it is actually missing excluding those disabled projects from queryset when calculating. But I am not sure how it is supposed to work, as the calculation is done per user, not per project. So in example:
There is an user who is added to project A, B and C.
He was not reporting for 6 days in any project.
You disable notifications for project C.
What does it change? :)

kbeker:
I might be wrong but notifications might need more features in feature #MakeNotificationGreatAgain

Issue #482, opened 2019-08-05
Incorrect display of reports


Karrp:
Undesirable behavior:
After entering one long word in report description the site is expanding and `edit report` button disappears.
Newline character is interpreted as space.

Should be done:
 ------------
 - text wrapping
 - support for the new line character

PR #486, opened 2019-08-06
Bugfix incorrect display of reports

Szymiks:
Resolves: https://github.com/Code-Poets/sheetstorm/issues/482


Szymiks:
> What happens with long urls?

I will check and also I will check text wrapping because It didn't work as I thought

Issue #70, opened 2019-03-01
Completed project shouldn't be suspended


Karrp:
Project with `stop_date` (completed project) shouldn't be suspended.
It should trigger when `stop_date` changes from None to something else.

**Should be done**
----------------
 - create trigger in database like: if (stop_date != None) then suspended = False

rwrzesien:
@Karrp @kbeker I have checked and it looks like this is already resolved, could you please test and verify, and if yes close the issue?

Karrp:
Maybe we'll additionally use a trigger like:
if stop_date then terminated=True

related to #71

rwrzesien:
@Karrp 

> Maybe we'll additionally use a trigger like:
> if stop_date then terminated=True

OK, so we actually want to mark the project with set `stop_date` as terminated, but instead of "terminated" use different term - is that correct?

Karrp:
Yes it is

PR #505, opened 2019-08-12
Bugfix stop_date and suspended field in project

Szymiks:
Resolve: https://github.com/Code-Poets/sheetstorm/issues/70

When project 'stop_date is not null' then it can't be suspended

### DIFF
--- a/managers/models.py
+++ b/managers/models.py
@@ -4,7 +4,10 @@ from typing import Set
 
 from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models import Case
 from django.db.models import Q
+from django.db.models import Value
+from django.db.models import When
 from django.db.models.query import Prefetch
 from django.db.models.query import QuerySet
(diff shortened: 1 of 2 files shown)


managers/models.py
```diff
@@ -86,3 +89,11 @@ def add_default_task_activities(sender: Project, **kwargs: Any) -> None:
         project = kwargs["instance"]
 
         project.project_activities.set(TaskActivityType.objects.get_defaults())
+
+
+@receiver(post_save, sender=Project)
+def change_suspended_if_project_has_stop_date(sender: Project, **kwargs: Any) -> None:
+    assert sender == Project
+    Project.objects.filter(pk=kwargs["instance"].pk).update(
+        suspended=Case(When(~Q(stop_date=None), then=Value(False)), default=kwargs["instance"].suspended)
```
kbeker: Nicely done :)

Issue #497, opened 2019-08-07
Update the look of the edit user page


Karrp:
Similar to #489 
It should look like a mockup, but in the #199 style.

![edit_user_account_details](https://user-images.githubusercontent.com/37847362/62635605-4812a780-b938-11e9-93d6-b45ece65211e.png)

Merrique:
Added to backlog in trello

Issue #489, opened 2019-08-07
Refactor update account view in html as on mockups


Szymiks:
Something between

![edit_user_account_details](https://user-images.githubusercontent.com/47628251/62618208-f1937200-b913-11e9-9003-3f5ffb095527.png)

and
![image](https://user-images.githubusercontent.com/37847362/62692240-9bd0cf80-b9d0-11e9-8cb1-8353b101e58f.png)

Karrp:
Change order of fields

|  |  |
| --- | --- |
| First Name | Last Name |
|Email | Phone |
| Country | Phone |
| Country | Date |
| Admin |  |

Add buttons
Update, Discard, Change password, Delete

Karrp:
Update because RODO

|  |  |
| --- | --- |
| First Name | Last Name |
| Email | User type |

Merrique:
Copy of issue https://github.com/Code-Poets/sheetstorm/issues/497

PR #506, opened 2019-08-12
Feature add update view for account details

Szymiks:
Resolve: https://github.com/Code-Poets/sheetstorm/issues/489
Resolve: https://github.com/Code-Poets/sheetstorm/issues/497


users/templates/partials/display_messages.html
```diff
@@ -0,0 +1,7 @@
+{% if messages %}
+    <ul class="messages" style="list-style-type: none;">
+        {% for message in messages %}
+            <li><font color="green"><b>{{ message }}</b></font></li>
```
dybi: Wouldn't it be better to move this formatting to CSS?

users/templates/partials/user_update_form.html
```diff
@@ -0,0 +1,32 @@
...
+                    <span><img src="{% static 'users/images/grey/ico-01.png' %}"/></span>
+                    <span class="page-title"><b>{% trans 'Account details' %}</b></span>
+                </div>
+            </div>
+            {% include 'partials/display_messages.html' %}
+        </div>
+        <form id="update-user-account" action="{{ current_view_url }}" method="POST" name="update-user-form">
+            {% csrf_token %}
+        {% if admin_is_in_user_update_view %}
+            {% include 'partials/admin_user_update_form.html' %}
```
dybi: both templates seem to be highly duplicated - maybe we can extract some common code?

users/templates/partials/display_messages.html
Szymiks: Definitely! :) Thanks I didn't see that

users/templates/partials/user_update_form.html
Szymiks: I am not sure if it is worth to do, because I could extract this all to one template but with many `if syntaxes` in other way I think I could extract onlly first row because `AdminUpdateForm` and `CustomUpdateForm` have some components in different place. However if you think that it will be better to extract as much as possible even just one row I will do this.

users/templates/partials/user_update_form.html
```diff
@@ -0,0 +1,32 @@
...
+{% with request.path|startswith:custom_users_list_url as admin_is_in_user_update_view %}
+    <div class="main-white-container">
+        <div class="panel-heading">
+            <div class="row">
+                <div class="col-xs-12">
+                    {% if admin_is_in_user_update_view %}
+                        <small>
+                          <a href="{% url 'custom-users-list' %}">
+                            <span class="glyphicon glyphicon-chevron-left"></span>
+                          </a>
```
kbeker: Please fix indents these 3 lines above

users/templates/partials/user_update_form.html
```diff
@@ -0,0 +1,32 @@
...
+                    {% endif %}
+                    <span><img src="{% static 'users/images/grey/ico-01.png' %}"/></span>
+                    <span class="page-title"><b>{% trans 'Account details' %}</b></span>
+                </div>
+            </div>
+            {% include 'partials/display_messages.html' %}
+        </div>
+        <form id="update-user-account" action="{{ current_view_url }}" method="POST" name="update-user-form">
+            {% csrf_token %}
+        {% if admin_is_in_user_update_view %}
```
kbeker: Also here please fix indents because this is inside `<form>` tag

users/templates/partials/admin_user_update_form.html
```diff
@@ -0,0 +1,32 @@
...
+        {{ form.user_type|as_crispy_field }}
+    </div>
+</div>
+<div class="form-group row">
+    <div class="form-group col-md-4 mb-0">
+        <input type="button" id="opener_user_account_update" class="btn btn-success" name="update"
+           value="{% trans 'Update' %}">
+    </div>
+    <div class="form-group col-md-4 mb-0">
+        <input type="button" class="btn btn-secondary" name="discard" value="{% trans 'Discard' %}"
```
Karrp: Due to different interpretations of `btn-secondary` class by browsers, you have to write your own class in css.

Safari Mac OS X version 13.0 (15608.1.37)
image

Opera Mac OS X version 62.0.3331.119
image

Chrome Mac OS X version 76.0.3809.100 (official) (64-bit)
image

users/templates/partials/custom_user_update_form.html
```diff
@@ -0,0 +1,25 @@
...
+        {{ form.last_name|as_crispy_field }}
+    </div>
+</div>
+<div class="form-group row">
+    <div class="form-group col-md-4 mb-0">
+        <input type="button" id="opener_user_account_update" class="btn btn-success" name="update"
+               value="{% trans 'Update' %}">
+    </div>
+    <div class="form-group col-md-4 mb-0">
+        <input type="button" class="btn btn-secondary" name="discard" value="{% trans 'Discard' %}"
```
Karrp: Same as above

users/static/users/center_style.css
```diff
@@ -520,3 +520,81 @@ span {
...
+  display: none;
+}
+
+.margin {
+  margin: 0;
+}
+
+.margin-first {
+  margin-top: 10px;
+}
```
kbeker, line 600: Did you tried not to use `!important` keyword?

users/templates/base.html
```diff
@@ -17,8 +17,8 @@
     />
     <link
         rel="shortcut icon"
-        href="{% static 'images/favicon.ico' %}"
-        integrity="{% staticinline 'images/favicon.ico' encode="sri" %}"
+        href="{% static 'users/images/favicon.ico' %}"
+        integrity="{% staticinline 'users/images/favicon.ico' encode="sri" %}"
```
kbeker, line 21: Why do you move `favicon.ico` to `users` app? This is same for all apps in Sheetstorm and it should stay in main directory

users/templates/partials/admin_user_update_form.html
```diff
@@ -0,0 +1,46 @@
...
+    <div class="form-group col-md-4 mb-0">
+        <input type="button" class="btn btn-secondary discard-button" name="discard" value="{% trans 'Discard' %}"
+               onclick="location.href='{{ current_view_url }}'">
+    </div>
+    <div class="form-group col-md-4 mb-0">
+        <button id="change-password" type="button" class="btn btn-warning" name="password_change" onclick="location.href='{% url 'password_change' %}'">
+            <span>{% trans 'Change password' %}</span>
+        </button>
+    </div>
+{% endif %}
```
kbeker, line 44: One missing indent

users/views.py
```diff
@@ -101,15 +99,15 @@ def get_success_url(self) -> str:
...
+    context_object_name = "user_detail"
     model = CustomUser
 
     def get_object(self, queryset: Optional[QuerySet] = None) -> CustomUser:
         return self.request.user
 
-    def get_form_class(self) -> Type[ModelForm]:
-        if self.request.user.user_type == CustomUser.UserType.ADMIN.name:
-            return self.admins_form_class
+    def get_form_class(self) -> Any:
```
kbeker, line 108: Typing for this place should be `Union[SimpleUserChangeForm, AdminUserChangeForm]`

users/views.py
```diff
@@ -101,15 +99,15 @@ def get_success_url(self) -> str:
 class UserUpdate(UpdateView):
     template_name = "user_update.html"
     form_class = SimpleUserChangeForm
-    admins_form_class = AdminUserChangeForm
```
kbeker, line 104: Why do you delete it and move it directly to method?

users/templates/base.html
maciejSamerdak: Alright, I thought it was a bug, since images everywhere are placed under `app/static/app` and **favicon.ico** is located in the **users** app directory either way. It is also called only once in entire app in *base.html*...
Which is also a part of the **users** app... 😕 
I'd say in that case it should go to somewhere like `sheetstorm/static` along with base.html, but that directory is ignored by git.
It is very confusing to have both directories `users/static/images` and `users/static/users/images` I don't think I'm the only one who would consider this a bug after stumbling upon it.

users/templates/partials/admin_user_update_form.html
maciejSamerdak: The `if` tag, right? I never know for sure how to properly format these html files with those Django tags. Do we have some sort of wiki for it?

users/views.py
maciejSamerdak: The only explanation I can think of right now is totally ludicrous...
... but it might be that I was so fixated on this thing not being implemented, that I totally just ignored the existing (and working!) implementation and wrote the exact same thing from scratch...
And it turns out that this was only a template issue... I feel like I should ask for a reverse-raise or something. 😖

users/views.py
maciejSamerdak: Yeah, totally unnecessary change on my part.
I wonder though, wouldn't it be better to keep a single, parent type (_ModelForm_) like before rather than a _Union_?

users/static/users/center_style.css
maciejSamerdak: Hey now, this wasn't my doing. 😛 
But they don't even work right now, look: they are separated with a semicolon!

maciejSamerdak: Done, I have eradicated all of them.

users/templates/partials/admin_user_update_form.html
maciejSamerdak: Done

users/views.py
maciejSamerdak: I've changed it back to the previous state.

kbeker:
Merged into master.

Issue #493, opened 2019-08-07
Update the look of the project edition page


Karrp:
Simmilar to #492
It should look like a mockup, but in the #199 style.

![update_project](https://user-images.githubusercontent.com/37847362/62633712-e6047300-b934-11e9-9ae0-c643cbcb15b9.png)

Karrp:
It's not finished yet.

Issue #492, opened 2019-08-07
Update the look of the create new project page


Karrp:
Similar to #493
It should look like a mockup, but in the #199 style.

![new_project](https://user-images.githubusercontent.com/37847362/62633449-6d9db200-b934-11e9-9e3a-2198224ce60f.png)

Merrique:
Task in trello backlog

PR #513, opened 2019-08-19
Feature refactor project create view

Szymiks:
Resovles: https://github.com/Code-Poets/sheetstorm/issues/492
Resolve: https://github.com/Code-Poets/sheetstorm/issues/493


Szymiks:
> The buttons should be slightly lower and wider, and the container itself should be shorter, as shown in the picture below.
> image
> 
> Rest is ok for me
@Karrp 
Which container ? You mean managers and members field should be narrower?

Karrp:
> Which container ? You mean managers and members field should be narrower?

@Szymiks Main white container

Karrp:
Merge activities with this view please.
like this:
image

Issue #494, opened 2019-08-07
Update the look of the project reports page


Karrp:
Similar to and after #199  - please update image after end of 199

![image](https://user-images.githubusercontent.com/37847362/62634752-b3f41080-b936-11e9-82b0-ac2e4ed6a955.png)

Should be done:
 ------------
 - 
 - 
 - 
 -

PR #517, opened 2019-08-23
Feature refactor author report list view

Szymiks:
**Related**: https://github.com/Code-Poets/sheetstorm/pull/520
Resolve: https://github.com/Code-Poets/sheetstorm/issues/495

I have changed here also code in `report_list.html`, I was trying to extract duplicated code and code with complicated forloops to partial htmls.

PR #520, opened 2019-08-26
Feature refactor project reports list view

Szymiks:
Resolve : https://github.com/Code-Poets/sheetstorm/issues/494
Related: https://github.com/Code-Poets/sheetstorm/pull/517


Szymiks:
> Where are the margins?
> image
> 
> Why export buttons are on top of site?
> I miss some sum of hours from the whole project.
> 
> Should looks similar to #517

@Karrp but we don't have this feature to count sum of hours project and we never had.

PR #536, opened 2019-09-06
Bugfix hide create project for managers

Szymiks:
No issue for this

This button should be displayed only for admin.


Szymiks:
> So only admins can create projects, yes?

Yes only admin is allowed to create project. Managers are allowed to updated project but can not add and remove managers from project.

PR #543, opened 2019-10-03
Bugfix redirect admin to his own report edition

Karrp:
There is no issue fo this.

Fixes:
 - admin now can delete his own report like any other employee
 - after editing own report admin is redirected to "my reports" site

### DIFF
--- a/employees/templates/employees/partial/display_reports/reports_in_date_for_author.html
+++ b/employees/templates/employees/partial/display_reports/reports_in_date_for_author.html
@@ -1,5 +1,6 @@
 {% load data_display_filters %}
 {% load data_structure_element_selectors %}
+{% url 'custom-report-list' year month as custom_report_list_url %}
 
 {% for reports in date %}
     {% for report in reports.list %}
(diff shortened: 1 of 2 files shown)


employees/templates/employees/partial/display_reports/reports_in_date_for_author.html
```diff
@@ -18,7 +19,15 @@
     </td>
     <td class="edit-button-column Invisible hidden-print">
         {% if report.editable %}
-            <a href="{% if request.user.is_admin %}{% url 'admin-report-detail' pk=report.id %}{% else %}{% url 'custom-report-detail' pk=report.id %}{% endif %}" class="btn btn-light hidden-print">
+            {% if request.user.is_admin%}
+                {% if request.path == custom_report_list_url %}
```
kbeker: Here should not be checked if this url match. Here should be checked if report which user want to edit is his. Please change this condition to check if this report is user's.

Karrp: done

PR #552, opened 2020-04-23
Feature improve monthly hours stats

maciejSamerdak:
The monthly work hours are now always calculated form the beginning of the month.
The information about actual amount of hours spent on project is displayed next to the percentage.

### DIFF
--- a/employees/common/strings.py
+++ b/employees/common/strings.py
@@ -23,8 +23,10 @@ class ReportListStrings(NotCallableMixin, Enum):
     HOURS_PER_MONTH_LABEL = _("Monthly hours")
     EDIT_REPORT_BUTTON = _("Edit")
     NO_PROJECTS_TO_JOIN = _("There are no other projects available.")
-    PROJECTS_WORK_PERCENTAGE_HEADER = _("Projects work percentage")
-    NO_REPORTS_MESSAGE = _("There is no reports to display")
+    NO_REPORTS_MESSAGE = _("There are no reports to display")
+    PROJECTS_WORK_PERCENTAGE_HEADER = _("Monthly work stats per project")
+    HOURS_SUM_HEADER = _("Total hours")
+    HOURS_PERCENTAGE_HEADER = _("Percentage")
 
 
 class ReportDetailStrings(NotCallableMixin, Enum):
@@ -72,7 +74,9 @@ class AuthorReportListStrings(NotCallableMixin, Enum):
     HOURS_PER_MONTH_LABEL = _("Monthly hours")
     NO_REPORTS_MESSAGE = _("This employee has no reports to display.")
     RETURN_BUTTON_MESSAGE = _("Back to employee list")
-    PROJECTS_WORK_PERCENTAGE_HEADER = _("Projects work percentage")
+    PROJECTS_WORK_PERCENTAGE_HEADER = _("Monthly work stats per project")
+    HOURS_SUM_HEADER = _("Total hours")
+    HOURS_PERCENTAGE_HEADER = _("Percentage")
 
 
 class ProjectReportListStrings(NotCallableMixin, Enum):
--- a/employees/tests/test_unit_report_model.py
+++ b/employees/tests/test_unit_report_model.py
@@ -284,7 +284,7 @@ class TestGetProjectsWorkPercentage(TestCase):
         self.project_2 = ProjectFactory()
         self.user.projects.add(self.project_1, self.project_2)
 
-    def test_get_projects_work_percentage_should_return_projects_with_work_time_percent(self):
+    def test_get_projects_work_hours_and_percentage_should_return_projects_with_work_time_percent(self):
         ReportFactory(
             author=self.user, project=self.project_1, work_hours=timezone.timedelta(hours=8), date=timezone.now().date()
         )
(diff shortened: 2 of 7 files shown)


users/models.py
```diff
@@ -175,9 +177,12 @@ def get_projects_work_percentage(self, from_date: Optional[date] = None, to_date
         work_hours_sum = work_hours_per_project.aggregate(
             sum_work_hours_sum=Coalesce(Sum("work_hours_sum"), timezone.timedelta())
         )["sum_work_hours_sum"]
-
         return {
-            project: ((project.work_hours_sum / work_hours_sum) * 100 if work_hours_sum.total_seconds() > 0 else 0)
+            project: (
+                [str(project.work_hours_sum)[:-3], (project.work_hours_sum / work_hours_sum) * 100]
+                if work_hours_sum.total_seconds() > 0
+                else 0
```
kbeker, line 184: Please create from this code separate private method. Method name should tell what is going on in here

employees/tests/test_unit_report_model.py
```diff
@@ -295,12 +295,12 @@ def test_get_projects_work_percentage_should_return_projects_with_work_time_perc
             author=self.user, project=self.project_2, work_hours=timezone.timedelta(hours=8), date=timezone.now().date()
         )
 
-        result = self.user.get_projects_work_percentage()
+        result = self.user.get_projects_work_hours_and_percentage()
 
         self.assertEqual(len(result), 2)
-        self.assertEqual(result, {self.project_1: 40.0, self.project_2: 60.0})
+        self.assertEqual(result, {self.project_1: ["8:00", 40.0], self.project_2: ["12:00", 60.0]})
```
kbeker, line 302: Please also add assert that summary all hours are exactly 20.

employees/tests/test_unit_report_model.py
```diff
@@ -312,12 +312,12 @@ def test_get_projects_work_percentage_should_return_data_from_current_month_star
         )
         self.user.projects.add(self.project_1, self.project_2)
 
-        result = self.user.get_projects_work_percentage()
+        result = self.user.get_projects_work_hours_and_percentage()
 
         self.assertEqual(len(result), 1)
-        self.assertEqual(result, {self.project_1: 100.0})
+        self.assertEqual(result, {self.project_1: ["8:00", 100.0]})
```
kbeker, line 318: Same here about summary hours.

employees/tests/test_unit_report_model.py
maciejSamerdak: But this particular functionality does not involve any sum of sums. `get_projects_work_hours_and_percentage()` returns the exact same dictionary as the one defined in the second assertion. Therefore, I don't see a point in adding any further assertion, if anything I'd say that the first one is already unnecessary.

employees/tests/test_unit_report_model.py
maciejSamerdak: Therefore, same remarks as above 😉

employees/tests/test_unit_report_model.py
kbeker: I mean assumption of this feature is to show hours from current month as a number and as percentages. Same as percentages should sum up to 100, as well hours returned from this function should summary be equal to hours spent by user in current month - that is check what I want you to add :)

employees/common/strings.py
```diff
@@ -72,7 +74,9 @@ class AuthorReportListStrings(NotCallableMixin, Enum):
     HOURS_PER_MONTH_LABEL = _("Monthly hours")
     NO_REPORTS_MESSAGE = _("This employee has no reports to display.")
     RETURN_BUTTON_MESSAGE = _("Back to employee list")
-    PROJECTS_WORK_PERCENTAGE_HEADER = _("Projects work percentage")
+    PROJECTS_WORK_PERCENTAGE_HEADER = _("Monthly work stats per project")
+    HOURS_SUM_HEADER = _("Total hours")
+    HOURS_PERCENTAGE_HEADER = _("Percentage")
```
maciejSamerdak, line 79: I am aware this is slightly redundant, but right now I cannot find any way to concatenate enums, since they do not allow subclassing. Unless there is some convenient workaround I am not aware of, we'd probably have to change those to regular dicts + custom template tag for accessing dicts (if one wasn't implemented already).
But this is a major change that I think goes beyond this PR.

users/models.py
```diff
@@ -188,6 +186,19 @@ def get_project_ordered_by_last_report_creation_date(self) -> QuerySet:
             )
         ).order_by("-last_report_creation_date")
 
+    @staticmethod
+    def __parse_work_hours_and_percentage_from_queryset_to_a_dict(
```
maciejSamerdak, line 190: The current unit tests in `test_mixins.py` already cover this portion of code. Therefore I didn't add any additional unit tests for this function alone.

employees/common/strings.py
kbeker: I think that this is not so important. Maybe your idea with enums would be good for this purpose but for now it is not so important :)

users/models.py
kbeker, line 190: @maciejSamerdak please fix it in some future PR, but please use only one `_` in private methods :)

utils/mixins.py
```diff
@@ -53,19 +53,18 @@ def get_context_data(self, **kwargs: Any) -> dict:
             year = int(self.kwargs["year"])
             month = int(self.kwargs["month"])
             current_date = timezone.now().date()
+            from_date = timezone.now().date().replace(year=year, month=month, day=1)
```
kbeker, line 56: Please change it to: `timezone.datetime(year=year, month=month, day=1).date()`

utils/tests/test_mixins.py
```diff
@@ -190,20 +194,26 @@ def test_project_work_percentage_mixin_should_call_get_projects_work_percentage_
...
-        with patch("users.models.CustomUser.get_projects_work_percentage") as get_projects_work_percentage:
+        with patch(
+            "users.models.CustomUser.get_projects_work_hours_and_percentage"
+        ) as get_projects_work_hours_and_percentage:
             response = self.view(self.request, year=current_date.year, month=current_date.month)
 
         self.assertEqual(response.status_code, 200)
-        get_projects_work_percentage.assert_called_once_with(current_date - timezone.timedelta(days=30), current_date)
+        get_projects_work_hours_and_percentage.assert_called_once_with(
+            timezone.now().date().replace(year=current_date.year, month=current_date.month, day=1), current_date
```
kbeker, line 204: Instead of `timezone.now().date.replace(...)` use `timezone.datetime(...).date()`

users/models.py
maciejSamerdak: Done

utils/tests/test_mixins.py
maciejSamerdak: Done.

utils/mixins.py
maciejSamerdak: Done.

PR #553, opened 2020-04-24
Bugfix inactive users are displayed in app

maciejSamerdak:
Fixes multiple issues regarding deactivating users by marking them as inactive using the _is_active_ field in _**CustomUser**_ model.

Now after deactivating a user through the admin panel, the following will happen:

- The user won't be listed on the employees' list anymore

- His user type will be changed to an employee and all of his privileges will be disabled

- The user will be removed from all projects as both member and a manager

- All of his reports will remain intact and will be displayed for his former projects

- It will be impossible to pick this user as a member or manager during project creation and editing

It is also confirmed that deactivation results in disabling notifications regarding the user and an appropriate unit test has been added.

### DIFF
--- a/employees/tests/test_views.py
+++ b/employees/tests/test_views.py
@@ -703,6 +703,40 @@ class ProjectReportListTests(TestCase):
             ),
         )
 
+    def test_project_report_list_view_should_display_inactive_members_reports(self):
+        current_date = datetime.datetime.now().date()
+        previous_date = current_date - datetime.timedelta(days=30)
+        inactive_user = UserFactory(is_active=False)
+
+        project_report = ReportFactory(
+            author=inactive_user, date=current_date, project=self.project, description="This is for current project."
+        )
+        other_project_report = ReportFactory(
+            author=inactive_user, date=current_date, description="This is for other project."
+        )
+        project_report_from_other_month = ReportFactory(
+            author=inactive_user, date=previous_date, project=self.project, description="This is for another month."
+        )
+
+        response = self.client.get(self.url)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, project_report.description)
+        self.assertNotContains(response, other_project_report.description)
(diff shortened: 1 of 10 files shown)


users/forms.py
```diff
@@ -52,6 +55,16 @@ class Meta:
...
         fields = "__all__"
 
+    def clean(self) -> Dict[str, Any]:
+        cleaned_data = super().clean()
+        cleaned_is_active = cleaned_data.get("is_active")
+        if self.instance.is_active and not cleaned_is_active:
+            cleaned_data["user_type"] = CustomUser.UserType.EMPLOYEE.name
+            cleaned_data["is_staff"] = False
+            cleaned_data["is_superuser"] = False
+
```
maciejSamerdak, line 65: This is really only a formality which I added, since it was mentioned that "a deactivated user should lose all of his privileges". However, I should point out that _is_active_ takes priority over _is_staff_ or _is_superuser_. Therefore, even if he does have those privileges, he still won't be able to sign in to the admin panel, as long as he's marked as inactive.
It's because of this irrelevancy that I only tested it manually and did not implement any formal unit test for this particular part. The more important user type change however is tested as a part of the signal's unit test.

employees/tests/test_views.py
```diff
@@ -703,6 +703,35 @@ def test_project_report_list_view_should_redirect_to_current_date_if_date_parame
...
+            reverse(
+                "project-report-list",
+                kwargs={"pk": self.project.pk, "year": previous_date.year, "month": previous_date.month},
+            )
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertNotContains(response, project_report.description)
+        self.assertNotContains(response, other_project_report.description)
+        self.assertContains(response, project_report_from_other_month.description)
+
```
kbeker, line 734: Please always leave one space between logical parts in test. For example:
Creating needed variables
----blank line----
request/needed operations
--- black line---
asserts etc.

PR #554, opened 2020-05-03
Chore update all project dependecies

kbeker:
Necessary dependencies update

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -166,6 +166,7 @@ disable=print-statement,
         abstract-method,
         arguments-differ,
         cyclic-import,
+        invalid-str-returned,
 
 
 # Enable the message, report, category or checker with the given id(s). You can
--- a/pytest.ini
+++ b/pytest.ini
@@ -7,3 +7,4 @@ python_files = tests.py test_*.py *_tests.py
 filterwarnings =
     ignore:.*Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated.*:DeprecationWarning
     ignore:.*Using the add method to register a processor or pattern is deprecated. Use the `register` method instead.*:DeprecationWarning
+    ignore:.*smart_text.*is deprecated in favor of smart_str:PendingDeprecationWarning
--- a/users/tests/test_views.py
+++ b/users/tests/test_views.py
@@ -52,9 +52,11 @@ class ChangePasswordTests(TestCase):
         response = self.client.post(reverse("password_change"), data)
         self.user.refresh_from_db()
         self.assertEqual(response.status_code, 200)
-        self.assertTrue(
-            ValidationErrorText.VALIDATION_ERROR_SIGNUP_PASSWORD_MESSAGE
-            in response.context["form"].errors.get("new_password2")
+        # Replacement between "’" and "'" it is caused because in response
+        # we got "’" but from strings hardcoded in `strings.py` we got "'"
+        self.assertEqual(
+            response.context["form"].errors.get("new_password2")[0].replace("’", "'"),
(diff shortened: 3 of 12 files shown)


utils/mixins.py
```diff
@@ -42,7 +42,7 @@ def get_queryset(self) -> QuerySet:
 
 class ProjectsWorkPercentageMixin(ContextMixin):
     def get_context_data(self, **kwargs: Any) -> dict:
-        from users.models import CustomUser
+        from users.models import CustomUser  # pylint: disable=import-outside-toplevel
```
maciejSamerdak, line 45: Why aren't these imported at toplevel anyway?

pylintrc
```diff
@@ -166,6 +166,7 @@ disable=print-statement,
         abstract-method,
         arguments-differ,
         cyclic-import,
+        invalid-str-returned,
```
maciejSamerdak, line 169: This seems important, why is this disabled?

users/tests/test_views.py
```diff
@@ -52,9 +52,11 @@ def test_change_user_password_view_should_not_change_user_password_when_new_pass
         response = self.client.post(reverse("password_change"), data)
         self.user.refresh_from_db()
         self.assertEqual(response.status_code, 200)
-        self.assertTrue(
-            ValidationErrorText.VALIDATION_ERROR_SIGNUP_PASSWORD_MESSAGE
-            in response.context["form"].errors.get("new_password2")
+        # Replacement between "’" and "'" it is caused because in response
+        # we got "’" but from strings hardcoded in `strings.py` we got "'"
```
maciejSamerdak, line 56: And this happens after the Django update? What the heck is this?

utils/mixins.py
kbeker: I didn't want to change it during this update. It might be change as separate chore, but now it his PR :)

pylintrc
kbeker: Because it seems to give wrong results. It was raised on for example `__str__` method in models when this method exactly returns string

pytest.ini
```diff
@@ -7,3 +7,4 @@ python_files = tests.py test_*.py *_tests.py
 filterwarnings =
     ignore:.*Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated.*:DeprecationWarning
     ignore:.*Using the add method to register a processor or pattern is deprecated. Use the `register` method instead.*:DeprecationWarning
+    ignore:.*smart_text.*is deprecated in favor of smart_str:PendingDeprecationWarning
```
maciejSamerdak, line 10: When is this thrown? Can we somehow update our code to the smart_str?

users/tests/test_views.py
kbeker: Yea, I didn't want to dig deep into django to find what exactly changed this is only about apostrophe :/

pytest.ini
kbeker: This warning is thrown inside 3-party library so we can't do anything about this

users/tests/test_views.py
maciejSamerdak: Yeah, I hear you, but it's so baffling! Should we add this to Jira to investigate later maybe?

utils/mixins.py
maciejSamerdak: Ok. Guess we should add this to Jira then?

pytest.ini
maciejSamerdak: Ok. I thought so. ☹️

users/tests/test_views.py
kbeker: Sure, please add it!

utils/mixins.py
kbeker: Please add it:)

PR #556, opened 2020-05-16
Bugfix monthly hour stats display

maciejSamerdak:
Fixes following bugs:
- Hours total sum is displayed with days
- If table is empty, it is still displayed
- Reports for manager-only projects are not counted
- All active projects are displayed, even if there's nothing to show for them


users/models.py
```diff
@@ -159,6 +160,10 @@ def is_manager(self) -> bool:
     def is_employee(self) -> bool:
         return self.user_type == CustomUser.UserType.EMPLOYEE.name
 
+    @property
+    def all_projects(self) -> QuerySet:
+        return (self.projects.all() | self.manager_projects.all()).distinct()
```
maciejSamerdak, line 165: It is forbidden to use filters with `union()` and `|` operator behaves strangely at times, duplicating projects.
Despite the distinct filter, `annotate()` can somehow still multiply results without setting the `distinct` flag.
If we were to use this property elsewhere, other annotation operations may not work as desired.
The only other solution is to repeat operations for each queryset and only then perform `union()` on them.

kbeker:
Merged into master.

PR #557, opened 2020-05-16
Change ordering in employees' list

maciejSamerdak:
Changes ordering in employees list in following hierarchy:
**user type** -> **last name** -> **first name** -> **email**

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -145,13 +145,27 @@ class UserUpdateByAdmin(UpdateView):
 class UserList(ListView):
     template_name = "users_list.html"
     model = CustomUser
-    queryset = CustomUser.objects.active().prefetch_related("projects")
+    queryset = (
+        CustomUser.objects.active()
+        .order_by("user_type", "last_name", "first_name", "email")
+        .prefetch_related("projects")
+    )
 
     def get_context_data(self, *, _object_list: Any = None, **kwargs: Any) -> dict:
         context_data = super().get_context_data(**kwargs)
         context_data["year"] = datetime.datetime.now().year
         context_data["month"] = datetime.datetime.now().month
+        context_data["ordered_object_list"] = self._get_ordered_list_of_users()
         return context_data
+
+    def _get_users_by_user_type(self, user_type: str) -> QuerySet:
+        return self.queryset.filter(user_type=user_type)
+
+    def _get_ordered_list_of_users(self):
+        admins = self._get_users_by_user_type(CustomUser.UserType.ADMIN.name)
+        managers = self._get_users_by_user_type(CustomUser.UserType.MANAGER.name)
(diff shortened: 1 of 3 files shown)


users/views.py
```diff
@@ -144,12 +144,20 @@ def form_valid(self, form: SimpleUserChangeForm) -> HttpResponse:
...
+    )
 
     def get_context_data(self, *, _object_list: Any = None, **kwargs: Any) -> dict:
         context_data = super().get_context_data(**kwargs)
         context_data["year"] = datetime.datetime.now().year
         context_data["month"] = datetime.datetime.now().month
+        admins = self.queryset.filter(user_type=CustomUser.UserType.ADMIN.name)
+        managers = self.queryset.filter(user_type=CustomUser.UserType.MANAGER.name)
+        employees = self.queryset.filter(user_type=CustomUser.UserType.EMPLOYEE.name)
+        context_data["ordered_object_list"] = list(admins) + list(managers) + list(employees)
```
kbeker, line 160: Please extract this to private methods for example something like this:
```python
def _get_users_by_user_type(self, user_type):
    return self.queryset.filter(user_type=user_type)

def _get_ordered_users_by_user_type(self):
    admins = self._get_users_by_user_type(user_type=CustomUser.UserType.ADMIN.name)
    managers = self._get_users_by_user_type(user_type=CustomUser.UserType.MANAGER.name)
    employees = self._get_users_by_user_type(user_type=CustomUser.UserType.EMPLOYEE.name)
    return list(admins) + list(managers) + list(employees)

context_data["ordered_object_list"] = self._get_ordered_users_by_user_type()
```
Also there is no test for this - please add some.

I can't find different way to order all users in one query, but I can't believe that there is not other way to query that queryset should by ordered by uset_type where user_type first is `ADMIN` then `MANAGER` etc. But don't waste time for this right now. I just telling you this to keep it in mind :)

maciejSamerdak: Actually, I could try to either use Python's `sort` (but this one returns a list) or write a function in the model manager. I was hesitant to use the latter, because I found it too complex for such simple task, but now that I look at it again, I think I could try it.

maciejSamerdak: Alright, I checked that model manager solution and it involves calling `annotate` to create a new integer field. It's values would be set based on the value of the field we want to sort by (so for "ADMIN" we set 0, for "MANAGER" we set 1, etc.). Then we call `.order_by` and pass that new field we created.
Seems like it's impossible in Django to fiddle with database ordering on the backend side.

kbeker: I think that we could accomplished that with some changes in `Employee` model. I just though about changing user type into `IntEnum`s and by this we could get that really simple  sorting mechanism but I think that we for this moment we can leave as it is:)

PR #565, opened 2020-07-27
Bugfix report list layout issues

maciejSamerdak:
Fixes several issues with report list layout:
- The export buttons are now hidden whenever there are no reports to display.
- The actual report list is now the only scrollable item in the x-axis, when viewport is too narrow. On project report list, each table can be scrolled individually.
- Table layout has been redesigned, each column now has custom width settings which allows report description to take the most space.


maciejSamerdak:
The change with the summary was my idea. I figured since those tables are very small, 2-3 columns, they look terrible stretched out.

And what do you mean by "centered" about those buttons? The paddings are the same as every other cell. Do you mean centered vertically? But won't that create a graphical inconsistency on longer descriptions or lower resolutions?

kbeker:
Won't do

PR #569, opened 2020-08-07
Chore command to create projects initial data

marsolekk:


sheetstorm/management/commands/generate_test_data.py
```diff
@@ -1,15 +1,23 @@
...
+from managers.models import Project
 from users.factories import UserFactory
 
 CustomUser = get_user_model()
 
 superuser_user_type = "SUPERUSER"
 
+suspended_project = "suspended"
+active_project = "active"
+completed_project = "completed"
```
kbeker, line 19: Please create from these variables enum class

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -26,7 +34,13 @@ def handle(self, *args, **options):
         if is_need_to_create_superuser:
             self.create_user(superuser_user_type)
 
+        project_options = self._get_project_options(options)
+
+        for (project_type, number_of_projects) in project_options.items():
+            self.create_project(project_type, number_of_projects)
```
kbeker, line 40: My proposal is to extract it also to separate instance method instead of keeping more and more things in `handle()`

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -61,6 +75,42 @@ def _set_user_factory_parameters(user_type):
             "is_superuser": user_type == superuser_user_type,
         }
 
+    @staticmethod
+    def _get_project_options(options):
```
kbeker, line 79: Too general method name. You're saying here that you're giving some kind of options in this method but from method body I can see that you're giving some numbers for projects - probably number how many projects should be created. Please be more precise in method name - if method does more things and it is hard to name this method please extract it to additional methods

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -61,6 +75,42 @@ def _set_user_factory_parameters(user_type):
...
+        return {
+            suspended_project: options[suspended_project] - Project.objects.filter_suspended().count()
+            if options[suspended_project] is not None
+            else 0,
```
kbeker, line 83: You can extract this part to separate method with parameters as switcher for suspended/active/completed projects.

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -61,6 +75,42 @@ def _set_user_factory_parameters(user_type):
...
+    def _set_project_factory_parameters(self, project_type):
+        return {
+            "start_date": self._set_project_start_date(timezone.now()),
+            "suspended": project_type == suspended_project,
+            "stop_date": self._set_project_stop_date(timezone.now()) if project_type == completed_project else None,
+        }
+
+    @staticmethod
+    def _set_project_start_date(date):
+        return date + relativedelta(months=-1, day=1)
```
kbeker, line 108: It will be problematic that you're giving here date and from this date you're subtract period of time. It should be done differently - from current date you should subtract period of time which should be has default value but also should be modifiable.

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -61,6 +75,42 @@ def _set_user_factory_parameters(user_type):
...
+
+    @staticmethod
+    def _set_project_stop_date(date):
+        return date + relativedelta(days=-14)
```
kbeker, line 112: Same here about current time

sheetstorm/management/commands/generate_test_data.py
kbeker: Also if you want to subtract date don't add negative date. Just subtract positive date and it will also work

sheetstorm/management/commands/generate_test_data.py
marsolekk: Do you want me to add this class in this file or create new class in Project class from managers/models.py like this is solved in CustomUser class case?

sheetstorm/management/commands/generate_test_data.py
marsolekk: Done

sheetstorm/management/commands/generate_test_data.py
marsolekk: Done

sheetstorm/management/commands/generate_test_data.py
marsolekk: Done

sheetstorm/management/commands/generate_test_data.py
marsolekk: Done

sheetstorm/management/commands/generate_test_data.py
marsolekk: Done

sheetstorm/management/commands/generate_test_data.py
marsolekk: Done

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -61,6 +75,65 @@ def _set_user_factory_parameters(user_type):
...
+            ProjectType.COMPLETED.name: self._compute_number_of_completed_projects_to_create(options),
+        }
+
+    @staticmethod
+    def _compute_number_of_suspended_projects_to_create(options):
+        return (
+            options[ProjectType.SUSPENDED.name] - Project.objects.filter_suspended().count()
+            if options[ProjectType.SUSPENDED.name] is not None
+            else 0
+        )
```
kbeker, line 97:
```python
class UnsupportedProjectTypeException(Exception):
    pass


def _compute_number_of_suspended_projects_to_create(options, project_type):
    if project_type == ProjectTypeEnum. SUSPENDED.name:
        zmienna = Project.objects.filter_suspended().count()
    else:
        raise UnsupportedProjectTypeException("")
    return options[zmienna] - zmienna if options[zmienna] is not None else 0
```

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -1,16 +1,27 @@
...
+from managers.factories import ProjectFactory
+from managers.models import Project
 from users.factories import UserFactory
 
 CustomUser = get_user_model()
 
 superuser_user_type = "SUPERUSER"
 
 
+class ProjectType(Enum):
```
kbeker, line 19: I was thinking about this and I think that we can leave it here as it is ;)

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -61,6 +75,65 @@ def _set_user_factory_parameters(user_type):
...
+
+    def _set_project_factory_parameters(self, project_type):
+        return {
+            "start_date": self._create_start_date(),
+            "suspended": project_type == ProjectType.SUSPENDED.name,
+            "stop_date": self._create_stop_date() if project_type == ProjectType.COMPLETED.name else None,
+        }
+
+    @staticmethod
+    def _create_start_date(time_delta=relativedelta(months=1, day=1)):
```
kbeker, line 130: I think that `time_delta` should be class variable and you should it here. Something like: `PROJECT_START_DATE_TIME_DELTA = relativedelta(months=1, day=1)` and then:
```python
def _create_start_date(time_delta=PROJECT_START_DATE_TIME_DELTA)
```

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -61,6 +75,65 @@ def _set_user_factory_parameters(user_type):
...
+        return timezone.now() - time_delta
+
+    @staticmethod
+    def _create_stop_date(time_delta=relativedelta(days=14)):
```
kbeker, line 134: Same here

sheetstorm/management/commands/generate_test_data.py
marsolekk: Done

sheetstorm/management/commands/generate_test_data.py
marsolekk: Done

sheetstorm/management/commands/generate_test_data.py
marsolekk: Ok :)

sheetstorm/management/commands/generate_test_data.py
marsolekk: Done

PR #570, opened 2020-08-12
Chore add missing type annotation

marsolekk:


sheetstorm/management/commands/generate_test_data.py
```diff
@@ -26,7 +30,7 @@ class Command(BaseCommand):
     help = "Create initial sample data for SheetStorm application testing."
 
     @transaction.atomic
-    def handle(self, *args, **options):
+    def handle(self, *args: Any, **options: Union[bool, int, None]) -> None:
```
kbeker, line 33: You're using everywhere else for `options` typing like this: `Dict[str, Union[bool, int, None]]`. Why haven't  you used here it also?

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -43,21 +47,21 @@ def handle(self, *args, **options):
         logging.info(f"Total number of projects in the database: {Project.objects.count()}")
 
     @staticmethod
-    def _get_user_options(options):
+    def _get_user_options(options: Dict[str, Union[bool, int, None]]) -> Dict[str, Optional[int]]:
```
kbeker, line 50: Now when I'm looking at name "options" I think it is too general, but it is not for now. You will need to fix it in some next PRs

kbeker: Also if you're using multiple times the same typing like `Dict[str, Union[bool, int, None]]`  you can extract it to separate well named variable and use instead of whole `Dict[str, Union[bool, int, None]]`

sheetstorm/management/commands/generate_test_data.py
marsolekk: There is only need to type annotation of potential values in **kwargs, and options will be `Dict[str, Union[bool, int, None]]`. If I will  write here a Dict type, I will get `Dict[str, Dict[str, Union[bool, int, None]]]` instead.

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -33,6 +33,8 @@ class UnsupportedProjectTypeException(Exception):
 class Command(BaseCommand):
     help = "Create initial sample data for SheetStorm application testing."
 
+    types_of_options_dictionary_content = Dict[str, Union[bool, int, None]]
```
kbeker, line 36: I think that it would be better to have this variable as OptionsDictType. We need to try follow typing convention.

kbeker: And that's why now I see that `options` variable is not good here and we need to change it to different more meaningful

PR #571, opened 2020-08-14
Chore add unit tests for command generate test data

marsolekk:


sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+        "EMPLOYEE": None,
+        "MANAGER": None,
+        "SUPERUSER": False,
+        "SUSPENDED": number_of_projects_to_create,
+        "ACTIVE": number_of_projects_to_create,
+        "COMPLETED": number_of_projects_to_create,
+    }
+
+
+class CreateUserMethodsTests(TestCase, CustomCommandOptions):
```
maciejSamerdak, line 33: In projects-related tests you are testing for quantity computation based on objects that already exist in the database, but I can't see any counterpart here.

Also, correct me if I'm mistaken, but I believe that at least one superuser was meant to always exist after the method is run? So why not test that? What if there are no superusers in the database. What if there are superusers in the database? What happens if we specify the number of superusers in both of those cases?

And what if we set a number of users that's lower than the existing quantity? I think that's also worth testing.

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+            self.assertEqual(
+                CustomUser.objects.filter(is_superuser=True).count(), CustomCommandOptions.number_of_users_to_create
+            )
+        else:
+            self.assertEqual(
+                CustomUser.objects.filter(user_type=user_type).count(), CustomCommandOptions.number_of_users_to_create
+            )
+
+
+class CreateProjectMethodsTests(TestCase):
```
maciejSamerdak, line 101: Any counterparts of missing tests that I described under the user's tests that aren't applied here: they also apply here 😛

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+        self.assertFalse(Command._get_superuser_request({self.SUPERUSER: False}))
+
+    @parameterized.expand(
+        [CustomUser.UserType.ADMIN.name, CustomUser.UserType.EMPLOYEE.name, CustomUser.UserType.MANAGER.name, SUPERUSER]
+    )
+    def test_only_admins_and_superuser_should_have_staff_property(self, user_type):
+        if user_type in (CustomUser.UserType.ADMIN.name, self.SUPERUSER):
+            self.assertTrue(Command._set_user_factory_parameters(user_type).get("is_staff"))
+        else:
+            self.assertFalse(Command._set_user_factory_parameters(user_type).get("is_staff"))
```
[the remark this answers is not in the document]
maciejSamerdak: Alternatively, you can define a second bool parameter `expected_result` and change the assertion to **assertEqual**, like this:
```
self.assertEqual(Command._set_user_factory_parameters(user_type).get("is_staff"), expected_result)
```

Both of those approaches I will consider appropriate. Go with whatever you find best. 👍

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+from parameterized import parameterized
+
+from managers.models import Project
+from sheetstorm.management.commands.generate_test_data import Command
+from sheetstorm.management.commands.generate_test_data import ProjectType
+from sheetstorm.management.commands.generate_test_data import UnsupportedProjectTypeException
+from users.models import CustomUser
+
+
+class CustomCommandOptions:
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+    number_of_projects_to_create = 1
+    number_of_users_to_create = 1
+    OPTIONS = {
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+            CustomUser.UserType.ADMIN.name: None,
+            CustomUser.UserType.EMPLOYEE.name: None,
+            CustomUser.UserType.MANAGER.name: None,
+        }
+        self.assertDictEqual(Command._get_user_options(CustomCommandOptions.OPTIONS), expected_options)
+
+    def test_function_should_return_true_if_is_superuser_request(self):
+        self.assertTrue(Command._get_superuser_request({self.SUPERUSER: True}))
+
+    def test_function_should_return_false_if_not_is_superuser_request(self):
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+        "SUSPENDED": number_of_projects_to_create,
+        "ACTIVE": number_of_projects_to_create,
+        "COMPLETED": number_of_projects_to_create,
+    }
+
+
+class CreateUserMethodsTests(TestCase, CustomCommandOptions):
+    SUPERUSER = "SUPERUSER"
+
+    def test_function_should_return_only_users_options(self):
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+                CustomCommandOptions.number_of_users_to_create,
+            )
+        elif user_type == self.SUPERUSER:
+            self.assertEqual(
+                CustomUser.objects.filter(is_superuser=True).count(), CustomCommandOptions.number_of_users_to_create
+            )
+        else:
+            self.assertEqual(
+                CustomUser.objects.filter(user_type=user_type).count(), CustomCommandOptions.number_of_users_to_create
+            )
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+    @parameterized.expand([ProjectType.SUSPENDED.name, ProjectType.ACTIVE.name, ProjectType.COMPLETED.name])
+    def test_function_should_return_difference_between_specified_number_and_number_of_projects_in_database(
+        self, project_type
+    ):
+        self.assertEqual(
+            Command._compute_number_of_projects_to_create(CustomCommandOptions.OPTIONS, project_type),
+            CustomCommandOptions.number_of_projects_to_create,
+        )
+
+    def test_function_should_raise_error_if_project_type_do_not_exist(self):
```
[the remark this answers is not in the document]
marsolekk: :confused:
Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+            expected_options,
+        )
+
+    def test_start_date_should_be_datetime_type_and_be_in_past(self):
+        self.assertTrue(isinstance(Command._create_start_date(), timezone.datetime))
+        self.assertLess(Command._create_start_date(), timezone.now())
+
+    def test_stop_date_should_be_datetime_type_and_be_in_past(self):
+        self.assertTrue(isinstance(Command._create_stop_date(), timezone.datetime))
+        self.assertLess(Command._create_stop_date(), timezone.now())
```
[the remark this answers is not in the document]
marsolekk: :disappointed: 
Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+    @parameterized.expand([ProjectType.SUSPENDED.name, ProjectType.ACTIVE.name, ProjectType.COMPLETED.name])
+    def test_only_completed_projects_should_have_stop_date(self, project_type):
+        if project_type == ProjectType.COMPLETED.name:
+            self.assertTrue(
+                Command(CustomCommandOptions)._set_project_factory_parameters(project_type).get("stop_date")
+            )
+        else:
+            self.assertFalse(
+                Command(CustomCommandOptions.OPTIONS)._set_project_factory_parameters(project_type).get("stop_date")
+            )
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+        if project_type == ProjectType.SUSPENDED.name:
+            self.assertEqual(
+                Project.objects.filter_suspended().count(), CustomCommandOptions.number_of_projects_to_create
+            )
+        elif project_type == ProjectType.ACTIVE.name:
+            self.assertEqual(Project.objects.filter_active().count(), CustomCommandOptions.number_of_projects_to_create)
+        elif project_type == ProjectType.COMPLETED.name:
+            self.assertEqual(
+                Project.objects.filter_completed().count(), CustomCommandOptions.number_of_projects_to_create
+            )
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+from managers.models import Project
+from sheetstorm.management.commands.generate_test_data import Command
+from sheetstorm.management.commands.generate_test_data import ProjectType
+from sheetstorm.management.commands.generate_test_data import UnsupportedProjectTypeException
+from users.models import CustomUser
+
+
+class CustomCommandOptions:
+    number_of_projects_to_create = 1
+    number_of_users_to_create = 1
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+    def test_number_of_projects_to_create_should_be_set_and_equal_to_computed_number(self):
+        expected_options = {
+            ProjectType.ACTIVE.name: CustomCommandOptions.number_of_projects_to_create,
+            ProjectType.SUSPENDED.name: CustomCommandOptions.number_of_projects_to_create,
+            ProjectType.COMPLETED.name: CustomCommandOptions.number_of_projects_to_create,
+        }
+        self.assertDictEqual(
+            Command(CustomCommandOptions)._set_number_of_projects_to_create(CustomCommandOptions.OPTIONS),
+            expected_options,
+        )
```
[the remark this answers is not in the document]
marsolekk: I create some projects at the beginning of each test using setUp method. I hope that it is better approach. :smile:

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: To add superuser to database required positional argument. There is no possibility to specify number of superusers to create, so only one superuser is created by default, but only if you add argument and superuser not exists in the database. 
I added test that you mentioned at the end. :)

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: I hope I done it well :)

marsolekk: I hope I applied all your suggestions :)

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
+from dateutil.relativedelta import relativedelta
+from django.test import TestCase
+from django.utils import timezone
+from parameterized import parameterized
+
+from managers.factories import ProjectFactory
+from managers.models import Project
+from sheetstorm.management.commands.generate_test_data import Command
```
kbeker, line 8: Mostly it is wrong behavior to change name of imported package/class/function but in this case I think it will be more readable and meaningful to use different name than just `Command`. Please change it to use different name like "GenerateTestDataCommand". I think it will be more friendly name :)

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+        self.number_of_active_projects_to_create = (
+            self.options.get(ProjectType.ACTIVE.name) - Project.objects.filter_active().count()
+        )
+        self.number_of_completed_projects_to_create = (
+            self.options.get(ProjectType.COMPLETED.name) - Project.objects.filter_completed().count()
+        )
+
+    def tearDown(self):
+        Project.objects.all().delete()
```
kbeker, line 123: You don't have to do this. TestCase should do it for you

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+        self.options = {ProjectType.SUSPENDED.name: 4, ProjectType.ACTIVE.name: 5, ProjectType.COMPLETED.name: 7}
+
+        for _project_number in range(2):
+            ProjectFactory(suspended=True)
+
+        for _project_number in range(2):
+            ProjectFactory()
+
+        for _project_number in range(2):
+            ProjectFactory(stop_date=timezone.now() - relativedelta(days=1))
```
kbeker, line 110: These setUp things should be extracted to separate method to be more understandable for any future reader. Please extract them to methods like "_create_suspended_projects", etc

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+        Command().create_user(user_type, number_of_users_in_database)
+        Command().create_user(user_type, number_of_users_to_create)
+        self.assertEqual(CustomUser.objects.filter(user_type=user_type).count(), number_of_users_in_database)
+
+
+class CreateProjectMethodsTests(TestCase):
+    def setUp(self):
+        self.options = {ProjectType.SUSPENDED.name: 4, ProjectType.ACTIVE.name: 5, ProjectType.COMPLETED.name: 7}
+
+        for _project_number in range(2):
```
kbeker, line 103: Numbers like here `2` should be described why you are creating here exactly two projects to three or three milion.

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+        self, user_type, number_of_users_in_database, number_of_users_to_create
+    ):
+        Command().create_user(user_type, number_of_users_in_database)
+        Command().create_user(user_type, number_of_users_to_create)
+        self.assertEqual(CustomUser.objects.filter(user_type=user_type).count(), number_of_users_in_database)
+
+
+class CreateProjectMethodsTests(TestCase):
+    def setUp(self):
+        self.options = {ProjectType.SUSPENDED.name: 4, ProjectType.ACTIVE.name: 5, ProjectType.COMPLETED.name: 7}
```
kbeker, line 101: Same here about describing numbers inside this dict

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+            ProjectFactory(suspended=True)
+
+        for _project_number in range(2):
+            ProjectFactory()
+
+        for _project_number in range(2):
+            ProjectFactory(stop_date=timezone.now() - relativedelta(days=1))
+
+        self.number_of_suspended_projects_to_create = (
+            self.options.get(ProjectType.SUSPENDED.name) - Project.objects.filter_suspended().count()
```
kbeker, line 113: To avoid query after database results here you can handle differently creating test projects. Instead of creating about Projects without handling return instance from `ProjectFactory` handle it inside list and then just check length of it. That's how you can simplify this setUp part.
For example:
```python
def _generate_test_projects(number_of_projects, project_type):
    project_list = []
    for _ in range(number_of_projects):
        if project_type == ProjectType.SUSPENDED.name:
            project = ProjectFactory(suspended=True)
        elif ....

        project_list.append(project)
    return project_list

self.suspended_projects = self._generate_test_projects(number_of_projects=3, project_type=ProjectType.SUSPENDED.name)

self.suspended_projects_to_create = self.options.[ProjectType.SUSPENDED.name] - len(self.suspended_projects)

```

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+            (CustomUser.UserType.EMPLOYEE.name, 4, 3),
+            (CustomUser.UserType.MANAGER.name, 2, 1),
+        ]
+    )
+    def test_create_user_function_should_not_create_users_if_number_in_database_is_greater_than_or_equal_to_specified(
+        self, user_type, number_of_users_in_database, number_of_users_to_create
+    ):
+        Command().create_user(user_type, number_of_users_in_database)
+        Command().create_user(user_type, number_of_users_to_create)
+        self.assertEqual(CustomUser.objects.filter(user_type=user_type).count(), number_of_users_in_database)
```
kbeker, line 96: Please always divide logical parts in tests with blank line. For example here blank line would be great about self.assertEqual line because in first part you are creating data to test and in the second you are checking this

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+    ):
+        self.assertEqual(Command._compute_number_of_projects_to_create({project_type: None}, project_type), 0)
+
+    def test_number_of_projects_to_create_should_be_set_and_equal_to_computed_number(self):
+        expected_options = {
+            ProjectType.ACTIVE.name: self.number_of_active_projects_to_create,
+            ProjectType.SUSPENDED.name: self.number_of_suspended_projects_to_create,
+            ProjectType.COMPLETED.name: self.number_of_completed_projects_to_create,
+        }
+        self.assertDictEqual(Command(self.options)._set_number_of_projects_to_create(self.options), expected_options)
```
kbeker, line 149: It would be enough to use just `self.assertEqual` instead of `self.assertDictEqual`

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def test_start_date_should_be_of_datetime_type_and_set_in_past(self):
+        start_date_returned_by_function = Command._create_start_date()
+        self.assertTrue(isinstance(start_date_returned_by_function, timezone.datetime))
```
kbeker, line 153: Instead of using `self.assertTrue` to check if inside instance is correct please use just `self.assertIsInstance`

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+        self.assertLess(start_date_returned_by_function, timezone.now())
+
+    def test_stop_date_should_be_of_datetime_type_and_set_in_past(self):
+        stop_date_returned_by_function = Command._create_stop_date()
+        self.assertTrue(isinstance(stop_date_returned_by_function, timezone.datetime))
```
kbeker, line 158: Same here about `assertIsInstance`

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,207 @@
...
+        self.assertLess(stop_date_returned_by_function, timezone.now())
+
+    def test_suspended_projects_should_have_suspended_property(self):
+        self.assertTrue(Command()._set_project_factory_parameters(ProjectType.SUSPENDED.name).get("suspended"))
```
kbeker, line 162: You are using `self.assertTrue` and `self.assertFalse` too often. You should be more precise during testing and check exactly what you want to check.

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done. I got rid of this

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done. I assigned these numbers to variables

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done. I left `self.assertTrue` or `self.assertFalse` only for case when function return bool.

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,259 @@
...
+        self.assertEqual(CustomUser.objects.filter(user_type=user_type).count(), number_of_users_to_create)
+
+    @parameterized.expand(
+        [
+            (CustomUser.UserType.ADMIN.name, 2, 2),
+            (CustomUser.UserType.EMPLOYEE.name, 4, 3),
+            (CustomUser.UserType.MANAGER.name, 2, 1),
+        ]
+    )
+    def test_create_user_function_should_not_create_users_if_number_in_database_is_greater_than_or_equal_to_specified(
```
maciejSamerdak, line 95: Why not including similar test for projects? 🙂

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,259 @@
...
+            number_of_projects_to_fill_database=2, project_type=ProjectType.ACTIVE.name
+        )
+        self.number_of_active_projects_to_create = self.options[ProjectType.ACTIVE.name] - len(self.active_projects)
+
+        self.completed_projects = self._generate_test_projects(
+            number_of_projects_to_fill_database=2, project_type=ProjectType.COMPLETED.name
+        )
+        self.number_of_completed_projects_to_create = self.options[ProjectType.COMPLETED.name] - len(
+            self.completed_projects
+        )
```
maciejSamerdak, line 134: I still think there should be tests for cases of creating projects when the database is empty and not all tests here require presence of projects in the database. Because of that, I would lean towards creating objects only in the few tests that require it. In that case, I would move it to the method to be able to call it conveniently in said tests.

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,259 @@
...
+
+    @parameterized.expand(
+        [(ProjectType.SUSPENDED.name, 2), (ProjectType.ACTIVE.name, 3), (ProjectType.COMPLETED.name, 5)]
+    )
+    def test_number_of_projects_to_create_should_be_a_difference_of_number_of_projects_in_database_and_definite_number(
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,259 @@
...
+        self.assertEqual(GenerateTestDataCommand._set_user_factory_parameters(user_type).get("is_staff"), False)
+
+    @parameterized.expand(
+        [CustomUser.UserType.ADMIN.name, CustomUser.UserType.EMPLOYEE.name, CustomUser.UserType.MANAGER.name]
+    )
+    def test_any_user_should_not_have_superuser_property_except_superuser(self, user_type):
+        self.assertEqual(GenerateTestDataCommand._set_user_factory_parameters(user_type).get("is_superuser"), False)
+
+    def test_superuser_should_have_superuser_property(self):
+        self.assertEqual(GenerateTestDataCommand._set_user_factory_parameters(self.superuser).get("is_superuser"), True)
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,259 @@
...
+        [
+            (CustomUser.UserType.ADMIN.name, 2, 2),
+            (CustomUser.UserType.EMPLOYEE.name, 4, 3),
+            (CustomUser.UserType.MANAGER.name, 2, 1),
+        ]
+    )
+    def test_create_user_function_should_not_create_users_if_number_in_database_is_greater_than_or_equal_to_specified(
+        self, user_type, mock_number_of_users_in_database, number_of_users_to_create
+    ):
+        GenerateTestDataCommand().create_user(user_type, mock_number_of_users_in_database)
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,259 @@
...
+    def test_suspended_projects_should_have_suspended_property(self):
+        self.assertEqual(
+            GenerateTestDataCommand()._set_project_factory_parameters(ProjectType.SUSPENDED.name).get("suspended"), True
+        )
+
+    @parameterized.expand([ProjectType.ACTIVE.name, ProjectType.COMPLETED.name])
+    def test_active_and_completed_projects_should_not_have_suspended_property(self, project_type):
+        self.assertEqual(
+            GenerateTestDataCommand()._set_project_factory_parameters(project_type).get("suspended"), False
+        )
```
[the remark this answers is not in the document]
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
marsolekk: Done

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -0,0 +1,331 @@
...
+    def test_despite_filled_database_result_of_create_user_function_should_be_specified_number_of_users_in_database(
+        self, user_type, number_of_users_to_create
+    ):
+        mock_number_of_users_in_database_less_than_next_request = 2
+
+        self._generate_test_users(mock_number_of_users_in_database_less_than_next_request, user_type)
+
+        GenerateTestDataCommand().create_user(user_type, number_of_users_to_create)
+
+        self.assertEqual(CustomUser.objects.filter(user_type=user_type).count(), number_of_users_to_create)
```
maciejSamerdak, line 103: What is your approach to line separation here? I'm just curious, because I would keep lines 97 and 99 together, treating them both equally as a sort of pre-test setup. What's the idea behind keeping them separated?

maciejSamerdak (approved):
I don't have any issues with the fixes and I haven't found anything more in the rest of the code, so I'd say:
image

Just one minor naming issue worth fixing and it's good to go!

kbeker (approved):
In my opinion it is OK :) Ready to merge

PR #573, opened 2020-08-21
Chore add default values to command generate test data

marsolekk:


sheetstorm/management/commands/constants.py
```diff
@@ -0,0 +1,43 @@
...
+def create_set_of_data_size(data_quantity: int) -> Dict[str, int]:
+    return {
+        SUPERUSER_USER_TYPE: 1,
+        CustomUser.UserType.ADMIN.name: data_quantity,
+        CustomUser.UserType.EMPLOYEE.name: data_quantity,
+        CustomUser.UserType.MANAGER.name: data_quantity,
+        ProjectType.SUSPENDED.name: data_quantity,
+        ProjectType.ACTIVE.name: data_quantity,
+        ProjectType.COMPLETED.name: data_quantity,
+    }
```
kbeker, line 37: That is not what we want to create. We don't want to have in every data set the same amount of projects and users. There should be 4 different datasets where for example in small dataset we will have 5 opened, 1 suspended and 2 completed projects - about users the same situation - 2 admins, 3 managers and 15 employees - something like this :)

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -65,7 +93,7 @@ def _get_user_options(options: OptionsDictType) -> Dict[str, Optional[int]]:
 
     @staticmethod
     def _get_superuser_request(options: Dict[str, Any]) -> bool:
-        is_superuser_create_request = options[superuser_user_type]
+        is_superuser_create_request = options[SUPERUSER_USER_TYPE]
```
kbeker, line 96: Changes like here i.e. changing variables are not related with this commit and should be in different one. Please extract changes for variable name refactoring to separate commit. This will require to amend commit and extract these changes from this commit.

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -186,3 +214,27 @@ def add_arguments(self, parser: Any) -> None:
...
+            dest=TestDataSize.LARGE.name,
+            action="store_true",
+            help="Use prepared large set to generate test data",
+        )
+        parser.add_argument(
+            "--extra_large",
+            dest=TestDataSize.EXTRA_LARGE.name,
+            action="store_true",
+            help="Use prepared extra large set to generate test data",
+        )
```
kbeker, line 240: What you added here require also adding additional tests to ensure that it works:)

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -40,6 +40,39 @@ class Command(BaseCommand):
...
+    @staticmethod
+    def _get_request_to_create_data_using_prepared_set(options: OptionsDictType) -> Any:
+        if options[TestDataSize.SMALL.name] is True:
+            requested_set = SMALL_SET
+        elif options[TestDataSize.MEDIUM.name] is True:
+            requested_set = MEDIUM_SET
+        elif options[TestDataSize.LARGE.name] is True:
+            requested_set = LARGE_SET
+        elif options[TestDataSize.EXTRA_LARGE.name] is True:
+            requested_set = EXTRA_LARGE_SET
```
kbeker, line 69: Please create map of this all variables and move them to constants. It will simplify whole method.

kbeker (changes requested):
It require some work. You also didn't add any test to this. Please add tests and address all my comments

sheetstorm/management/commands/generate_test_data.py
kbeker: These parameters must be validated that they can't be selected together. Please check if there is possibility provided by ArgParser

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -40,6 +37,51 @@ class Command(BaseCommand):
...
+            TestDataSize.LARGE.name,
+            TestDataSize.EXTRA_LARGE.name,
+        ]
+        for set_size in set_sizes_list:
+            if options[set_size]:
+                requested_set = DATA_SETS[set_size]
+                break
+        else:
+            raise NotAnySetRequestedException("Not any set requested")
+        return requested_set
```
kbeker, line 82: This method is too complicated, you need to simplify it. This is wrong that you're iterating here element by element for list of data sizes. Please start from adding init method for all variables and then you won't need `options` here because you all will have accessible from instance

sheetstorm/tests/test_integrity_generate_test_data.py
```diff
@@ -248,3 +250,62 @@ def test_that_passing_text_instead_of_number_for_projects_arguments_should_raise
     def test_that_passing_any_value_for_positional_superuser_argument_should_raise_error(self, test_value):
         with self.assertRaises(CommandError):
             management.call_command("generate_test_data", f"--superuser={test_value}")
+
+    @parameterized.expand([(1,), ("test_text",)])
+    def test_that_passing_any_value_for_positional_set_size_argument_should_raise_error(self, test_value):
```
kbeker, line 255: There should be also test that when you select one of sizes you should be able to chose second

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -186,3 +223,27 @@ def add_arguments(self, parser: Any) -> None:
...
+            help="Use prepared medium set to generate test data",
+        )
+        parser.add_argument(
+            "--large",
+            dest=TestDataSize.LARGE.name,
+            action="store_true",
+            help="Use prepared large set to generate test data",
+        )
+        parser.add_argument(
+            "--extra_large",
```
kbeker, line 245: I didn't noticed it at first time. When you creating parameters should like this which will be used in bash it should be divided by "-" not by "_" because in bash is convention that words should be split with dash

sheetstorm/tests/test_integrity_generate_test_data.py
```diff
@@ -248,3 +250,62 @@ def test_that_passing_text_instead_of_number_for_projects_arguments_should_raise
...
+            DataSize.MEDIUM.name: False,
+            DataSize.LARGE.name: False,
+            DataSize.EXTRA_LARGE.name: False,
+        }
+
+        management.call_command("generate_test_data", **combined_options)
+
+        self.compare_quantity_in_database_with_specified_set()
+
+    def test_that_passing_more_than_one_set_size_argument_should_create_smaller_specified_set(self):
```
kbeker, line 290: This test wrong. This is not expected command behavior. These command parameters should excluded each other

sheetstorm/tests/test_integrity_generate_test_data.py
```diff
@@ -248,3 +250,62 @@ def test_that_passing_text_instead_of_number_for_projects_arguments_should_raise
...
+        management.call_command("generate_test_data", "--small", "--large")
+
+        self.compare_quantity_in_database_with_specified_set()
+
+    def compare_quantity_in_database_with_specified_set(self):
```
kbeker, line 295: You created method here which is named "with_specified_set" but you don't provide any parameter here and hardcoded just small size. You need to chose whether you want to specify or hardcode :)

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -327,3 +368,56 @@ def _compute_number_of_projects_to_create(options, projects_in_database, project
...
+        options = {
+            DataSize.SMALL.name: True,
+            DataSize.MEDIUM.name: False,
+            DataSize.LARGE.name: False,
+            DataSize.EXTRA_LARGE.name: True,
+        }
+
+        self.assertTrue(GenerateTestDataCommand._get_request_to_create_data_using_prepared_set(options))
+
+    def test_get_request_to_create_data_using_prepared_set_function_should_return_True_if_not_any_set_requested(self):
```
kbeker, line 384: "...should_return_True_if_not_any_set_requested"? 🤔  It should be rephrased to something like "...should_return_false_if_paramter_is_not_provided"

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -327,3 +368,56 @@ def _compute_number_of_projects_to_create(options, projects_in_database, project
...
+
+    def test_pick_dataset_to_create_function_should_return_one_specified_set_if_there_is_request(self):
+        options = {
+            DataSize.SMALL.name: True,
+            DataSize.MEDIUM.name: False,
+            DataSize.LARGE.name: False,
+            DataSize.EXTRA_LARGE.name: False,
+        }
+
+        self.assertEqual(GenerateTestDataCommand()._pick_dataset_to_create(options), DATA_SETS[DataSize.SMALL.name])
```
kbeker, line 402: This should be done by parameterized test and check all possibilities

kbeker (changes requested):
There are still few things to fix. If you will have any question to how handle this comments please contact me :)

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -327,3 +429,86 @@ def _compute_number_of_projects_to_create(options, projects_in_database, project
...
+        self.assertEqual(GenerateTestDataCommand()._pick_dataset_to_create(options), DATA_SETS[set_name])
+
+    def test_pick_dataset_to_create_function_should_create_smaller_set_in_case_of_allow_to_specify_more_than_one(self):
+        options = {
+            **self.unrelated_options,
+            DataSize.SMALL.name: True,
+            DataSize.MEDIUM.name: False,
+            DataSize.LARGE.name: True,
+            DataSize.EXTRA_LARGE.name: True,
+        }
```
maciejSamerdak, line 500: Since it is currently impossible to specify more than one option, I don't think this should be tested, because right now it seems more like a side effect feature rather than a conscious design. 😉 

But that's more of my personal opinion, so I will leave the decisive voice to @kbeker

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -21,6 +24,13 @@ def test_get_users_options_function_should_return_only_users_options(self):
             CustomUser.UserType.EMPLOYEE.name: None,
             CustomUser.UserType.MANAGER.name: 2,
             self.superuser: False,
+            ProjectType.SUSPENDED.name: None,
+            ProjectType.ACTIVE.name: None,
+            ProjectType.COMPLETED.name: None,
+            DataSize.SMALL.name: False,
+            DataSize.MEDIUM.name: False,
+            DataSize.LARGE.name: False,
+            DataSize.EXTRA_LARGE.name: False,
```
maciejSamerdak, line 33: Okay, so now you have what looks like a ton of options unrelated to your tests, but must be present. This is a perfect moment to move it to the **setUp**. Either declare **mock_options** there and alter necessary values individually in tests or put these values in a separate dicts and concatenate it  with mock_options.

sheetstorm/tests/test_integrity_generate_test_data.py
```diff
@@ -248,3 +250,61 @@ def test_that_passing_text_instead_of_number_for_projects_arguments_should_raise
...
+            management.call_command("generate_test_data", "--small", "--large")
+
+
+class CreateDataFromPreparedSetTests(TestCase):
+    SUPERUSER = "SUPERUSER"
+
+    def test_that_command_should_create_specified_set_when_there_is_request(self):
+        management.call_command("generate_test_data", "--small")
+
+        self.compare_quantity_in_database_with_small_set()
```
maciejSamerdak, line 273: I think it's worth expanding to cover each of the data sets.

sheetstorm/tests/test_integrity_generate_test_data.py
```diff
@@ -248,3 +250,61 @@ def test_that_passing_text_instead_of_number_for_projects_arguments_should_raise
...
+            ProjectType.COMPLETED.name: 1,
+            DataSize.SMALL.name: True,
+            DataSize.MEDIUM.name: False,
+            DataSize.LARGE.name: False,
+            DataSize.EXTRA_LARGE.name: False,
+        }
+
+        management.call_command("generate_test_data", **combined_options)
+
+        self.compare_quantity_in_database_with_small_set()
```
maciejSamerdak, line 292: I think it's crucial to see what happens if at least one of the set of options actually **exceeds** the number included in the set to be certain that it's working properly.

It'll be enough for me if you change one of the users' and one of the projects' option to reflect that.

sheetstorm/tests/test_integrity_generate_test_data.py
```diff
@@ -248,3 +250,61 @@ def test_that_passing_text_instead_of_number_for_projects_arguments_should_raise
...
+            CustomUser.UserType.EMPLOYEE.name: None,
+            CustomUser.UserType.MANAGER.name: None,
+            self.SUPERUSER: False,
+            ProjectType.SUSPENDED.name: 100,
+            ProjectType.ACTIVE.name: None,
+            ProjectType.COMPLETED.name: None,
+            DataSize.SMALL.name: True,
+            DataSize.MEDIUM.name: False,
+            DataSize.LARGE.name: False,
+            DataSize.EXTRA_LARGE.name: False,
```
maciejSamerdak, line 287: Let's add some variety here: you have 3 types for both users and projects, so let's take an advantage of that - make it a larger number than set, a smaller number than set and None. 😉

maciejSamerdak (changes requested):
There is a major change that needs to be made regarding usage of data sets - the command should have a singular parameter rather than four separate flags.

I didn't point it out previously, because I thought there might be a conscious design choice behind it, but yesterday I had an opportunity to discuss it with **Remik** and **Karol** and we're all on the same page here - **data set option should be in a single parameter**, like `--data-size s/m/l/xl` for example.

It makes much more sense logically and programatically, since we allow only one to be picked, therefore the set of flags must go. Remik made a great analogy to a **radio button** and a **checkbox** - parameter corresponds here with the first and flags with the latter.

This might take some effort to adjust the code, but it will also further simplify logic in your code, so it's a win-win 👍 .

sheetstorm/tests/test_unit_generate_test_data.py
```diff
@@ -327,3 +377,43 @@ def _compute_number_of_projects_to_create(options, projects_in_database, project
...
+            (False, False, False, True, DataSize.EXTRA_LARGE.name),
+        ]
+    )
+    def test_pick_dataset_to_create_function_should_return_one_specified_set_if_there_is_request(
+        self, small_requested, medium_requested, large_requested, extra_large_requested, set_name
+    ):
+        GenerateTestDataCommand.is_small_set_request = small_requested
+        GenerateTestDataCommand.is_medium_set_request = medium_requested
+        GenerateTestDataCommand.is_large_set_request = large_requested
+        GenerateTestDataCommand.is_extra_large_set_request = extra_large_requested
```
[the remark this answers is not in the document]
marsolekk: Done. I changed it everywhere I found

sheetstorm/tests/test_integrity_generate_test_data.py
marsolekk: Done.

marsolekk:
> There is a major change that needs to be made regarding usage of data sets - the command should have a singular parameter rather than four separate flags.
> 
> I didn't point it out previously, because I thought there might be a conscious design choice behind it, but yesterday I had an opportunity to discuss it with **Remik** and **Karol** and we're all on the same page here - **data set option should be in a single parameter**, like `--data-size s/m/l/xl` for example.
> 
> It makes much more sense logically and programatically, since we allow only one to be picked, therefore the set of flags must go. Remik made a great analogy to a **radio button** and a **checkbox** - parameter corresponds here with the first and flags with the latter.
> 
> This might take some effort to adjust the code, but it will also further simplify logic in your code, so it's a win-win +1 .

Done. I got rid of flags that you mentioned and add single parameter for data set size. I hope I applied it correctly

kbeker (approved):

kbeker (approved):

kbeker (approved):

PR #579, opened 2021-02-15
Feature report list view

marsolekk:


employees/static/employees/popup_style.css
```diff
@@ -1,29 +1,25 @@
...
     width: 40px;
     height: 30px;
     top: 35%;
     background:white;
-    border: 2px solid rgb(230, 230, 230);
-}
-
-.ui-dialog .ui-dialog-titlebar-close:hover {
-    background-color: rgb(230, 230, 230);
-    border: 2px solid rgb(173, 173, 173);
```
marsolekk, line 28: Is that file required after implementing django-bootstrap-modal-forms?

employees/views.py
```diff
@@ -157,22 +159,9 @@ def redirect_to_current_month(self) -> HttpResponseRedirectBase:
...
-@method_decorator(
-    check_permissions(
-        allowed_user_types=[
-            CustomUser.UserType.EMPLOYEE.name,
-            CustomUser.UserType.MANAGER.name,
-            CustomUser.UserType.ADMIN.name,
-        ]
-    ),
-    name="dispatch",
-)
```
kbeker, line 170: What about these permission decorators? Without it somebody else can access your reports

employees/tests/test_views.py
```diff
@@ -939,3 +838,157 @@ def _get_total_hours_per_project_and_percentage_from_month(self):
...
+            "project": self.report.project.pk,
+            "author": self.user.pk,
+            "work_hours": "8:00",
+            "task_activities": self.task_activity.pk,
+        }
+
+    def test_create_report_view_should_add_new_report_on_post_when_form_is_valid(self):
+        response = self.client.post(self.url, self.data)
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(Report.objects.all().count(), 2)
```
kbeker, line 883: If you're checking that there should be 2 reports you should also be sure that there was only one before whole operation. Please add this check at the beginning of test

employees/tests/test_views.py
```diff
@@ -939,3 +838,157 @@ def _get_total_hours_per_project_and_percentage_from_month(self):
...
+            "description": "",
+            "project": self.report.project.pk,
+            "author": self.user.pk,
+            "work_hours": "8:00",
+            "task_activities": self.task_activity.pk,
+        }
+
+    def test_create_report_view_should_add_new_report_on_post_when_form_is_valid(self):
+        response = self.client.post(self.url, self.data)
+        self.assertEqual(response.status_code, 302)
```
kbeker, line 882: Please check also to where 302 try to redirect and also what is the status code of endpoint to where we were redirected

employees/tests/test_views.py
```diff
@@ -939,3 +838,157 @@ def _get_total_hours_per_project_and_percentage_from_month(self):
...
+        self.assertEqual(Report.objects.all().count(), 2)
+        self.assertEqual(response.url, self.success_url)
+
+    def test_create_report_view_should_not_redirect_and_create_report_on_submit_when_form_is_invalid(self):
+        response = self.client.post(self.url, self.invalid_data)
+        self.assertEqual(Report.objects.all().count(), 1)
+        self.assertEqual(response.status_code, 200)
```
kbeker, line 889: That is so strange :D It returns 200 when form is invalid and 302 when form is correct? It should return 400 when form is invalid and 201 if object is created

employees/tests/test_views.py
```diff
@@ -939,3 +838,157 @@ def _get_total_hours_per_project_and_percentage_from_month(self):
...
+
+    def test_create_report_form_view_should_not_add_report_when_task_activity_is_not_related_not_project(self):
+        not_related_to_project_task_activity = TaskActivityTypeFactory()
+        author_reports_before_post = self.user.report_set.all()
+
+        self.data["task_activities"] = not_related_to_project_task_activity.pk
+
+        response = self.client.post(self.url, self.data)
+
+        self.assertEqual(response.status_code, 200)
```
kbeker, line 919: Again - 200 is not a error code. Please fix that

employees/tests/test_views.py
```diff
@@ -939,3 +838,157 @@ def _get_total_hours_per_project_and_percentage_from_month(self):
...
+        )
+        self.success_url = reverse(
+            "custom-report-list",
+            kwargs={"year": datetime.datetime.now().date().year, "month": datetime.datetime.now().date().month},
+        )
+
+    def test_join_project_view_should_add_user_to_project_selected_in_project_join_form_on_post(self):
+        new_project = ProjectFactory(name="New Project", start_date=datetime.datetime.now())
+        response = self.client.post(path=self.url, data={"projects": new_project.id})
+        self.assertEqual(response.status_code, 302)
```
kbeker, line 983: It should be 201 or 200 :)

kbeker (changes requested):
There are some things which need to be fixed. I know that those status codes are probably inside new library but please fix them because this is bug that when form is invalid it returns 2** codes and when form is correct it gives 3**. Second one I am able to understand because it might want to redirect to report list but in a case with invalid form we should return 400

employees/tests/test_views.py
marsolekk: Done. I've added new test case, with `follow=True` on post.

employees/tests/test_views.py
marsolekk: Done

employees/tests/test_views.py
marsolekk: We’ve discussed it. I believe that this is proper behavior in this case.

employees/tests/test_views.py
marsolekk: It’s also already discussed.

employees/tests/test_views.py
marsolekk: Again - I believe that this is proper behavior in this case.

marsolekk (commented):
I think that I addressed all of your comments. 

- I left status codes as is, according to what we've discussed. 
- Also, there is a new file openSmallModals.js for opening "small modals" - join project and delete report modals

employees/common/strings.py
```diff
@@ -19,7 +19,7 @@ class ReportListStrings(NotCallableMixin, Enum):
     WORK_HOURS_COLUMN_HEADER = _("Work hours")
     TASK_ACTIVITIES_COLUMN_HEADER = _("Task Activity")
     DESCRIPTION_COLUMN_HEADER = _("Description")
-    HOURS_PER_DAY_LABEL = _("Daily hours")
+    HOURS_PER_DAY_LABEL = _("All projects")
```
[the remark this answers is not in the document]
marsolekk: Done

employees/templates/employees/partial/modal_forms/create_and_update_report.html
```diff
@@ -0,0 +1,162 @@
...
+<script type="text/javascript">
+    $(function () {
+
+        $("#delete-report").modalForm({
+            formURL: "{% url 'custom-report-delete' report.pk %}"
+        });
+
+    });
+</script>
+{% endif %}
```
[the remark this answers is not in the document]
marsolekk: Done. I've created new file `openSmallModals.js` for that.

employees/static/employees/scripts/expandReportList.js
```diff
@@ -0,0 +1,38 @@
...
+};
+
+function calculateReportListHeight(currentRow, numberOfReports) {
+    let reportListHeight = currentRow.offsetHeight;
+    let nextRow = currentRow.nextElementSibling;
+
+    for (var reportNumber=1; reportNumber < numberOfReports; reportNumber++) {
+        reportListHeight = reportListHeight + nextRow.offsetHeight;
+        nextRow = nextRow.nextElementSibling;
+
```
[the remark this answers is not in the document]
marsolekk: Done

employees/templates/employees/report_list.html
```diff
@@ -49,34 +51,25 @@
...
-    {% endif %}
+
+    <script type="text/javascript">
+        $(function () {
+             $("#opener_create").modalForm({
+                    modalID: "#modal",
+                    formURL: "{% url 'create-report' year month %}"
+                });
+        });
+    </script>
```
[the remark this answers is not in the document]
marsolekk: Done. I moved it to `modalsFunctionality.js` file

employees/tests/test_views.py
```diff
@@ -1,9 +1,11 @@
+# pylint: disable=C0302
```
kbeker, line 1: That's actually good warning. Too big files are hard to be maintainable. Please create test file with more meaningful name then just `test_views`

kbeker (approved):

PR #580, opened 2021-03-09
Feature update employees list view

marsolekk:


users/static/users/style.css
```diff
@@ -1,72 +1,458 @@
...
+    height: 35px;
+    width: 70px;
+}
+
+.disabled-link{
+    cursor: default;
+    pointer-events: none;
+    text-decoration: none;
+}
+/* start style for slider */
```
marsolekk, line 366: Remove comments if unnecessary.

users/templates/base.html
```diff
@@ -141,10 +141,13 @@
...
+    <script src="https://code.jquery.com/ui/1.11.3/jquery-ui.min.js"></script>
     <script
         src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js"
         integrity="sha384-Tc5IQib027qvyjSMfHjOMaLkfuWVxZxUPnCJA7l2mCWNIpG9mGCD8wGNIcPD7Txa"
         crossorigin="anonymous"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/2.6.0/umd/popper.min.js" integrity="sha512-BmM0/BQlqh02wuK5Gz9yrbe7VyIVwOzD1o40yi1IsTjriX/NGF37NyXHfmFzIlMmoSIBXgqDiG1VNU6kB5dBbA==" crossorigin="anonymous"></script>
+    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js" integrity="sha384-Tc5IQib027qvyjSMfHjOMaLkfuWVxZxUPnCJA7l2mCWNIpG9mGCD8wGNIcPD7Txa" crossorigin="anonymous"></script>
     <script
         src="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.1/js/all.min.js"
         integrity="sha512-F5QTlBqZlvuBEs9LQPqc1iZv2UMxcVXezbHzomzS6Df4MZMClge/8+gXrKw2fl5ysdk4rWjR0vKS7NNkfymaBQ=="
```
marsolekk, line 153: This may/should be in a separate/different commit.

kbeker:
Please take a look on this because it sometimes fails.
```
_______________________________________________________________________________ TestProjectQuerySet.test_filter_active_projects_should_not_return_project_with_stop_date_less_than_current_date ________________________________________________________________________________

self = <managers.tests.test_unit_model.TestProjectQuerySet testMethod=test_filter_active_projects_should_not_return_project_with_stop_date_less_than_current_date>

    @freeze_time("2019-08-28")
    def test_filter_active_projects_should_not_return_project_with_stop_date_less_than_current_date(self):
        self.project.stop_date = timezone.now() - timezone.timedelta(days=1)
>       self.project.full_clean()

managers/tests/test_unit_model.py:131: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Project: System data article.>, exclude = [], validate_unique = True

    def full_clean(self, exclude=None, validate_unique=True):
        """
        Call clean_fields(), clean(), and validate_unique() on the model.
        Raise a ValidationError for any errors that occur.
        """
        errors = {}
        if exclude is None:
            exclude = []
        else:
            exclude = list(exclude)
    
        try:
            self.clean_fields(exclude=exclude)
        except ValidationError as e:
            errors = e.update_error_dict(errors)
    
        # Form.clean() is run even if other validation fails, so do the
        # same with Model.clean() for consistency.
        try:
            self.clean()
        except ValidationError as e:
            errors = e.update_error_dict(errors)
    
        # Run unique checks, but only for fields that passed validation.
        if validate_unique:
            for name in errors:
                if name != NON_FIELD_ERRORS and name not in exclude:
                    exclude.append(name)
            try:
                self.validate_unique(exclude=exclude)
            except ValidationError as e:
                errors = e.update_error_dict(errors)
    
        if errors:
>           raise ValidationError(errors)
E           django.core.exceptions.ValidationError: {'__all__': ['A project can not be created after expired date!']}

../../../../.local/share/virtualenvs/sheetstorm-Y5voKRoL/lib/python3.7/site-packages/django/db/models/base.py:1222: ValidationError
```

users/views.py
```diff
@@ -125,29 +135,41 @@ def form_valid(self, form: SimpleUserChangeForm) -> HttpResponse:
...
     model = CustomUser
-    queryset = (
-        CustomUser.objects.active()
+    queryset_active = (
+        CustomUser.objects.filter(is_active=True)
+        .order_by("user_type", "last_name", "first_name", "email")
+        .prefetch_related("projects")
+    )
+    queryset_disabled = (
+        CustomUser.objects.filter(is_active=False)
```
kbeker, line 172: Instead of two database calls you can do here only one:
``python
ordered_users_queryset = (
    CustomUser.objects
    .order_by("user_type", "last_name", "first_name", "email")
    .prefetch_related("projects")
)

disabled_users_queryset = ordered_users_queryset.filter(is_active=False)
active_users_queryset = ordered_users_queryset.filter(is_active=True)
```
It should only once call database and also should be properly ordered. Please ensure that proper ordering is done in this case.

File: users/views.py [hunk 4]
```diff
@@ -156,18 +178,44 @@ def get_context_data(self, *, _object_list: Any = None, **kwargs: Any) -> dict:
...
+        return redirect("custom-users-list")
+
+    def _manage_user_status(self, user_id: int) -> None:
+        self._reverse_user_status(CustomUser.objects.get(id=user_id))
+
+    @staticmethod
+    def _reverse_user_status(user: CustomUser) -> CustomUser:
+        user_status = user.is_active
+
+        user.is_active = not user_status
```
kbeker, line 214: Why not just:
```python
user.is_active = not user.is_active
```
?

users/views.py
```diff
@@ -156,18 +178,44 @@ def get_context_data(self, *, _object_list: Any = None, **kwargs: Any) -> dict:
...
+        user.save()
```
kbeker, line 215: Missing `full_clean` method call

users/views.py
```diff
@@ -156,18 +178,44 @@ def get_context_data(self, *, _object_list: Any = None, **kwargs: Any) -> dict:
...
+
+        if user_id_to_disable is not None:
+            self._manage_user_status(user_id_to_disable)
+        elif user_id_to_activate is not None:
+            self._manage_user_status(user_id_to_activate)
+
+        return redirect("custom-users-list")
+
+    def _manage_user_status(self, user_id: int) -> None:
+        self._reverse_user_status(CustomUser.objects.get(id=user_id))
```
kbeker, line 208: It seems llike this method is not needed at all. You can move quering user to `_reverse_user_status` method

users/forms.py
```diff
@@ -13,15 +15,26 @@
...
+    )
 
     class Meta:
         model = CustomUser
-        fields = ("email",)
+        fields = ("email", "first_name", "last_name", "user_type")
+        labels = {"email": "", "first_name": "", "last_name": "", "user_type": ""}
+
+    def save(self, commit: bool = False) -> object:
+        # pylint: disable=E1003
```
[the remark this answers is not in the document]
marsolekk: If I try to change the order of inheritance I receive

> Error: Cannot create a consistent method resolution order (MRO). 

 
 I can omit this anyway, but still `save()` method needs to be implemented here, to make it possible to create a user.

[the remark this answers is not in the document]
marsolekk: I've changed error code to `# pylint: disable=bad-super-call`

users/views.py
marsolekk: Done

users/views.py
marsolekk: Good question. Probably I've had a different concept here, but I decided to change approach. I've fixed that.

users/views.py
marsolekk: Done

users/views.py
marsolekk: Done

users/static/users/style.css
marsolekk: Done

users/forms.py
```diff
@@ -13,15 +15,26 @@
...
-class CustomUserCreationForm(UserCreationForm):
-    """
-    A form that creates a user, without privileges,
-    from given email and password.
-    """
+class CustomUserCreationForm(UserCreationForm, BSModalModelForm, CreateUpdateAjaxMixin):
+    password1 = forms.CharField(label="", widget=forms.PasswordInput)
+    password2 = forms.CharField(
+        label="", widget=forms.PasswordInput, help_text="Enter the same password as above, for verification."
+    )
```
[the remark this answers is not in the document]
marsolekk: I removed this field from this form, but they are obviously still implemented due to inheritance from `UserCreationForm`.

kbeker (approved):
Everything besides this `save()` method looks fine. Let's discuss it this tomorrow on stand up

PR #589, opened 2021-05-13
Bugfix application admin has access to django admin panel

Seveex:


marsolekk (commented):
Overall okay, nothing to change in code for me. :+1: 

What you need to change for sure is the name of Pull Request. Please take a look at how names of other pull requests were created. You will notice that we always start with `Feature/Bugfix/Chore` and then take the rest of the branch name, so for this case it should be "Bugfix admin has access to django admin panel". It is a part of convention, therefore the name should be changed. 

Also another small issue is that a name of the branch may be a little bit more specific like `bugfix/application-admin-has-access-to-django-admin-panel` or maybe name that will indicate that this issue occurred only for admin generated by custom command - not in the whole application. I am not 100% convinced about that so let's also wait for Karol's (@kbeker) opinion if that change is necessary.

kbeker:
I agree with all what @marsolekk said. Please follow his instructions because code looks good

PR #593, opened 2021-05-20
Bugfix account details update by admin

marsolekk:
I've created base class in forms for updating user by admin, from which inherits forms used in a specified views. The problem was with ajax mixins as the same form has been used for updating users info by admin (django-bootstrap-modal-forms used here) and to update admin's account details. After fix this problem should be resolved. 

Also I've removed popup message from account details view after updating other user by admin.


kbeker (approved):
Nice work, ready to merge

PR #594, opened 2021-05-24
Chore add full clean to save method in models

marsolekk:


utils/models.py
```diff
@@ -0,0 +1,12 @@
...
+from django.db import models
+
+
+class BaseModel(models.Model):
+    class Meta:
+        abstract = True
+
+    def save(self, *args: Any, **kwargs: Any) -> None:  # pylint: disable=signature-differs
+        self.full_clean()
+        super(BaseModel, self).save(*args, **kwargs)
```
kbeker, line 12: I'm sure we need to update pylint because it also should be raised. Please not use python 2 inheritance method. Please use just `super().save(*args, **kwargs)`

users/models.py
```diff
@@ -63,7 +64,7 @@ def create_superuser(self, email: str, password: str) -> "CustomUser":
         )
 
 
-class CustomUser(AbstractBaseUser, PermissionsMixin):
+class CustomUser(AbstractBaseUser, PermissionsMixin, BaseModel):
```
kbeker, line 67: I'm not sure if new model class shouldn't be as a last to inherit from class because inside of other classes there might be `save` method overridden somewhere. But If in this PR I will find test for this that `full_clean` is called this comment will be invalid

users/factories.py
```diff
@@ -14,14 +17,7 @@ class Meta:
...
     is_active = True
-
-    @factory.post_generation
-    def set_password(  # type: ignore
-        self, create, extracted, **kwargs
-    ):  # pylint: disable=unused-argument, no-value-for-parameter
-        self.set_password("password")
-        if create:
-            self.save()
+    password = make_password(FACTORY_USER_PASSWORD)
```
kbeker, line 20: Told ya :D

users/tests/test_unit_customuser_model.py
```diff
@@ -64,6 +64,11 @@ def test_user_queryset_active_method_should_return_queryset_of_only_active_users
         self.assertTrue(active_user in queryset)
         self.assertFalse(inactive_user in queryset)
 
+    def test_full_clean_method_should_be_called_on_save(self):
+        user = CustomUser()
+        with self.assertRaises(ValidationError):
+            user.save()
```
kbeker, line 70: What you test here is only that `save` method raise `ValidationError`. You haven't test that `full_clean` method is called. Please add `patch` for `full_clean` method and test that method was called ideally once without any arguments.

utils/models.py
```diff
@@ -0,0 +1,12 @@
+from typing import Any
+
+from django.db import models
+
+
+class BaseModel(models.Model):
+    class Meta:
+        abstract = True
+
+    def save(self, *args: Any, **kwargs: Any) -> None:  # pylint: disable=signature-differs
```
kbeker, line 10: If signature is different than it is originally please fix it. As far as I remember in `save` method signature was also `full_clean` parameter.

kbeker (changes requested):
Few things need to be improved

users/models.py
marsolekk: Do you think that test that I've added is sufficient to test that?

users/factories.py
marsolekk: You were right :D

users/tests/test_unit_customuser_model.py
marsolekk: Done. I use mock and check `assert_called_once()` now.

utils/models.py
marsolekk: Done. I based my previous solution on [this](https://github.com/PyCQA/pylint-django/issues/94) for the same issue. I found some voices that this shouldn't be raised by pylint. Also if I'm not wrong in the future version of pylint this won't be raised if `*args` and `**kwargs` are passed.

utils/models.py
marsolekk: Done

utils/models.py
kbeker: I'm not sure if pylint won't raise problem about it because it's quite important to always have same unchanged signature because in python we don't recognize different method based on signature as it is done for example in c++ or java

employees/tests/test_unit_report_model.py
```diff
@@ -84,6 +82,13 @@ def test_report_model_editable_field_should_have_default_value(self):
     def test_report_model_last_update_field_should_be_changed_on_update(self):
         self.field_auto_now_test("last_update", "description", "Updated")
 
+    @mock.patch("employees.models.Report")
+    def test_full_clean_method_should_be_called_on_save(self, mocked_model):
```
kbeker, line 86: You shouldn't mock whole Report and you don't really need to write test for every class which will use new model once it only inherent from only this class. :)

kbeker: I mean it's not bad, but it actually is unnecessary :P If you would test only once if it's working it would be fine. The only thing is that you're patch every time exact model instead of `full_clean` method. Please change it to mock `full_clean` ;)

users/models.py
kbeker: Yea, but please follow things which I wrote about test itself that you should mock `full_clean` method instead of whole model`

utils/models.py
```diff
@@ -0,0 +1,14 @@
...
+
+class BaseModel(models.Model):
+    class Meta:
+        abstract = True
+
+    def save(
+        self, force_insert: bool = False, force_update: bool = False, using: Any = None, update_fields: Any = None
+    ) -> None:
+        self.full_clean()
+        super().save()
```
kbeker, line 14: You need to pass all the `save` parameters because right now if anything use them you just don't allow them to use them correctly

kbeker (changes requested):
There are still few things to be improved

kbeker (approved):

PR #603, opened 2021-06-02
Bugfix admin not able to create user

marsolekk:


users/models.py
```diff
@@ -206,3 +206,11 @@ def update_remove_inactive_user_from_projects(sender: "CustomUser", **kwargs: An
     if not user.is_active:
         user.projects.clear()
         logger.debug(f"User: {user.pk} has been removed from all projects")
+
+
+@receiver(post_save, sender=CustomUser)
+def create_custom_user_preferences(sender: "CustomUser", created: bool, **kwargs: Any) -> None:
```
kbeker, line 207: If you're using `CustomUser` class above you don't need to add typing in quotes here

users/tests/test_views.py
```diff
@@ -188,6 +188,20 @@ def test_user_create_view_should_not_add_new_user_on_post_if_form_is_invalid(sel
...
 
+    def test_user_create_view_should_add_new_user_on_post_able_to_login_if_form_is_valid(self):
+        self.client.post(path=self.url, data=self.valid_data)
+
+        new_user = CustomUser.objects.get(email=self.new_user_email)
+        new_user.is_active = True
+        new_user.save()
+
+        response = self.client.post(
+            path="/users/accounts/login/?next=/",
```
kbeker, line 199: Please instead of hardcoding string path value use `reverse` django function to call proper endpoint

users/tests/test_unit_customuser_model.py
```diff
@@ -71,7 +71,7 @@ def test_full_clean_method_should_be_called_on_save(mocked_full_clean):
         user = CustomUser()
         user.save()
 
-        mocked_full_clean.assert_called_once()
+        mocked_full_clean.assert_called()
```
kbeker, line 74: How many times now `full_clean` is called that you changed this? :P

users/models.py
```diff
@@ -147,15 +147,10 @@ def email_user(self, subject: str, message: str, from_email: str = None) -> None
 
     def activate_user(self) -> None:
         """
-        Activate and create a CustomUserPreferences object for the user.
-        Should be used only as initial activation for newly creating users.
+        Activate user.
```
kbeker, line 150: You can delete this docs string completely

users/admin.py
```diff
@@ -20,7 +20,7 @@ class CustomUserAdmin(UserAdmin):
     )
     add_fieldsets = ((None, {"classes": ("wide"), "fields": ("email", "password1", "password2")}),)
     form = CustomUserChangeForm
-    add_form = CustomUserCreationForm
+    add_form = UserCreationForm
```
kbeker, line 23: When you're doing changes like this please always add small description to commit because it's hard now to know what you wanted to do here.

kbeker (changes requested):
Few things need to be polished

users/tests/test_unit_customuser_model.py
marsolekk: `CustomUserPreferences` inherits from `BaseModel` (abstract model) now. If I am not wrong, as we added signals to create object of `CustomUserPreferences` it will also call `full_clean()` for that creation.

users/models.py
marsolekk: Done

users/models.py
marsolekk: Done

users/tests/test_views.py
marsolekk: Done

users/admin.py
marsolekk: Done. Please check that this description is sufficient.

kbeker (approved):
Ready to merge

PR #605, opened 2021-06-07
Bugfix previous password can be new password

marsolekk:
It should be treated as "draft" at this moment due to an unresolved issue with typing in the `validate()` method of `ReusedPreviousPasswordValidator` class. I've added type: ignore comment to prevent mypy raise error. I'll look into it once more later, but I didn't find a simple, sufficient solution for that yet.


marsolekk:
I've implemented `ReusedPreviousPasswordValidator` that checks previous(current) user password and raise ValidationError if is the same as typed.

users/tests/test_views.py
```diff
@@ -496,3 +497,32 @@ def test_menu_expansion_update_should_update_menu_expansion_setting_in_user_pref
...
+        self.user.set_password(self.user_password)
+        self.user.save()
+
+        reset_password_response = self.client.post(reverse("password_reset"), data={"email": self.user.email})
+        self.uid = reset_password_response.context["uid"]
+        token = reset_password_response.context["token"]
+        reset_password_confirm_response = self.client.post(
+            reverse("password_reset_confirm", kwargs={"uidb64": self.uid, "token": token})
+        )
+        self.url = reset_password_confirm_response.url
```
kbeker, line 515: It's not a best place for such things. You should create separate method in this test class and just call method with these API calls in this method. `setUp` method should be used only for variables initiation. We're losing some context when you adding such things in `setUp`. Also at least you should such logic move to separate method.

users/tests/test_unit_validators.py
```diff
@@ -81,3 +83,19 @@ def test_validator_should_pass_when_name_contain_dash(self):
...
+        self.password = "Password1234!"
+        self.user.set_password(self.password)
+        self.user.save()
+
+    def test_that_validator_raise_exception_when_new_password_is_the_same_as_previous_password(self):
+        with self.assertRaises(ValidationError) as exception:
+            self.password_validator.validate(password=self.password, user=self.user)
+        self.assertEqual(
+            exception.exception.messages, [ValidationErrorText.VALIDATION_ERROR_NEW_PASSWORD_THE_SAME_AS_PREVIOUS]
+        )
```
kbeker, line 101: You should have at least one test that checks if different password pass as assumed.

kbeker (changes requested):
Please fix tests. Code logic looks good

users/tests/test_views.py
marsolekk: Done

users/tests/test_unit_validators.py
marsolekk: Done

users/tests/test_views.py
```diff
@@ -557,3 +558,36 @@ def test_menu_expansion_update_should_update_menu_expansion_setting_in_user_pref
...
         self.assertEqual(user_menu_expansion_setting_after_update, not expanded_menu)
+
+
+class ResetPasswordTests(TestCase):
+    def setUp(self) -> None:
+        self.user_password = "Password1234!"
+        self.user = UserFactory()
+        self.user.set_password(self.user_password)
+        self.user.save()
+
```
kbeker, line 569: Why do you left here blank space? ;)

kbeker (approved):
Please just fix this one comment and ready to merge

users/tests/test_views.py
marsolekk: Hmmm... good question. This is my bad. Sometimes I put blank space unnecessarily.

marsolekk: Done

PR #607, opened 2021-06-09
Bugfix possibility to delete user

marsolekk:


users/models.py
```diff
@@ -35,6 +35,9 @@ def get_with_prefetched_reports(self, reports: QuerySet) -> QuerySet:
     def active(self) -> QuerySet:
         return self.filter(is_active=True)
 
+    def delete(self) -> None:
+        self.update(is_active=False)
```
kbeker, line 39: Can you please move all things from `CustomUserQuerySet` to `CustomUserManager` and then in `CustomUser` model just just:
```python
objects = CustomUserManager()
```

After this please delete `CustomUserQuerySet`

kbeker (commented):
Code itself looks good, but please fix what I found in area of code which you  were changing.

users/models.py
marsolekk: I tried to do that, but I only found solution for that using `CustomUserQuerySet.as_manager()`. Even after moving all methods to this class I need to also override `delete()` method in `CustomUser` class. I didn't implemented it yet, so I will work on it later if needed.

kbeker (approved):

PR #611, opened 2021-06-15
Feature change project list view

marsolekk:


managers/views.py
```diff
@@ -74,6 +75,24 @@ def get_queryset(self) -> QuerySet:
...
+
+        return redirect("custom-projects-list")
+
+    @staticmethod
+    def _manage_project_status(project_id: int) -> Project:
+        project = Project.objects.get(id=project_id)
+        project_status = project.suspended
+
+        project.suspended = not project_status
+        project.save()
```
kbeker: Please add `update_fields` parameter to save method and add only those fields which you're updating here

kbeker (approved):
Please fix just one thing which I mentioned and ready to merge

managers/views.py
Merrique: Added

PR #616, opened 2021-07-26
Feature create billing models

Seveex:


billing/models.py
```diff
@@ -1,3 +1,91 @@
...
+    currency = models.CharField(
+        max_length=BillingConstants.CURRENCY_MAX_LENGTH.value, choices=Currency.choices(), default=Currency.PLN
+    )
+    accounting_method = models.CharField(
+        max_length=BillingConstants.ACCOUNTING_METHOD_MAX_LENGTH.value,
+        choices=AccountingMethod.choices(),
+        default=AccountingMethod.HOURLY.name,
+    )
+    from_date = models.DateTimeField(
+        BillingDetailsAbstractModelText.FROM_DATE, auto_now_add=False, blank=False, null=False
```
kbeker, line 47: Do we need to add `auto_now_add=False`? Shouldn't it be `False` by default?

billing/models.py
```diff
@@ -1,3 +1,91 @@
...
+    )
+    to_date = models.DateTimeField(BillingDetailsAbstractModelText.TO_DATE, auto_now_add=False, blank=True, null=True)
```
kbeker, line 49: Same here about `auto_now_add=False`

billing/models.py
```diff
@@ -1,3 +1,91 @@
...
+
+    @property
+    def get_currency_symbol(self) -> str:
+        return CurrencyConstants.CurrencyDict[self.currency]["Symbol"]
+
+
+class UserSalaryDetails(BillingDetailsAbstract):
+    def save(self, *args: Any, **kwargs: Any) -> None:
+        self._fix_user_has_two_ongoing_salaries()
+        super(UserSalaryDetails, self).save(*args, **kwargs)
```
kbeker, line 77: You can simplify this -> `super().save(*args, **kwargs)`

billing/models.py
```diff
@@ -1,3 +1,91 @@
...
+
+    def _fix_user_has_two_ongoing_salaries(self) -> None:
+        if self.to_date is None:
+            if UserSalaryDetails.objects.filter(user_id=self.user.pk).filter(to_date__isnull=True).exists():
```
kbeker, line 81: You're filtering by `id` but you're using `pk`, please fix it :)

billing/models.py
```diff
@@ -1,3 +1,91 @@
...
+                salary_details = UserSalaryDetails.objects.filter(user_id=self.user.pk).get(to_date__isnull=True)
```
kbeker, line 82: I see that you're using it multiple times. You can create from this Manager and use it without writing again and again this queries.

Please create `UserSalaryDetailsManager`. I'll show you example of that:

```python
class UserSalaryDetailsManager(Manager):
    def get_ongoing_salary_or_none(self):
        try:
            return self.get(user_id=self.user.id, to_date__isnull=True)
        except self.model.DoesNotExist:
            return None
```
and then you need to attach this manager to model:

```python
class UserSalaryDetails(BaseModel):
    objects = UserSalaryDetailsManager()
```

And by this you will be able to use it everywhere where you want.

In this case you will be able to use it that way:
```python
salary_details = UserSalaryDetails.objects.get_ongoing_salary_or_none()
```

billing/models.py
```diff
@@ -1,3 +1,91 @@
...
+    def get_currency_symbol(self) -> str:
+        return CurrencyConstants.CurrencyDict[self.currency]["Symbol"]
+
+
+class UserSalaryDetails(BillingDetailsAbstract):
+    def save(self, *args: Any, **kwargs: Any) -> None:
+        self._fix_user_has_two_ongoing_salaries()
+        super(UserSalaryDetails, self).save(*args, **kwargs)
+
+    def _fix_user_has_two_ongoing_salaries(self) -> None:
```
kbeker, line 79: It's better to keep such things in clean then in save :)

Please also change name of it because `fix` is not a perfect word for such things. It'd be better to have something like `check_ongoing_salaries` or something like that :)

kbeker (changes requested):
There are few things to be polished but in overall good job!

billing/models.py
Seveex: That's true. Removed

billing/models.py
Seveex: Done

billing/models.py
Seveex: Done

billing/models.py
Seveex: Ok

billing/models.py
Seveex: You already did it with custom manager :)

billing/models.py
Seveex: Done

kbeker (approved):

billing/models.py
```diff
@@ -1,3 +1,103 @@
...
+        super().clean()
+
+        if self.to_date is not None and self.to_date <= self.from_date:
+            raise ValidationError(
+                message=BillingValidationErrorText.VALIDATION_ERROR_FROM_DATE_LATER_THAN_TO_DATE.value
+            )
+        if self.to_date is None and not self.user.is_active:
+            raise ValidationError(
+                message=BillingValidationErrorText.VALIDATION_ERROR_TO_DATE_EMPTY_FOR_INACTIVE_USER.value
+            )
```
Merrique, line 61: Why check for `self.to_date`? What if user inactive but `self.to_date` is given?

billing/models.py
```diff
@@ -1,3 +1,103 @@
...
+
+class UserSalaryDetails(BillingDetailsAbstract):
+    objects = UserSalaryDetailsManager()
+
+    def clean(self) -> None:
+        self._check_ongoing_salaries()
+        super().clean()
+
+    def _check_ongoing_salaries(self) -> None:
+        if self.to_date is None:
```
Merrique, line 92: As above. What if `self.to_date` is defined? Than we don't want to check ongoing salaries? Why?!?

billing/models.py
```diff
@@ -1,3 +1,103 @@
...
+
+    @property
+    def get_currency_symbol(self) -> str:
+        return CurrencyConstants.CurrencyDict[self.currency]["Symbol"]
+
+
+class UserSalaryDetailsManager(models.Manager):
+    def get_ongoing_salary_or_none(self, user: CustomUser) -> Union[models.QuerySet, None]:
+        try:
+            return self.get(user_id=user.id, to_date__isnull=True)
```
Merrique, line 79: Why do we check again if `to_date` is `Null`?   I believe we should have access to salary as long as the date we are checking for is before `to_date`

billing/models.py
```diff
@@ -1,3 +1,103 @@
...
+
+    def _check_ongoing_salaries(self) -> None:
+        if self.to_date is None:
+            salary_details = UserSalaryDetails.objects.get_ongoing_salary_or_none(self.user)
+            if salary_details is not None:
+                current_time = timezone.datetime.now(tz=timezone.get_default_timezone())
+                salary_details.to_date = timezone.make_aware(
+                    datetime.datetime(year=current_time.year, month=current_time.month, day=current_time.day)
+                )
+                salary_details.save()
```
Merrique, line 99: What is the purpose of `_check_ongoing_salaries`. Why is it in clean? Why we check again only if `self.to_date is None` ?Why do we apply with this function `to_date` value?

billing/models.py
```diff
@@ -1,3 +1,103 @@
...
+
 
-# Create your models here.
+class ProjectRevenueDetails(BillingDetailsAbstract):
+    project = models.ForeignKey(Project, on_delete=models.PROTECT, related_name="%(class)s")
```
Merrique, line 103: Here I am missing validation for cases when the project is finished, the project is suspended, the user is not in the project. The start date of the project is after the `from_date`. The finish date of the project is before the 'to_date'.

billing/tests/test_unit_model.py
```diff
@@ -0,0 +1,75 @@
...
+    ):
+        with self.assertRaises(ValidationError) as custom_exception:
+            ProjectRevenueDetails.objects.create(**self._billing_details_parameters(has_to_date=True, is_revenue=True))
+
+        self.assertEqual(
+            BillingValidationErrorText.VALIDATION_ERROR_FROM_DATE_LATER_THAN_TO_DATE.value,
+            custom_exception.exception.messages[0],
+        )
+
+    def test_creating_user_salary_details_of_inactive_user_with_blank_to_date_raise_custom_validation_error(self):
```
Merrique, line 46: Why again blank to_date?

billing/tests/test_unit_model.py
```diff
@@ -0,0 +1,75 @@
...
+            custom_exception.exception.messages[0],
+        )
+
+    def test_creating_user_salary_details_with_blank_to_date_if_user_already_has_one_should_automatically_set_to_date_of_first_salary_details(
+        self
+    ):
+        salary_details = UserSalaryDetails.objects.create(**self._billing_details_parameters())
+        UserSalaryDetails.objects.create(**self._billing_details_parameters())
+
+        self.assertNotEqual(UserSalaryDetails.objects.get(pk=salary_details.pk).to_date, None)
```
Merrique, line 61: Why we need this to work like this?!?

Merrique (changes requested):

Merrique:
PR moved to and worked further on it in GitLab

PR #621, opened 2021-08-09
Bugfix report model validation

Merrique:


employees/common/strings.py
```diff
@@ -111,12 +109,13 @@ class ReportValidationStrings(NotCallableMixin, Enum):
     WORK_HOURS_SUM_FOR_GIVEN_DATE_FOR_SINGLE_AUTHOR_EXCEEDED = _(
         "Sum of work hours of all reports from a given day for single author must not exceed 24."
     )
-    WORK_HOURS_MIN_VALUE_NOT_EXCEEDED = _(
-        f"Minimum value for work hours for single report must not be less than {int(ReportModelConstants.MIN_WORK_HOURS.value.total_seconds() / 60)} minutes"
-    )
+    WORK_HOURS_MIN_VALUE_NOT_EXCEEDED = _("Ensure this value is greater than or equal to 0:15:00.")
```
kbeker, line 121: As far as I remember we should cut off last numbers (seconds in this case) and just show HH:MM

sheetstorm/management/commands/generate_test_data.py
```diff
@@ -403,16 +403,21 @@ def create_report(self, user: Any, report_publication_date: timezone.datetime, n
...
-        return random.choice(user_projects_list)
+    def _pick_random_active_user_project(user: Any) -> Any:
+        if user.projects.filter(suspended=False).count() != 0:
+            user_projects_list = list(user.projects.filter(suspended=False))
+            return random.choice(user_projects_list)
+        else:
+            active_projects = list(user.projects.filter(suspended=True))
+            random_suspended_project = random.choice(active_projects)
+            random_suspended_project.suspended = False
+            return random_suspended_project
```
kbeker, line 420: You don't need else clause here if first if statement returns anything

kbeker (approved):

